PR #5, opened 2015-02-11
Explicit message when wait for user input

softmonkeyjapan:
`Press RETURN to continue...` is not explicit on what needs to be done. We guess it's about the enter key, but it's more natural to read `Press ENTER to continue...` IMHO

### DIFF
--- a/install
+++ b/install
@@ -54,7 +54,7 @@ end
 
 def wait_for_user
   puts
-  puts "Press RETURN to continue or any other key to abort"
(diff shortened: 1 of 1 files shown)


tdsmith:
Thanks for the suggestion! "Return" is intentional and quite explicit: previously discussed at Homebrew/homebrew#26042.

softmonkeyjapan:
Thanks for this clarification

PR #6, opened 2015-02-11
Don't tell new users to run brew doctor

tdsmith:
Running `brew doctor` before a user runs into trouble seems to inspire anxiety uncorrelated to the magnitude of the problems encountered, which has gotten worse as `brew doctor` has gotten more verbose (and more helpful!). Telling users to run `doctor` out of the gate conflicts with the spirit of the message in the `doctor` output telling users not to worry about reports until they've actually run into trouble. I think removing this advice will help improve the new user experience.

### DIFF
--- a/install
+++ b/install
@@ -219,5 +219,4 @@ else
   puts "Install #{Tty.white}Xcode#{Tty.reset}: https://developer.apple.com/xcode" unless File.exist? "/usr/bin/cc"
 end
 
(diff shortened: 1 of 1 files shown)


mikemcquaid:
This makes sense to me :+1:.

FWIW I think we can use the "Merge pull request" button in this repo. I've been using it in the homepage repo too.

PR #8, opened 2015-04-02
Default to shallow clone

xu-cheng:
Since most of users wouldn't care about the history, let's default
to shallow clone unless `--full` is passed.

Related issues:

* https://github.com/Homebrew/homebrew/issues/37592
* https://github.com/Homebrew/homebrew/pull/38268

cc @mikemcquaid

### DIFF
--- a/install
+++ b/install
@@ -192,7 +192,7 @@ Dir.chdir HOMEBREW_PREFIX do
     system git, "config", "remote.origin.fetch", "+refs/heads/*:refs/remotes/origin/*"
 
     args = git, "fetch", "origin", "master:refs/remotes/origin/master", "-n"
-    args << "--depth=1" if ARGV.include? "--fast"
+    args << "--depth=1" unless ARGV.include?("--full") || !ENV["HOMEBREW_DEVELOPER"].nil?
     system(*args)
 
     system git, "reset", "--hard", "origin/master"


mikemcquaid:
Now this is one I probably would set to true if `HOMEBREW_DEVELOPER` is set. Otherwise will want other @Homebrew/owners thoughts but could be a good call.

mikemcquaid:
Can you benchmark the speed difference (for your machine)?

tdsmith:
Oh, I accidentally did that test. On Travis it's about 21s for a deep clone: https://travis-ci.org/tdsmith/homebrew-pypi-poet/jobs/55593708
and 3s for a shallow clone: https://travis-ci.org/tdsmith/homebrew-pypi-poet/jobs/55726449

What do we tell people who got a shallow clone and want to recover formulas from history?

mikemcquaid:
> What do we tell people who got a shallow clone and want to recover formulas from history?

We tell them they shouldn't be doing that anyway :wink:

xu-cheng:
Or we can ask them to run `git fetch --unshallow`.

xu-cheng:
Updated to check `HOMEBREW_DEVELOPER` as well. Though, it's kind of wried to me. As How many developers would run this install script multiple times rather than using git clone?

mikemcquaid:
Looks good to me :+1:

Chronix123:
v<sdc

PR #9, opened 2015-04-10
install: Fix stray non-breaking space

chdiza:
Somehow a regular space got turned into a non-breaking space.


mikemcquaid:
Thanks @chdiza!

PR #13, opened 2015-04-20
uninstall: don't remove `bin` folder

xu-cheng:
Ref: https://github.com/Homebrew/homebrew/pull/38704#issuecomment-94282009


mikemcquaid:
:+1:

PR #14, opened 2015-05-30
Allow parameterization of standard locations like HOMEBREW_PREFIX

nz:
This pull request allows constants like `HOMEBREW_PREFIX` to be overridden via environment variables. This makes it easier to install Homebrew to a non-standard location — for example, a user-specific installation in `$HOME/.brew` — while reusing the standard install script.

```bash
HOMEBREW_PREFIX=$HOME/.brew \
ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
```

### DIFF
--- a/install
+++ b/install
@@ -1,10 +1,10 @@
 #!/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby
 # This script installs to /usr/local only. To install elsewhere you can just
 # untar https://github.com/Homebrew/homebrew/tarball/master anywhere you like or
-# change the value of HOMEBREW_PREFIX.
-HOMEBREW_PREFIX = '/usr/local'
-HOMEBREW_CACHE = '/Library/Caches/Homebrew'
-HOMEBREW_REPO = 'https://github.com/Homebrew/homebrew'
+# provide a HOMEBREW_PREFIX environment variable.
+HOMEBREW_PREFIX = ENV.fetch('HOMEBREW_PREFIX') { '/usr/local' }
+HOMEBREW_CACHE =  ENV.fetch('HOMEBREW_CACHE')  { '/Library/Caches/Homebrew' }
+HOMEBREW_REPO =   ENV.fetch('HOMEBREW_REPO')   { 'https://github.com/Homebrew/homebrew' }
 
 module Tty extend self
   def blue; bold 34; end


mikemcquaid:
What's the advantage of this approach over e.g. encouraging people to just clone the repository there? I'm personally :-1: on making it easier or "official" to support outside of `/usr/local` as you don't get most binary packages which makes it significantly more error-prone (and this adds to our support burden).

nz:
I totally defer to your opinions here, given that you have a lot more context than I do. Let me lay out my use case in a bit more detail.

My motivation here is to assist with a scripted installation (e.g., via something like thoughtbot/laptop) for shared multi-user systems. Multi-user shared installs can be tricky. My first preference was to use group permissions on /usr/local, but Postgres in particular is pretty strict about its ownership and permissions. That's leading me toward more isolated installations in each user's home directory.

My next step was going to be the suggested/documented approach of cloning the repository itself into a desired location. Then I noticed that the install script does handle a bit more than just that (even if it's mostly validation), and it has constants at the top of the file that seemed pretty ripe for parameterization. From there I figured env-var parameterization was a pretty trivial change that would support my use case, and help provide a bit of future-proofing should the install script ever change over time.

It may be that the non-standard insall location is problematic enough to make this a total non-starter. But that seems a bit orthogonal from where I'm sitting.

Again, totally defer to your opinion on this, just found myself tweaking code and thought I'd send along a PR with my thoughts :)

mikemcquaid:
Unfortunately our install script is not completely able to run non-interactively (e.g. if you don't have the Xcode CLT installed).

It's perfectly possible to run multi-user installations of Homebrew in `/usr/local` to I'd strongly advise you towards that as otherwise (as mentioned) the users will get a significantly worse experience (more bugs, much longer install times). Homebrew is fine doing pretty much anything except `install/upgrade/reinstall/uninstall` with read-only permissions so you should be able to fix them such that it can write to those places, even if that requires using `sudo -u brewinstall brew install` or something that you give all users permission to.

nz:
Sounds like I'm going back to /usr/local — if you don't have any other need to fetch those paths from the environment, feel free to close this.

mikemcquaid:
Great, thanks. Good luck and shout if you want any more help integrating Homebrew in that way; it's something I'm interested in myself too.

PR #15, opened 2015-06-16
modifications to prompt for CLT installation before allowing the popup

woodruffw:
In keeping with Homebrew/homebrew/pull/40721, `install.rb` shouldn't bug the user with requests to install the Command Line Tools once they are no longer required for bottles. This probably shouldn't be merged in until the above is, as the CLT are still required.

### DIFF
--- a/install
+++ b/install
@@ -60,6 +60,15 @@ def wait_for_user
   abort unless c == 13 or c == 10
 end
 
+def prompt str
+  loop do
+    print "#{str} (y/n): "
+    ans = gets.strip
+    return true if ans =~ /y/i
+    return false if ans =~ /n/i
+  end
+end
+
 class Version
   include Comparable
   attr_reader :parts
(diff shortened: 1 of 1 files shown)


mikemcquaid:
I'm not sure about this given they are also needed for e.g. `git`.

woodruffw:
Yeah, there's going to be a problem where locating `git` on systems with the dummy binaries in `/usr/bin` will cause that popup.
My popup solution means that a real `git` in `/usr/bin` is ignored if the CLT aren't already installed, but the installation still proceeds via `curl`. Once the user actually needs to update with `brew update`, I could add an information message telling them to `brew install git` first (which currently works with the CLT-less bottling system).

mikemcquaid:
I think it's fine to just get people to use the CLT for the initial installer. This is something we can revisit in e.g. 6 months.

woodruffw:
Alright!

PR #17, opened 2015-06-26
uninstall: rescue exceptions when deleting file

### LINKED ISSUE #16: uninstaller doesn't check if the directories were deleted or not.
When I uninstall homebrew, the uninstaller prompts it will delete /Library/Caches/Homebrew/ but it fails because the permission of /Library/Caches/Homebrew/Formula was of another user that had already installed homebrew. Unless manually deleting /Library/Caches/Homebrew/ directory, installing rvm will fail while installing libraries. It will be better if homebrew uninstaller checks if all the directories supposed to be deleted actually were deleted or not. 
Mac OS X 10.10.3
uninstaller: ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall)"

xu-cheng:
Fixes #16.

### DIFF
--- a/uninstall
+++ b/uninstall
@@ -171,7 +171,12 @@ $HOMEBREW_FILES.each do |file|
   if options[:dry_run]
     puts "Would delete #{file}"
   else
-    FileUtils.rm_rf(file)
+    begin
+      FileUtils.rm_rf(file)
+    rescue => e
+      warn "Failed to delete #{file}"
+      puts e.message
+    end
   end
 end
 


mikemcquaid:
:+1: although it'd be good to elevate permissions to actually delete them.

xu-cheng:
>  it'd be good to elevate permissions to actually delete them.

I am not sure we should do this in the situation of #16, which is caused by share folder with other user. So I am gonna merge this as it is.

mikemcquaid:
@xu-cheng That's the actual problem, though; people are trying to uninstall Homebrew and failing because this script doesn't elevate permissions.

xu-cheng:
In that case, user should run entire script with sudo. Since in normal single user situation, root privilege is not required. Maybe we should alter the document to suggest user to re-run script with sudo in the even of failure.

mikemcquaid:
Yep and it'd be good if that was suggested by error messages too.

xu-cheng:
#18

PR #20, opened 2015-07-07
Use #!/usr/bin/env ruby for Linuxbrew

sjackman:
Would you consider the following patch for compatibility with Linuxbrew? Feel free to close this pull request and decline if not.
See https://github.com/Homebrew/linuxbrew/issues/49#issuecomment-119038295


mikemcquaid:
I think this increases the chances of errors on OS X so I'd rather not.

PR #22, opened 2015-09-10
Allow environment overrides for main config variables

arcivanov:


mikemcquaid:
What's the use-case here? I'd rather not make it easier to use custom locations, personally.

arcivanov:
Use case is exactly to allow to install into custom locations - I'm trying to get home/linuxbrew into a Python virtualenv to provide an isolated environment. I fully understand the dangers and will deal with them. 

There is no harm here for people who don't want overrides while it provides the perilous flexibility to others.

mikemcquaid:
You can install into custom locations without using the installer.

arcivanov:
I am aware, but with installer it stays a one-liner and I don't have to actually take install script and copy out the chunks of it and then maintain them or maintain a fork of install scripts.

This change preserves full backwards compatibility and does not affect the users who do not want to use custom location in any way. What's wrong with making it convenient to experiment? These options can stay undocumented and unsupported.

DomT4:
I'm uncomfortable with the potential security issue of letting a local `HOMEBREW_REPO` env manipulate where Homebrew is installed from.

arcivanov:
vs the path where it installs to? If someone can set your environment they at least have as much privileges as you do, i.e. maliciousness can spread to everything user does, no?

This feature is useful because I may choose to use a file url to clone a disk repo or an internally hosted clone that is a cached snapshot of Homebrew GH repo.

This is all about automation here.

DomT4:
By default the script will shout about where you're installing it *before* the process begins; It doesn't output where the `git clone` is being performed from until the end of that process.

```
==> This script will install:
/opt/testing/bin/brew
/opt/testing/Library/...
/opt/testing/share/man/man1/brew.1

==> /usr/bin/sudo /bin/mkdir /opt/testing
Password:
==> /usr/bin/sudo /bin/chmod g+rwx /opt/testing
==> /usr/bin/sudo /usr/bin/chgrp admin /opt/testing
==> /usr/bin/sudo /bin/mkdir /Library/Caches/Homebrew
==> /usr/bin/sudo /bin/chmod g+rwx /Library/Caches/Homebrew
==> Downloading and installing Homebrew...
remote: Counting objects: 284545, done.
remote: Compressing objects: 100% (3394/3394), done.
remote: Total 284545 (delta 2349), reused 0 (delta 0), pack-reused 281133
Receiving objects: 100% (284545/284545), 50.83 MiB | 833.00 KiB/s, done.
Resolving deltas: 100% (213538/213538), done.
From https://github.com/Homebrew/homebrew
```

There's no "Is this the genuine Homebrew repo you expected" step there. We could add one, but I'm not hugely sold on the env here.

arcivanov:
But even though this is an HTTPS connection there is no certificate validation anyway. I.e. you can't know you're contacting a genuine GitHub, and for that M-i-M attack ENV doesn't even need to be used.

arcivanov:
I can add printing of the repo if this would alleviate your concern

mikemcquaid:
Unless another maintainer strongly feels this should go in I'm still :-1:.

DomT4:
> I can add printing of the repo if this would alleviate your concern

A little. In an ideal world I'd like to stop using a script for installation entirely, really. 

I'm not sure on the use case of being able to install Homebrew offline though. You'd have your hands tied on updating, installing & various commands until you regained connection.

For better or worse, even "undocumented and unsupported" options end up getting used and people file issues with us about them. I think if we introduced this people would expect us to support it indefinitely, as has been the case with other things.

arcivanov:
Ok, thanks for your time. Withdrawn.

DomT4:
Thanks for the PR, even if we didn't accept it at this time, appreciate the submission.

mikemcquaid:
Yup, thanks from me too and sorry to not accept it.

Issue #21, opened 2015-08-08
When installing - Error: Permission denied - /Library/Caches/Homebrew/Formula


DomT4:
_From @fbucek on August 7, 2015 12:35_

When I follow web page brew.sh to install HomeBrew
1) Install HomeBrew: 
ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
OK - working

2) Install wget
brew install wget
Error: Permission denied - /Library/Caches/Homebrew/Formula

I have to "fix" this using command found here ( found here in issues somewhere )
sudo chown -R $USER:admin $(brew --cache)

Another mention of this problem:
http://stackoverflow.com/questions/17610417/permission-denied-library-caches-homebrew-formula-nmap-brewing
Answer helped some people.

OS: OSX 10.10.5

_Copied from original issue: Homebrew/homebrew#42606_

DomT4:
_From @dunn on August 7, 2015 15:44_

Thanks for the report.  This is probably something we can check during installation.

dunn:
I wasn't sure whether to move it because this is something that _could_ be addressed in `formulary.rb` (https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/formulary.rb#L130).  But my guess it that we should figure out a way to preempt the issue here, since it might need `sudo`.

DomT4:
Yeah, my reading was that since the user didn't do anything to the cache before `brew install wget` the error, if it is an error, was the installer's. I could be wrong on that reading though.

We already create the [cache](https://github.com/Homebrew/homebrew/blob/master/Library/Homebrew/config.rb#L1-L24) to be world-writeable unless it already exists, although only the top level directory.

I'm not quite sure how it could end up in this situation, unless `HOMEBREW_CACHE+"Formula"` already exists for some reason and isn't getting generated after the cache is created with world permissions.

Generally I'm not a fan of cross-user world-writeable cache's being supported for this and other reasons, and would prefer for our default to be a per-user cache, but I haven't won that argument to date.

mikemcquaid:
This could be fixed in https://github.com/Homebrew/install/pull/23 I guess?

PR #23, opened 2015-09-23
install: chown as user.

mikemcquaid:
We've been trying to avoid doing this for a while but it seems sensible to just do this by default considering how often we're recommending users do it.

Closes https://github.com/Homebrew/install/issues/21.

CC @Homebrew/owners

### DIFF
--- a/install
+++ b/install
@@ -100,6 +100,10 @@ def chmod?(d)
   File.directory?(d) && !(File.readable?(d) && File.writable?(d) && File.executable?(d))
 end
 
+def chown?(d)
+  !File.owned?(d)
+end
+
 def chgrp?(d)
   !File.grpowned?(d)
 end
@@ -157,11 +161,16 @@ chmods = %w( . bin etc include lib lib/pkgconfig Library sbin share var var/log 
              share/man/man5 share/man/man6 share/man/man7 share/man/man8
              share/info share/doc share/aclocal ).
              map { |d| File.join(HOMEBREW_PREFIX, d) }.select { |d| chmod?(d) }
+chowns = chmods.select { |d| chown?(d) }
 chgrps = chmods.select { |d| chgrp?(d) }
 
 unless chmods.empty?
   ohai "The following directories will be made group writable:"
   puts(*chmods)
+end
+unless chowns.empty?
+  ohai "The following directories will have their owner set to #{Tty.underline 39}#{ENV['USER']}#{Tty.reset}:"
+  puts(*chowns)
 end
 unless chgrps.empty?
   ohai "The following directories will have their group set to #{Tty.underline 39}admin#{Tty.reset}:"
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -157,12 +161,17 @@ chmods = %w( . bin etc include lib lib/pkgconfig Library sbin share var var/log
...
              map { |d| File.join(HOMEBREW_PREFIX, d) }.select { |d| chmod?(d) }
+chowns = chmods.select { |d| chown?(d) }
 chgrps = chmods.select { |d| chgrp?(d) }
 
 unless chmods.empty?
   ohai "The following directories will be made group writable:"
   puts(*chmods)
 end
+unless chowns.empty?
+  ohai "The following directories will have their owner set to #{Tty.underline 39}#{ENV['USER']}#{Tty.reset}:"
```
bfontaine: We should put this `Tty.underline 39` in a variable with a meaningful name; it’s hard to guess what it means here.

DomT4:
Agree on the readability issue re Baptiste's point, but :+1:. Has more or less become our go-to fix for permission errors reported.

install
```diff
@@ -172,16 +181,19 @@ wait_for_user if STDIN.tty?
...
   sudo "/bin/mkdir", HOMEBREW_PREFIX
   sudo "/bin/chmod", "g+rwx", HOMEBREW_PREFIX
   # the group is set to wheel by default for some reason
-  sudo "/usr/bin/chgrp", "admin", HOMEBREW_PREFIX
+  sudo "/usr/sbin/chown", "#{ENV['USER']}:admin", HOMEBREW_PREFIX
 end
 
 sudo "/bin/mkdir", HOMEBREW_CACHE unless File.directory? HOMEBREW_CACHE
 sudo "/bin/chmod", "g+rwx", HOMEBREW_CACHE if chmod? HOMEBREW_CACHE
+sudo "/usr/sbin/chown", ENV['USER'], HOMEBREW_PREFIX if chown? HOMEBREW_CACHE
```
dunn: Is this supposed to be `sudo "/usr/sbin/chown", ENV['USER'], HOMEBREW_CACHE if chown? HOMEBREW_CACHE` (and the next line)?

mikemcquaid: Yes, well spotted.

install
```diff
@@ -172,16 +181,19 @@ wait_for_user if STDIN.tty?
...
   sudo "/bin/chmod", "g+rwx", HOMEBREW_PREFIX
   # the group is set to wheel by default for some reason
-  sudo "/usr/bin/chgrp", "admin", HOMEBREW_PREFIX
+  sudo "/usr/sbin/chown", "#{ENV['USER']}:admin", HOMEBREW_PREFIX
 end
 
 sudo "/bin/mkdir", HOMEBREW_CACHE unless File.directory? HOMEBREW_CACHE
 sudo "/bin/chmod", "g+rwx", HOMEBREW_CACHE if chmod? HOMEBREW_CACHE
+sudo "/usr/sbin/chown", ENV['USER'], HOMEBREW_CACHE if chown? HOMEBREW_CACHE
+sudo "/usr/bin/chgrp", "admin", HOMEBREW_CACHE if chgrp? HOMEBREW_CACHE
```
xu-cheng: I'm confused why we need to chown/chgrp for cache directory? We seem never recommend users to do this(though we recommend this for PREFIX a lot). If we really need this, why don't we just start using `~/Library/Cache/Homebrew` as the default cache directory?

Also, we did't mention to chown/chgrp this directory in the warning message above, it's bad to do this without users' awareness.

mikemcquaid: The reason we don't announce that is because it's a Homebrew-specific directory whereas the others may contain other software. I don't think the installer needs to say everything that it does.

mikemcquaid: We wanted to have a cache directory that's shareable between users (hence the various group-writable permissions). However, this is less than optimal and our oldest open PR is trying to deal with it: https://github.com/Homebrew/homebrew/pull/23233. The thing blocking that was the need to migrate from the existing cache directories.

If you feel like picking that up it'd be :+1:

mikemcquaid: These particular changes are due to users reporting issues on or after installation.

mikemcquaid:
Unless there's any massive last-minute objections when this is :green_heart: I'm going to merge. It fixes user bugs and although there's definitely tweaks that should be made I think they need to block this PR. We have CI on this repository now so hopefully that encourages more work here.

install
DomT4: > The thing blocking that was the need to migrate from the existing cache directories.

Personally want to move it to `HOMEBREW_PREFIX/var/cache/homebrew`. I raised the question a while ago now but there didn't seem much interest in changing it, so I left it be :smile:.

mikemcquaid: That works for me.

DomT4: Did some work on this tonight. Works nicely on local testing, all tests passing, no functionality breakage, etc. Will polish it and add documentation later and push for review. I kept it fairly minimal, which may be controversial, but we'll see.

Have a bit more time over the next week than I have lately, so hoping to catch up on a lot of the changes I've left sitting for a while.

mikemcquaid: Nice. The main thing is just making sure it migrates the existing files :+1:

PR #25, opened 2015-10-22
Allowing HOMEBREW_PREFIX to be set by an environment variable.

cbrichford:
With this change Homebrew can be installed into a directory other
than /usr/local by running the installer with this command:
```
env HOMEBREW_PREFIX=/some/other/directory ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
```

Given the changes in OSX El Capitan it now seems desirable to install Homebrew in a location that is not under /usr.  This change would allow El Capitan users to specify a location other than /usr/local for the HOMEBREW_PREFIX without downloading the installer and modifying it.

### DIFF
--- a/install
+++ b/install
@@ -2,7 +2,7 @@
 # This script installs to /usr/local only. To install elsewhere you can just
 # untar https://github.com/Homebrew/homebrew/tarball/master anywhere you like or
 # change the value of HOMEBREW_PREFIX.
-HOMEBREW_PREFIX = '/usr/local'
+HOMEBREW_PREFIX = ENV.fetch('HOMEBREW_PREFIX', '/usr/local')
 HOMEBREW_CACHE = '/Library/Caches/Homebrew'
 HOMEBREW_REPO = 'https://github.com/Homebrew/homebrew'
 


bfontaine:
We don’t want to encourage users to use a different location; see #22.

cbrichford:
@bfontaine #22 allows changing of all setup variables.  Someone rightly pointed out that setting the location of the HOMEBREW_REPO via an environment variable could be a security issue as it would allow an attacker that controls and environment variable to specify where brew packages come from.

Setting the HOMEBREW_PREFIX seems less dangerous.

What I'm really trying to fix is not having to do these steps https://git.io/vnC7w on El Capitan.  An another solution to the problem I'm trying to solve would be to change the default HOMEBREW_PREFIX on El Capitan to be something other than /usr/local.

bfontaine:
@cbrichford You can [install Homebrew elsewhere](https://github.com/Homebrew/homebrew/blob/master/share/doc/homebrew/Installation.md#untar-anywhere) by cloning it or downloading a tarball and untarring it where you want.

You could do e.g.:
```sh
mkdir ~/.brew && cd ~/.brew && git clone --depth=1 https://github.com/Homebrew/homebrew.git
```

Then add `export PATH="$HOME/.brew/bin:$HOME/.brew/sbin:$PATH"` in your `~/.bashrc` and you’re done.

mikemcquaid:
> What I'm really trying to fix is not having to do these steps https://git.io/vnC7w on El Capitan. The reboot into the "Recovery Mode" is particularly obnoxious in the current workflow. Another solution to the problem I'm trying to solve would be to change the default HOMEBREW_PREFIX on El Capitan to be something other than /usr/local.

You don't have to do that unless you've deleted `/usr/local`. We're not changing the default prefix, sorry, or allowing it to be customised here.

PR #27, opened 2015-11-01
install: update 10.11 /usr/local instructions.

mikemcquaid:
These seem to not apply for everyone on 10.11.0 any more (as explained in https://github.com/Homebrew/homebrew/issues/45387).

CC @DomT4 @ilovezfs @CamJN

### DIFF
--- a/install
+++ b/install
@@ -145,8 +145,10 @@ Before running the installer again please agree to the license by opening
 Xcode.app or running:
     sudo xcodebuild -license
 EOABORT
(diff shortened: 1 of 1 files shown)


mistydemeo:
Does that file exist in all supported versions of OS X, or do we need to guard this with a `File.exist?` check?

mikemcquaid:
@mistydemeo Good point, probably needs a test.

PR #29, opened 2015-11-09
Share some code with a Homebrew command.

mikemcquaid:
Also fixes an issue where things would only be `chown`d or `chgrp`d if they had incorrect `chmod` too.

Also see https://github.com/Homebrew/homebrew/pull/45842

CC @DomT4 @bfontaine @xu-cheng

### DIFF
--- a/install
+++ b/install
@@ -36,30 +36,6 @@ def system *args
   abort "Failed during: #{args.shell_s}" unless Kernel.system(*args)
 end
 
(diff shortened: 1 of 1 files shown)


mikemcquaid:
Closing in favour of @xu-cheng's version for this repo of https://github.com/Homebrew/homebrew/issues/45934.

PR #31, opened 2015-11-15
Save sandbox abort test from greedy heredoc

ilovezfs:
Heredocs begin immediately on the next line, which means that the
expression containing <<-EOABORT and the modifier-if must not spill onto
the next line even if '\' is used.


ilovezfs:
I was also wondering if each of the abort tests should be refactored into helper functions to improve readability and to eliminate the "artificial pressure" to keep them as short as possible.

ilovezfs:
@mikemcquaid You're welcome :)

PR #33, opened 2016-01-24
Enable headless/non-interactive install of Command Line Tools.

marcosscriven:
It bugged me a little there's just this one GUI interaction required to install brew on a newly installed machine. By creating the temporary file ```/tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress```, you can instead use ```softwareupdate``` to download without any interaction required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,7 @@ matrix:
 # This ensures the Command Line Tool install functionality is tested.
 before_script:
   - sudo rm -rf /Applications/Xcode.app /Library/Developer/CommandLineTools
+  - sudo pkgutil --forget com.apple.pkg.CLTools_Executables
 
 script:
   - ./uninstall -d
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,6 +15,11 @@ matrix:
       osx_image: beta-xcode6.2
       rvm: system
 
+# Remove Xcode and CommandLineTools included in default OS X image.
+# This ensures the Command Line Tool install functionality is tested.
+before_script:
+  - sudo rm -rf /Applications/Xcode.app /Library/Developer/CommandLineTools
+
 script:
   - ./uninstall -d
   - ./uninstall -f
--- a/install
+++ b/install
@@ -201,10 +201,12 @@ sudo "/usr/bin/chgrp", "admin", HOMEBREW_CACHE if chgrp? HOMEBREW_CACHE
 if macos_version >= "10.9"
   developer_dir = `/usr/bin/xcode-select -print-path 2>/dev/null`.chomp
   if developer_dir.empty? || !File.exist?("#{developer_dir}/usr/bin/git")
-    ohai "Installing the Command Line Tools (expect a GUI popup):"
-    sudo "/usr/bin/xcode-select", "--install"
-    puts "Press any key when the installation has completed."
-    getc
+    ohai "Searching online for the Command Line Tools"
+    # This temporary file prompts software update to list the Command Line Tools
+    `touch /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress`
+    clt = `softwareupdate -l | grep "* Command Line Tools" | sed 's/^.*\\(Command Line Tools.*\\)/\\1/'`.chomp
+    ohai "Installing #{clt}"
+    `softwareupdate -i "#{clt}"`
   end
 end
 
--- a/install
+++ b/install
@@ -215,16 +215,19 @@ if should_install_command_line_tools
   clt_label = `softwareupdate -l | grep -B 1 -E "Command Line (Developer|Tools)" | awk -F"*" '/^ +\\*/ {print $2}' | sed 's/^ *//' | head -n1`.chomp
   ohai "Installing #{clt_label}"
   sudo "/usr/sbin/softwareupdate", "-i", clt_label
-  sudo "/bin/rm", clt_placeholder
+  sudo "/bin/rm", "-f", clt_placeholder
 end
 
 # Headless install may have failed, so fallback to original 'xcode-select' method
 if should_install_command_line_tools
-  ohai "Headless installation of Command Line Tools has failed"
-  ohai "Installing the Command Line Tools (expect a GUI popup):"
-  sudo "/usr/bin/xcode-select", "--install"
-  puts "Press any key when the installation has completed."
-  getc
+  if STDIN.tty?
+    ohai "Installing the Command Line Tools (expect a GUI popup):"
+    sudo "/usr/bin/xcode-select", "--install"
+    puts "Press any key when the installation has completed."
+    getc
+  else
+    abort "Unable to proceed with manual Command Line Tools install without user input."
+  end
 end
 
 ohai "Downloading and installing Homebrew..."
--- a/install
+++ b/install
@@ -77,13 +77,10 @@ def macos_version
   @macos_version ||= Version.new(`/usr/bin/sw_vers -productVersion`.chomp[/10\.\d+/])
 end
 
-def should_install_command_line_tools
-  if macos_version >= "10.9"
-    developer_dir = `/usr/bin/xcode-select -print-path 2>/dev/null`.chomp
-    return developer_dir.empty? || !File.exist?("#{developer_dir}/usr/bin/git")
-  else
-    return false
-  end
+def should_install_command_line_tools?
+  return false if macos_version < "10.9"
+  developer_dir = `/usr/bin/xcode-select -print-path 2>/dev/null`.chomp
+  developer_dir.empty? || !File.exist?("#{developer_dir}/usr/bin/git")
 end
 
 def git
@@ -207,7 +204,7 @@ sudo "/bin/chmod", "g+rwx", HOMEBREW_CACHE if chmod? HOMEBREW_CACHE
 sudo "/usr/sbin/chown", ENV['USER'], HOMEBREW_CACHE if chown? HOMEBREW_CACHE
 sudo "/usr/bin/chgrp", "admin", HOMEBREW_CACHE if chgrp? HOMEBREW_CACHE
 
-if should_install_command_line_tools
+if should_install_command_line_tools?
   ohai "Searching online for the Command Line Tools"
   # This temporary file prompts the 'softwareupdate' utility to list the Command Line Tools
   clt_placeholder = "/tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress"
@@ -219,14 +216,14 @@ if should_install_command_line_tools
 end
 
 # Headless install may have failed, so fallback to original 'xcode-select' method
-if should_install_command_line_tools
+if should_install_command_line_tools?
   if STDIN.tty?
     ohai "Installing the Command Line Tools (expect a GUI popup):"
     sudo "/usr/bin/xcode-select", "--install"
     puts "Press any key when the installation has completed."
     getc
   else
-    abort "Unable to proceed with manual Command Line Tools install without user input."
+    abort "Error: Cannot proceed with manual Command Line Tools install without user input!"
   end
 end
 
--- a/install
+++ b/install
@@ -77,6 +77,15 @@ def macos_version
   @macos_version ||= Version.new(`/usr/bin/sw_vers -productVersion`.chomp[/10\.\d+/])
 end
 
+def should_install_command_line_tools
+  if macos_version >= "10.9"
+    developer_dir = `/usr/bin/xcode-select -print-path 2>/dev/null`.chomp
+    return developer_dir.empty? || !File.exist?("#{developer_dir}/usr/bin/git")
+  else
+    return false
+  end
+end
+
 def git
   @git ||= if ENV['GIT'] and File.executable? ENV['GIT']
     ENV['GIT']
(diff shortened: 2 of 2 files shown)


marcosscriven:
I was pleased to note CI passed, but looking at the logs, it seems they are setup to test only installation on OS X with xcode (and thus the command line tools) already installed.

I've tested this on a fresh install of 10.11.2 (with the Vagrant image here https://atlas.hashicorp.com/jhcook/boxes/osx-elcapitan-10.11)

marcosscriven:
FYI - Output for OS X 10.11.2 with Vagrant:

```
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'jhcook/osx-elcapitan-10.11'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'jhcook/osx-elcapitan-10.11' is up to date...
==> default: A newer version of the box 'jhcook/osx-elcapitan-10.11' is available! You currently
==> default: have version '10.11.2'. The latest is version '10.11.3'. Run
==> default: `vagrant box update` to update.
==> default: Setting the name of the VM: test_default_1453647162747_78145
==> default: Clearing any previously set network interfaces...
==> default: Preparing network interfaces based on configuration...
    default: Adapter 1: nat
==> default: Forwarding ports...
    default: 22 (guest) => 2222 (host) (adapter 1)
==> default: Running 'pre-boot' VM customizations...
==> default: Booting VM...
==> default: Waiting for machine to boot. This may take a few minutes...
    default: SSH address: 127.0.0.1:2222
    default: SSH username: vagrant
    default: SSH auth method: private key
==> default: Machine booted and ready!
==> default: Checking for guest additions in VM...
    default: No guest additions were detected on the base box for this VM! Guest
    default: additions are required for forwarded ports, shared folders, host only
    default: networking, and more. If SSH fails on this machine, please install
    default: the guest additions and repackage the box to continue.
    default:
    default: This is not an error message; everything may continue to work properly,
    default: in which case you may ignore this message.
==> default: Running provisioner: shell...
    default: Running: inline script
==> default: The chosen disk supports resize; disregarding your new file system type and volume name
==> default: Merging partitions into a new partition
==> default:      Start partition: disk0s2 Macintosh HDD
==> default:      Finish partition: disk0s3 Blank
==> default: Started partitioning on disk0
==> default: Merging partitions
==> default: Waiting for the disks to reappear
==> default: Growing disk
==> default: Finished partitioning on disk0
==> default: /dev/disk0 (external, physical):
==> default:    #:                       TYPE NAME                    SIZE       IDENTIFIER
==> default:    0:      GUID_partition_scheme                        *137.4 GB   disk0
==> default:    1:                        EFI EFI                     209.7 MB   disk0s1
==> default:    2:                  Apple_HFS Macintosh HDD           137.1 GB   disk0s2
==> default: Running provisioner: shell...
    default: Running: inline script
==> default: ==> This script will install:
==> default: /usr/local/bin/brew
==> default: /usr/local/Library/...
==> default: /usr/local/share/man/man1/brew.1
==> default: ==> The following directories will be made group writable:
==> default: /usr/local/.
==> default: ==> The following directories will have their owner set to vagrant:
==> default: /usr/local/.
==> default: ==> The following directories will have their group set to admin:
==> default: /usr/local/.
==> default: ==> /usr/bin/sudo /bin/chmod g+rwx /usr/local/.
==> default: ==> /usr/bin/sudo /usr/sbin/chown vagrant /usr/local/.
==> default: ==> /usr/bin/sudo /usr/bin/chgrp admin /usr/local/.
==> default: ==> /usr/bin/sudo /bin/mkdir /Library/Caches/Homebrew
==> default: ==> /usr/bin/sudo /bin/chmod g+rwx /Library/Caches/Homebrew
==> default: ==> /usr/bin/sudo /usr/sbin/chown vagrant /Library/Caches/Homebrew
==> default: ==> Searching online for the Command Line Tools
==> default: ==> Installing Command Line Tools (OS X 10.11) for Xcode-7.2
==> default: ==> Downloading and installing Homebrew...
==> default: From https://github.com/Homebrew/homebrew
==> default:  * [new branch]      master     -> origin/master
==> default: HEAD is now at ec5a0a0 consul: update 0.6.3 bottle.
==> default: ==> Installation successful!
==> default: ==> Next steps
==> default: Run `brew help` to get started
```

tdsmith:
I like what this does (and thanks for demonstrating how to test it!) but I worry that this is depending on implementation details in third-party software. I assume that anything that's deploying Homebrew automatically will have already taken care of deploying the CLT before Homebrew (and I'm happier to let that break on their end than ours).

marcosscriven:
I understand your concern; there's two risks:

* That the temp file no longer prompts ```softwareupdate``` to list the CLTs at all
* That even if listed, the regexp may not match due to differing output.

However, would you be averse to a strategy of structuring the code to try this first, and falling back to the old method if it didn't work? The way I see it is that the brew install does after all need and attempt to install the Command Line Tools. If we can make that zero interaction by default, but fall back to the 'click here' approach, that might be the best of both worlds?

mikemcquaid:
@tdsmith Boxen has been using this method quite a while so I'm not worried about the "implementation details" side of it (although it does perhaps warrant a `xcode-select --install` fallback).

install
```diff
@@ -201,10 +201,12 @@ sudo "/usr/bin/chgrp", "admin", HOMEBREW_CACHE if chgrp? HOMEBREW_CACHE
 if macos_version >= "10.9"
   developer_dir = `/usr/bin/xcode-select -print-path 2>/dev/null`.chomp
   if developer_dir.empty? || !File.exist?("#{developer_dir}/usr/bin/git")
-    ohai "Installing the Command Line Tools (expect a GUI popup):"
-    sudo "/usr/bin/xcode-select", "--install"
-    puts "Press any key when the installation has completed."
-    getc
+    ohai "Searching online for the Command Line Tools"
+    # This temporary file prompts software update to list the Command Line Tools
+    `touch /tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress`
```
mikemcquaid: Use the `sudo` method for this.

install
```diff
@@ -201,10 +201,12 @@ sudo "/usr/bin/chgrp", "admin", HOMEBREW_CACHE if chgrp? HOMEBREW_CACHE
...
+    clt = `softwareupdate -l | grep "* Command Line Tools" | sed 's/^.*\\(Command Line Tools.*\\)/\\1/'`.chomp
```
mikemcquaid: This may be more robust: https://github.com/mikemcquaid/strap/blob/78e162d5d239d1bfb5788d4dd9fb8bb7222fac13/bin/strap.sh#L125-L127

install
```diff
@@ -201,10 +201,12 @@ sudo "/usr/bin/chgrp", "admin", HOMEBREW_CACHE if chgrp? HOMEBREW_CACHE
...
+    ohai "Installing #{clt}"
+    `softwareupdate -i "#{clt}"`
```
mikemcquaid: Need `sudo` for this and also need to delete the placeholder file after successful installation. It'd be good to fall back to `xcode-select --install` if the installation wasn't successful

install
marcosscriven: Happy to do so, but could you expand on why it's needed please? ```/tmp``` has open perms, so perhaps you mean ```softwareupdate``` in certain cases doesn't like it if the tmp file is not owned by root?

install
marcosscriven: Interesting - curious why you take the line before (```B1```) in your grep, and also where you've seen ```Developer``` in the name?
Strap itself looks interesting, will definitely look into that!

install
marcosscriven: Sure, will add these two.

install
mikemcquaid: And also just it feels better if it's about to be used by a root process that we ensure it won't be accidentally removed in the meantime by a user process.

install
mikemcquaid: Thnks about Strap. The `Developer` applies on 10.9, if I remember correctly.

install
mikemcquaid: :+1:

marcosscriven:
Do you have any suggestions on getting this aspect into CI? I note you already uninstall brew before testing, because brew's already on the Travis OS X images :)

So perhaps have a 4th job that also uninstalls xcode first?

mikemcquaid:
> Do you have any suggestions on getting this aspect into CI? I note you already uninstall brew before testing, because brew's already on the Travis OS X images :)

Again linking to Strap because I do that: https://github.com/mikemcquaid/strap/blob/78e162d5d239d1bfb5788d4dd9fb8bb7222fac13/.travis.yml#L17-L20

marcosscriven:
Ah, nice. 

I guess with brew install we want to test both with and without xcode available. Not sure if you want to double up all the jobs (E.g. 10.9 to 10.11, both with and without xcode)?
I can either leave that to you or submit an update to the Travis config, don't mind which.

mikemcquaid:
> I can either leave that to you or submit an update to the Travis config, don't mind which.

I think it'd be nicer to just verify the installation on all platforms. I think it's fine not testing with Xcode available.

marcosscriven:
One other thing I just though - say the the headless update fails in CI, and it falls back to the GUI. 

I assume we count that as a failure? I'm also under the impression Travis will deal with this situation automatically by virtue of the 10 minute timeout: https://docs.travis-ci.com/user/build-timeouts

mikemcquaid:
> I assume we count that as a failure? I'm also under the impression Travis will deal with this situation automatically by virtue of the 10 minute timeout: https://docs.travis-ci.com/user/build-timeouts

We can, yes. We can also use the `if STDIN.tty?` to do something smarter.

marcosscriven:
Explanation of my latest commit:

1. ```softwareupdate``` unfortunately only returns a zero exit code (regardless of listing or installation). Rather than complicate the code with conditionals (both at the listing point and the attempted install), instead split out the test for whether to install the command line tools into a function, and simply check again.
2. Used ```sudo``` where suggested.
3. Cleaned up the temp file

I've left the ```Installing the Command Line Tools (expect a GUI popup):``` output untouched, so it really should behave as before if the headless install fails.

marcosscriven:
Also, I used @mikemcquaid's regex verbatim (aside from escaping the backslash in the awk command for Ruby), as that's clearly been tried and tested in his ```Strap``` script, and as he mentioned, in Boxen: https://github.com/boxen/boxen-web/blob/master/app/views/splash/script.sh.erb#L56

marcosscriven:
Lastly, I've gone ahead and changed the travis config to remove ```Xcode``` and ```CommandLineTools``` first.

I tried such a build *without* the headless install, and the prompt for input in the Travis environment ensures it fails immediately anyway:

```
Press any key when the installation has completed.
stty: stdin isn't a terminal
```

.travis.yml
```diff
@@ -15,6 +15,11 @@ matrix:
       osx_image: beta-xcode6.2
       rvm: system
 
+# Remove Xcode and CommandLineTools included in default OS X image.
+# This ensures the Command Line Tool install functionality is tested.
+before_script:
+  - sudo rm -rf /Applications/Xcode.app /Library/Developer/CommandLineTools
```
mikemcquaid: Probably want to also remove that other file I remove for the package manifest, just to be extra safe.

install
```diff
@@ -198,14 +207,24 @@ sudo "/bin/chmod", "g+rwx", HOMEBREW_CACHE if chmod? HOMEBREW_CACHE
...
-  end
+if should_install_command_line_tools
+  ohai "Searching online for the Command Line Tools"
+  # This temporary file prompts the 'softwareupdate' utility to list the Command Line Tools
+  clt_placeholder = "/tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress"
+  sudo "/usr/bin/touch", clt_placeholder
+  clt_label = `softwareupdate -l | grep -B 1 -E "Command Line (Developer|Tools)" | awk -F"*" '/^ +\\*/ {print $2}' | sed 's/^ *//' | head -n1`.chomp
+  ohai "Installing #{clt_label}"
+  sudo "/usr/sbin/softwareupdate", "-i", clt_label
+  sudo "/bin/rm", clt_placeholder
```
mikemcquaid: Maybe do `-f` just so if it's missing this doesn't fail.

install
```diff
@@ -198,14 +207,24 @@ sudo "/bin/chmod", "g+rwx", HOMEBREW_CACHE if chmod? HOMEBREW_CACHE
...
+end
+
+# Headless install may have failed, so fallback to original 'xcode-select' method
+if should_install_command_line_tools
+  ohai "Headless installation of Command Line Tools has failed"
```
mikemcquaid: Think you can skip this line.

install
```diff
@@ -198,14 +207,24 @@ sudo "/bin/chmod", "g+rwx", HOMEBREW_CACHE if chmod? HOMEBREW_CACHE
...
+  ohai "Installing the Command Line Tools (expect a GUI popup):"
+  sudo "/usr/bin/xcode-select", "--install"
+  puts "Press any key when the installation has completed."
+  getc
```
mikemcquaid: I know this isn't your bug but this should also probably check for a `tty`.

marcosscriven: Sure I can add that, but then what about CI? The fact that fails at the moment conveniently fails the CI if the headless install fails.

install
marcosscriven: My thinking here was avoiding confusion as to why it was trying a second time. I'm not greatly attached to it though, so happy to drop it.

.travis.yml
marcosscriven: To be sure, this one? 

https://github.com/mikemcquaid/strap/blob/78e162d5d239d1bfb5788d4dd9fb8bb7222fac13/.travis.yml#L20

install
mikemcquaid: I think it's desirable to have headless installs fail more gracefully in that situation. We should still do an `exit 1` but just with a better error message.

.travis.yml
mikemcquaid: Yup :+1:

install
marcosscriven: Gotcha. Will fix that up.

marcosscriven:
Tested out the case where automatic install fails, by deliberately breaking the regex on a separate branch:

https://travis-ci.org/marcosscriven/install/jobs/104639434#L2491

marcosscriven:
Think I've covered all review points thus far, and the updated build it passing. Let me know if there's any more changes to make, and thanks for reviewing.

install
```diff
@@ -77,6 +77,15 @@ def macos_version
   @macos_version ||= Version.new(`/usr/bin/sw_vers -productVersion`.chomp[/10\.\d+/])
 end
 
+def should_install_command_line_tools
+  if macos_version >= "10.9"
+    developer_dir = `/usr/bin/xcode-select -print-path 2>/dev/null`.chomp
+    return developer_dir.empty? || !File.exist?("#{developer_dir}/usr/bin/git")
```
mikemcquaid: Don't need the `return` here; it's implicit in Ruby.

install
```diff
@@ -77,6 +77,15 @@ def macos_version
...
+  else
+    return false
```
mikemcquaid: Don't need the `return` here; it's implicit in Ruby. Also might be nicer to just do `return false if macos_version < "10.9"` above?

install
```diff
@@ -77,6 +77,15 @@ def macos_version
   @macos_version ||= Version.new(`/usr/bin/sw_vers -productVersion`.chomp[/10\.\d+/])
 end
 
+def should_install_command_line_tools
```
mikemcquaid: Typical Ruby style would have this method be called `should_install_command_line_tools?`

install
```diff
@@ -198,13 +207,26 @@ sudo "/bin/chmod", "g+rwx", HOMEBREW_CACHE if chmod? HOMEBREW_CACHE
...
+
+# Headless install may have failed, so fallback to original 'xcode-select' method
+if should_install_command_line_tools
+  if STDIN.tty?
     ohai "Installing the Command Line Tools (expect a GUI popup):"
     sudo "/usr/bin/xcode-select", "--install"
     puts "Press any key when the installation has completed."
     getc
+  else
+    abort "Unable to proceed with manual Command Line Tools install without user input."
```
mikemcquaid: Maybe prefix with `Error:`, suffix with `!` and make sure it fits in 80 chars.

mikemcquaid:
A few final nits and this is good to :ship:!

marcosscriven:
I shall sort out these final things in the morning (GMT). Thanks again for your patient review points, and the impromptu Ruby lessons!

install
marcosscriven: Do you only apply the 80 char limit to the pure Ruby lines? There's a few shell script calls which exceed that, including the one I added parsing the output of ```softwareupdate -l```.

install
marcosscriven: You crazy Ruby peeps :)

install
mikemcquaid: In my perfect world none of them would exceed that but the important thing is that the string `Error: unable to proceed with manual Command Line Tools install without user input!` is less than 80 characters as Bash's prose wrapping leaves a bit to be desired.

install
mikemcquaid: :grinning:

marcosscriven:
Just pinging this one to see if you're happy with it to be merged?

mikemcquaid:
Yup, thanks. Just FYI we don't get notified on commits so I missed this.

marcosscriven:
Ah, cool. Thanks again for the review points @mikemcquaid

PR #37, opened 2016-03-29
install: set autocrlf to false

xu-cheng:


ilovezfs:
What is the reasoning for false vs. input?

xu-cheng:
> What is the reasoning for false vs. input?

We are using false in update.sh. So it's more consistence.

Issue #39, opened 2016-04-08
Directory permissions not modified during install


justsostephen:
When running `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`, the install fails and produces the following message:

`error: unable to create file etc/bash_completion.d/brew (Permission denied)`
`error: unable to create file share/zsh/site-functions/_brew (Permission denied)`
`fatal: Could not reset index file to revision 'origin/master'.`
`Failed during: git reset --hard origin/master`

On my system, the directories `/usr/local/etc/bash_completion.d`, `/usr/local/share/zsh` and `/usr/local/share/zsh/site-functions` were created during an install of Rust with the permissions `rwxr-xr-x`, owner `root` and group `wheel`. Unlike the other directories used by Homebrew, the permissions and ownership of these directories are not modified for user level access during install, which leads to the error above. Modifying them manually before installing Homebrew fixes this.

Should the aforementioned directories be included in the `chmods` array on line 165 of the install script along with the other directories used, so their permissions and ownership are modified automatically during install?

ilovezfs:
Yes, I think that's the logical consequence of the change that was made here: https://github.com/Homebrew/brew/commit/edf000e4cd30c3626ccc28c52ed32f2d84a200dd
@xu-cheng

justsostephen:
Resolved by merged PR #40.

PR #40, opened 2016-04-08
Added shell completion script paths to chmods/chowns/chgrps array.

justsostephen:
My proposed fix for issue #39, where the destination directories for the shell completion scripts do not have their permissions modified for user level access during the installation process, as with the other install directories.


mikemcquaid:
CC @xu-cheng than thoughts

xu-cheng:
LGTM

mikemcquaid:

justsostephen:
You're very welcome! Thanks for running an awesome project. You also rock!

PR #42, opened 2016-04-25
Notify about analytics more explicitly.

mikemcquaid:
And avoid duplicating the message unnecessarily at the first `brew update`.

References https://github.com/Homebrew/brew/issues/142.

### DIFF
--- a/install
+++ b/install
@@ -269,4 +269,11 @@ end
 
 puts "Run `brew help` to get started"
 puts "Further documentation: https://git.io/brew-docs"
-puts "Anonymous aggregate user behaviour analytics documentation: https://git.io/brew-analytics"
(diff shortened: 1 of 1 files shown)


DomT4:
👍.

NickCraver:
Opting out shouldn't require all of the effort still involved here. For most users, the steps will be:
1. Opening a browser
2. Go to that link
3. Read a 5KB file (to the very bottom)
4. (likely) Learn how to set an environmental variable
5. Create a `.bash_profile`
6. Set the variable
...and hope it worked. 

The bar is still far, far too high. There should be something akin to `brew metrics off` as pitched in https://github.com/Homebrew/brew/issues/142

PR #45, opened 2016-06-25
Use `HOMEBREV_PREFIX` env variable if set

maarons:
I'd like to be able to easily install Homebrew to a custom location in
a non-interactive way (like Chef actually doing the work).  Current
method of editing the install file is not really suited for automated
use.

### DIFF
--- a/install
+++ b/install
@@ -1,8 +1,8 @@
 #!/System/Library/Frameworks/Ruby.framework/Versions/Current/usr/bin/ruby
 # This script installs to /usr/local only. To install elsewhere you can just
 # untar https://github.com/Homebrew/brew/tarball/master anywhere you like or
-# change the value of HOMEBREW_PREFIX.
-HOMEBREW_PREFIX = '/usr/local'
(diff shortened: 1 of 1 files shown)


ilovezfs:
If this were something we were going to support, I think it would need to be via a command-line switch and/or a configuration file.

MikeMcQuaid:
We don't support installing outside /usr/local in the installer, sorry.

maarons:
Any particular reason you don't want to support that?

MikeMcQuaid:
People don't get binary packages (which means more issues get filed), some software doesn't exist.

mareksapota-lyft:
@MikeMcQuaid got it.  Maybe the FAQ should get updated, it doesn't really discourage prefix use and it seems it should.

maarons:
@MikeMcQuaid got it. Maybe the FAQ should get updated, it doesn't really discourage prefix use and it seems it should.

MikeMcQuaid:
@maarons good idea but we already provide some gentle discouragement through the installer and the `brew doctor`. Thanks!

PR #47, opened 2016-07-14
install: use curl to install core tap if git is unavailable

xu-cheng:


install
```diff
@@ -243,18 +244,24 @@ Dir.chdir HOMEBREW_PREFIX do
     system(*args)
 
     system git, "reset", "--hard", "origin/master"
+
+    # no analytics during installation
+    ENV["HOMEBREW_NO_ANALYTICS_THIS_RUN"] = "1"
```
MikeMcQuaid: Maybe just worth setting this super early in this file so if other `brew` commands are added outside of this `if` then they won't have analytics run.

MikeMcQuaid:
One nit otherwise 👍

PR #53, opened 2016-08-10
Replace chmod with chown in post-uninstall message

Eitot:
`/usr/local` belongs to root by default and the Homebrew installation changes this. The post-uninstallation message omits this information.


MikeMcQuaid:
Seems sensible. Thanks for your contribution to Homebrew! Without people like you submitting PRs we couldn't run this project. You rock!

PR #55, opened 2016-08-18
install: tweak constants.

MikeMcQuaid:
Rename and freeze them for future changes.


DomT4:
Seems reasonable.

ilovezfs:
Would it be worth warning about the deprecation if we see the old variable defined in the environment?

MikeMcQuaid:
@ilovezfs We're not actually doing anything with the environment here, just changing naming within this file.

ilovezfs:
@MikeMcQuaid oh.

PR #59, opened 2016-09-03
Create all root directories.

MikeMcQuaid:
This stops issues with them being created at a later point after e.g. the `/usr/local` permissions may have changed.

Also, don't make ZSH's directories group writable as ZSH complains.


MikeMcQuaid:
@DomT4 No rush but: thoughts on this? It's a mid-way point to changing the installed to defaulting to a non-/usr/local `HOMEBREW_REPOSITORY` and means we should hopefully never have to do the "something resets `/usr/local` permissions" dance again.

DomT4:
Might want to include `Frameworks` in the created directories as well potentially, but changes look reasonable 👍.

PR #60, opened 2016-09-08
install: default repo to /usr/local/Homebrew.

MikeMcQuaid:
This means that we can move stuff around inside the repository without causing conflicts on `update` or dumping things in `/usr/local` but keep all our existing bottles working as-is.

CC @Homebrew/maintainers for thoughts because this is a pretty big change.

### DIFF
--- a/install
+++ b/install
@@ -3,7 +3,7 @@
 # untar https://github.com/Homebrew/brew/tarball/master anywhere you like or
 # change the value of HOMEBREW_PREFIX.
 HOMEBREW_PREFIX = "/usr/local".freeze
-HOMEBREW_REPOSITORY = HOMEBREW_PREFIX
+HOMEBREW_REPOSITORY = "/usr/local/Homebrew".freeze
 HOMEBREW_CACHE = "#{ENV["HOME"]}/Library/Caches/Homebrew".freeze
 BREW_REPO = "https://github.com/Homebrew/brew".freeze
 CORE_TAP_REPO = "https://github.com/Homebrew/homebrew-core".freeze
@@ -155,24 +155,24 @@ EOABORT
 
 ohai "This script will install:"
 puts "#{HOMEBREW_PREFIX}/bin/brew"
-puts "#{HOMEBREW_PREFIX}/share/doc/homebrew"
-puts "#{HOMEBREW_PREFIX}/share/man/man1/brew.1"
-puts "#{HOMEBREW_PREFIX}/share/zsh/site-functions/_brew"
-puts "#{HOMEBREW_PREFIX}/etc/bash_completion.d/brew"
-puts "#{HOMEBREW_REPOSITORY}/Library/..."
+if git
+  puts "#{HOMEBREW_PREFIX}/share/doc/homebrew"
+  puts "#{HOMEBREW_PREFIX}/share/man/man1/brew.1"
+  puts "#{HOMEBREW_PREFIX}/share/zsh/site-functions/_brew"
+  puts "#{HOMEBREW_PREFIX}/etc/bash_completion.d/brew"
+end
+puts "#{HOMEBREW_REPOSITORY}"
 
-group_chmods = %w( . bin etc etc/bash_completion.d Frameworks include lib lib/pkgconfig Library sbin share var var/log share/locale share/man
-                   share/man/man1 share/man/man2 share/man/man3 share/man/man4
-                   share/man/man5 share/man/man6 share/man/man7 share/man/man8
-                   share/info share/doc share/aclocal share/zsh share/zsh/site-functions ).
+group_chmods = %w( bin etc Frameworks include lib sbin share var ).
                    map { |d| File.join(HOMEBREW_PREFIX, d) }.select { |d| chmod?(d) }
 # zsh refuses to read from these directories if group writable
-user_chmods = %w( share/zsh share/zsh/site-functions ).
+zsh_dirs = %w( share/zsh share/zsh/site-functions )
+user_chmods = zsh_dirs.
                   map { |d| File.join(HOMEBREW_PREFIX, d) }.select { |d| chmod?(d) }
 chmods = group_chmods + user_chmods
 chowns = chmods.select { |d| chown?(d) }
 chgrps = chmods.select { |d| chgrp?(d) }
-mkdirs = %w( Cellar Frameworks bin etc include lib opt sbin share var ).
+mkdirs = %w( Cellar Homebrew Frameworks bin etc include lib opt sbin share share/zsh share/zsh/site-functions var ).
              map { |d| File.join(HOMEBREW_PREFIX, d) }.reject { |d| File.directory?(d) }
 
 unless group_chmods.empty?
@@ -205,14 +205,13 @@ if File.directory? HOMEBREW_PREFIX
   sudo "/usr/bin/chgrp", "admin", *chgrps unless chgrps.empty?
 else
   sudo "/bin/mkdir", "-p", HOMEBREW_PREFIX
-  sudo "/bin/chmod", "g+rwx", HOMEBREW_PREFIX
-  # the group is set to wheel by default for some reason
-  sudo "/usr/sbin/chown", "#{ENV['USER']}:admin", HOMEBREW_PREFIX
+  sudo "/usr/sbin/chown", "root:wheel", HOMEBREW_PREFIX
 end
 
 unless mkdirs.empty?
   sudo "/bin/mkdir", "-p", *mkdirs
   sudo "/bin/chmod", "g+rwx", *mkdirs
+  sudo "/bin/chmod", "u+rwx", *zsh_dirs
   sudo "/usr/sbin/chown", ENV['USER'], *mkdirs
   sudo "/usr/bin/chgrp", "admin", *mkdirs
 end
@@ -273,13 +272,19 @@ Dir.chdir HOMEBREW_REPOSITORY do
 
     system git, "reset", "--hard", "origin/master"
 
+    system "ln", "-sf", "#{HOMEBREW_REPOSITORY}/bin/brew", "#{HOMEBREW_PREFIX}/bin/brew"
+
     system "#{HOMEBREW_PREFIX}/bin/brew", "tap", "homebrew/core"
+
+    system "#{HOMEBREW_PREFIX}/bin/brew", "update"
   else
     # -m to stop tar erroring out if it can't modify the mtime for root owned directories
     # pipefail to cause the exit status from curl to propagate if it fails
     curl_flags = "fsSL"
     core_tap = "#{HOMEBREW_PREFIX}/Library/Taps/homebrew/homebrew-core"
     system "/bin/bash -o pipefail -c '/usr/bin/curl -#{curl_flags} #{BREW_REPO}/tarball/master | /usr/bin/tar xz -m --strip 1'"
+
+    system "ln", "-sf", "#{HOMEBREW_REPOSITORY}/bin/brew", "#{HOMEBREW_PREFIX}/bin/brew"
 
     system "/bin/mkdir", "-p", core_tap
     Dir.chdir core_tap do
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -205,14 +203,13 @@ if File.directory? HOMEBREW_PREFIX
   sudo "/usr/bin/chgrp", "admin", *chgrps unless chgrps.empty?
 else
   sudo "/bin/mkdir", "-p", HOMEBREW_PREFIX
-  sudo "/bin/chmod", "g+rwx", HOMEBREW_PREFIX
-  # the group is set to wheel by default for some reason
-  sudo "/usr/sbin/chown", "#{ENV['USER']}:admin", HOMEBREW_PREFIX
+  sudo "/usr/sbin/chown", "root:wheel", HOMEBREW_PREFIX
 end
```
tdsmith: What's your thinking here?

tdsmith:
I like this change. Can we think of any circumstances where someone is using `$(brew --prefix)` where it turns out they meant `$(brew --repository)` and this could cause trouble? git commands should fail benignly, which seems okay.

install
tdsmith: ah, cf https://github.com/Homebrew/brew/pull/896

apjanke:
Seems like a good idea, especially with Apple moving to lock down `/usr/local` more. Keeps things cleaner. Will need a bit of time to think if there are repercussions to this, but don't see any off the bat.

DomT4:
👍 on the idea. Owning less of `/usr/local` is definitely a good step. Haven't tested the implementation yet, but the idea is a reasonable one.

zmwangx:
```ruby
ohai "This script will install:"
puts "#{HOMEBREW_PREFIX}/bin/brew"
puts "#{HOMEBREW_PREFIX}/share/doc/homebrew"
puts "#{HOMEBREW_PREFIX}/share/man/man1/brew.1"
puts "#{HOMEBREW_PREFIX}/share/zsh/site-functions/_brew"
puts "#{HOMEBREW_PREFIX}/etc/bash_completion.d/brew"
puts "#{HOMEBREW_REPOSITORY}/Library/..."
```

We should be installing all these (except the last, which should probably be replaced with `HOMEBREW_REPOSITORY` alone), but I'm only seeing a `ln -sf` for `bin/brew`.

Also, after this change we're going to see weird paths like `/usr/local/Homebrew/Library/Homebrew`, right? Do we still need the `Homebrew` subdirectory in `Library` after this change? Of course migration might be an issue, but otherwise the subdirectory seems useless.

install
MikeMcQuaid: @tdsmith Yep. Basically: that's the default permissions so if this directory doesn't exist, let's restore them to avoid any weird changes in future.

MikeMcQuaid:
> Can we think of any circumstances where someone is using $(brew --prefix) where it turns out they meant $(brew --repository) and this could cause trouble? git commands should fail benignly, which seems okay.

Yeh, this is somewhat possible but unavoidable. Thankfully the case where they use `$(brew --prefix)` but meant `$(brew --repo)` means they'll be e.g. looking for core Homebrew files rather than e.g. the Cellar or installed files so it's a lot more "implementation-detaily" (even if those have been documented correctly for a long time).

Oh, also: well done @chdiza, you changed the Homebrew default 😀

MikeMcQuaid:
> We should be installing all these (except the last, which should probably be replaced with HOMEBREW_REPOSITORY alone), but I'm only seeing a ln -sf for bin/brew.

They are all installed by the later `brew update`. We could do that manually but I'm kinda meh on the code duplication and tying more installer code to locations in Homebrew's repository. Agreed about changing the last, will do that.

> Also, after this change we're going to see weird paths like /usr/local/Homebrew/Library/Homebrew, right? Do we still need the Homebrew subdirectory in Library after this change? Of course migration might be an issue, but otherwise the subdirectory seems useless.

Yep. A bit part of the motivation for this change is being able to move around the contents of our repository without dumping crap in `/usr/local`'s root any more. We may well be able to move all that stuff to `lib` or `src` or similar.

zmwangx:
> They are all installed by the later brew update.

Oh, good to know, that's good enough.

MikeMcQuaid:
@zmwangx Improved messaging in 2182f13

ilovezfs:
>especially with Apple moving to lock down /usr/local more.

@apjanke what are you referring to here?

MikeMcQuaid:
@ilovezfs If nothing else: they reset permissions on every major OS upgrade which is a) annoying b) a bit of a sign that they don't consider our permission changes there to be desirable.

ilovezfs:
Pretty sure that's basically unintentional since /usr/local doesn't even exist in the default installation.

In any case, I'm not sure what's been achieved with respect to /usr/local unless the actual default non-relocatable bottle installation path changes. But increasing encapsulation of the brew repo is a sufficient reason even if I see nothing else being achieved.

MikeMcQuaid:
> In any case, I'm not sure what's been achieved with respect to /usr/local unless the actual default non-relocatable bottle installation path changes.

Yep, we're not changing the default there so there will still be `/usr/local/Cellar` and respective symlinks in `/usr/local/bin` but we just punt the repo elsewhere.

zmwangx:
> they don't consider our permission changes there to be desirable.

Well of course, and I believe that's for good reasons. `/usr/local/bin` is in `/etc/paths` right before `/usr/bin` and `/bin`, so if `/usr/local/bin` is owned by `johnappleseed`, any program running as `johnappleseed` can write to `/usr/local/bin` and mask basic utilities in `/usr/bin` and `/bin`, which is a security threat. If I see a system with `/usr/local` owned by `johnappleseed` and I don't know John installed Homebrew, I would presume it's done by an attacker.

I might be wrong but I think even repairing disk permissions will restore ownership of `/usr/local`.

MikeMcQuaid:
> any program running as johnappleseed can write to /usr/local/bin

This will remain.

zmwangx:
> This will remain.

I'm not talking about changing Homebrew's default. Just saying from Apple's point of view it's certainly not desirable.

ilovezfs:
>Yep, we're not changing the default there so there will still be /usr/local/Cellar and respective symlinks in /usr/local/bin but we just punt the repo elsewhere.

FWIW, I think it would be good to start making a list somewhere of potential new locations so people can start cogitating on what actually is going to work best.

MikeMcQuaid:
> FWIW, I think it would be good to start making a list somewhere of potential new locations so people can start cogitating on what actually is going to work best.

I don't think that's going to change any time soon (perhaps ever) because it'd require rebuilding all our bottles and then remove us from the default search path from many other tools.

ilovezfs:
>I don't think that's going to change any time soon (perhaps ever)

I bet this "suddenly" becomes a priority for some unexpected reason. :crystal_ball:

chdiza:
Sweet!

I guess I would like it to be `/usr/local/homebrew` and not with a capital "H".  Seeing things starting with capitals in `/usr/local` is really weird-looking.  E.g., the capital "L" in `/usr/local/Library` has always bugged me.  Other software stays lowercase when they dive into `/usr/local`, e.g., `/usr/local/ssl`.

DomT4:
> because it'd require rebuilding all our bottles

If we had the resources I'd like to bottle/test two default prefixes where the bottle wasn't `cellar :any` but realistically implementing that with the current CI is just a path to pain.

> E.g., the capital "L" in /usr/local/Library has always bugged me.

`/usr/local/Frameworks` goes in with a capital as well. There's also `/usr/local/Caskroom` if your Homebrew repo == Homebrew prefix.

MikeMcQuaid:
`Homebrew` is less likely to clash with existing locations, in my experience. Regardless: it's done now 😀 

> the capital "L" in /usr/local/Library has always bugged me

It'll likely change in future but the previous reasoning was because you have `~/Library`

chdiza:
> /usr/local/Frameworks goes in with a capital as well. There's also /usr/local/Caskroom if your Homebrew repo == Homebrew prefix.

`Frameworks` at least has an excuse for the capitalization (to be consistent with /Library/Frameworks and /System/Library/Frameworks, and probably some software is expecting a capital).  But `Homebrew`, `Library`, `Caskroom`, `Cellar` have no such excuse.  (`Library` should never have been used in the first place, upper *or* lowercase, since it's HB-specific and not some general dumping ground for pieces of many programs.  I think HB invented that location.)

> Homebrew is less likely to clash with existing locations, in my experience

What would already exist in `/usr/local/homebrew`, that `/usr/local/Homebrew` would avoid clashing with?

chdiza:
In any case (heh heh), this is a nice change! :smile:

apjanke:
> >especially with Apple moving to lock down /usr/local more.

> @apjanke what are you referring to here?

I just meant the SIP introduced with El Capitan. Nothing new in Sierra AFAIK.

ilovezfs:
@MikeMcQuaid does this PR have any consequences for what should/shouldn't be in https://github.com/Homebrew/brew/blob/master/.gitignore ?

MikeMcQuaid:
@ilovezfs Unfortunately not until at least after we do the migration and possibly never if we allow people to also keep `HOMEBREW_PREFIX` and `HOMEBREW_REPOSITORY` the same. I hate that `.gitignore` with the 🔥 of a thousand ☀️.

ilovezfs:
Yeah that .gitignore freaks me out every time I look at it.

MikeMcQuaid:
@ilovezfs The best/worst thing is that `libgit2` doesn't consistently support it meaning many other Git utilities don't support it.

ilovezfs:
Oh that's just stellar.

DomT4:
> Unfortunately not until at least after we do the migration and possibly never if we allow people to also keep HOMEBREW_PREFIX and HOMEBREW_REPOSITORY the same.

Personal temptation is for us to do a hard migration without an opt-out.

MikeMcQuaid:
@DomT4 Sorry, I mean for people installed in non-`/usr/local` locations unless we want to start having "HOMEBREW_REPOSITORY is always in a HOMEBREW_PREFIX subdirectory"?

sandstrom:
This is great but why not use `/usr/local/homebrew`? (lowercase)

MikeMcQuaid:
@sandstrom Personal preference and to match `Cellar` (which we can't change).

zmwangx:
Also, look at the name of the organization ;)

sandstrom:
Homebrew is awesome so I won't argue (even though I disagree 😃 ). Is it configurable?

the-real-tokai:
I'm also in favour of a small letter 'homebrew', in fact I use it like that since the very beginning:

```
$ brew config
HOMEBREW_VERSION: 0.9.9
ORIGIN: https://github.com/Homebrew/brew.git
...
HOMEBREW_PREFIX: /usr/local
HOMEBREW_REPOSITORY: /usr/local/homebrew
HOMEBREW_CELLAR: /usr/local/homebrew/Cellar
HOMEBREW_BOTTLE_DOMAIN: https://homebrew.bintray.com
...
```

```
$ cd /usr/local/
/usr/local $ ls -l
total 8
drwxr-xr-x  769 root  wheel  26146 Aug 31 21:11 bin
drwxr-xr-x   23 root  wheel    782 Aug 20 18:59 etc
drwxrwxr-x   15 root  wheel    510 Sep 10 16:29 homebrew
...

/usr/local $ ls -l homebrew/
total 32
-rw-r--r--    1 root  wheel  3161 Apr  9 18:20 CODEOFCONDUCT.md
drwsr-sr-x  133 root  wheel  4522 Aug 31 21:11 Cellar
-rw-r--r--    1 root  wheel  1241 Dec 31  2015 LICENSE.txt
drwxr-xr-x    8 root  wheel   272 Aug 26 09:32 Library
-rw-r--r--    1 root  wheel  5673 Sep 10 16:29 README.md
drwxr-xr-x    3 root  wheel   102 Aug 26 09:32 bin
drwxr-xr-x    3 root  wheel   102 Apr  9 18:20 etc
drwxr-xr-x    5 root  wheel   170 Apr  9 18:20 share
```

I'm very happy with that setup, so I'm rather afraid to `brew update`...

Will it break/ mess up things? The to-be-killed root "support" is dangling like a Damocles sword too (but no way in hell I'll ever have any user owned files in /usr/local/ on a machine that is shared by multiple accounts) 😱

zmwangx:
@sandstrom 

> Is it configurable?

On a case-sensitive filesystem:

```
> sudo mv /usr/local/Homebrew /usr/local/homebrew
> ln -sf /usr/local/Homebrew/bin/brew /usr/local/bin/brew
> brew update --force
```

On a case-insensitive but case-preserving filesystem:

```
> sudo mv /usr/local/Homebrew /usr/local/Homebrew.bak
> sudo mv /usr/local/Homebrew.bak /usr/local/homebrew
```

On a case-insensitive but case non-preserving filesystem:

Excuse me?

sandstrom:
@zmwangx thanks! I was thinking more of an install flag/option, a prompt when installing or a cli message upon install pointing to docs/info on how to modify.

zmwangx:
@sandstrom We surely won't recommend that so putting it in the install script is out of question. We do link to https://github.com/Homebrew/brew/blob/master/docs/Installation.md#installation from brew.sh which gives alternative installation instructions, including installing to custom locations. Now `/usr/local/homebrew` isn't really a custom location (it could be when APFS ships, we don't know if it will be case-insensitive by default just yet), but still, cloning the git repo will always work.

PR #64, opened 2016-09-20
install: use update --force.

MikeMcQuaid:
This will ensure that e.g. the symlinks and any other migrations are done even if there's no updates.

References https://github.com/Homebrew/brew/issues/1045.


zmwangx:
👍

PR #75, opened 2016-10-12
Fix #72 - the group write permissions for zsh dir

### LINKED ISSUE #72: wrong file mode for /usr/local/share/zsh
After installing homebrew on a fresh Sierra Mac, I noticed that homebrew's zsh side-functions don't get loaded.

executing compinit shows
```
zsh compinit: insecure directories, run compaudit for list.
Ignore insecure directories and continue [y] or abort compinit [n]?
```

Running the compaudit returns the follows:
```
There are insecure directories:
/usr/local/share/zsh/site-functions
```

I could solve the problem for me by changing the file mode as follows:

```
sudo chmod -R 755 /usr/local/share/zsh
```

Maybe it is possible to update the installer, if this problem arises for many users.

dreambottle:
This should fix #72

### DIFF
--- a/install
+++ b/install
@@ -192,9 +192,9 @@ group_chmods = %w[ bin etc Frameworks include lib sbin share var
                .map { |d| File.join(HOMEBREW_PREFIX, d) }
                .select { |d| chmod?(d) }
 # zsh refuses to read from these directories if group writable
-zsh_dirs = %w( share/zsh share/zsh/site-functions )
-                  .map { |d| File.join(HOMEBREW_PREFIX, d) }
-                  .select { |d| chmod?(d) }
+zsh_dirs = %w[share/zsh share/zsh/site-functions]
+           .map { |d| File.join(HOMEBREW_PREFIX, d) }
+           .select { |d| chmod?(d) }
 user_chmods = zsh_dirs
 chmods = group_chmods + user_chmods
 chowns = chmods.select { |d| chown?(d) }
--- a/install
+++ b/install
@@ -225,13 +225,7 @@ end
 
 wait_for_user if STDIN.tty? && !ENV["TRAVIS"]
 
-if File.directory? HOMEBREW_PREFIX
-  sudo "/bin/chmod", "g+rwx", *group_chmods unless group_chmods.empty?
-  sudo "/bin/chmod", "u+rwx", *user_chmods unless user_chmods.empty?
-  sudo "/bin/chmod", "g-w", *user_chmods unless user_chmods.empty?
-  sudo "/usr/sbin/chown", ENV["USER"], *chowns unless chowns.empty?
-  sudo "/usr/bin/chgrp", "admin", *chgrps unless chgrps.empty?
-else
+unless File.directory? HOMEBREW_PREFIX
   sudo "/bin/mkdir", "-p", HOMEBREW_PREFIX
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -162,8 +162,8 @@ group_chmods = %w( bin etc Frameworks include lib sbin share var
                    map { |d| File.join(HOMEBREW_PREFIX, d) }.select { |d| chmod?(d) }
 # zsh refuses to read from these directories if group writable
 zsh_dirs = %w( share/zsh share/zsh/site-functions )
-user_chmods = zsh_dirs.
-                  map { |d| File.join(HOMEBREW_PREFIX, d) }.select { |d| chmod?(d) }
+                  .map { |d| File.join(HOMEBREW_PREFIX, d) }
```
MikeMcQuaid: Can you explain why `chmod?` was removed? I think it should still be part of this check otherwise you may try and set permissions on directories that don't exist or already have the right permissions.

dreambottle:
It skipped existing directories with wrong permissions (group-writable). This is the reason why I dropped the check entirely.

I think that ideally the filter could be extended to pass the files which are group-writable. But I'm not a ruby expert, and I don't know how to write it.

install
dreambottle: It skipped existing directories with wrong permissions (group-writable). This is the reason why I dropped the check entirely.

I think that ideally the filter could be extended to pass the files which are group-writable. But I'm not a ruby expert, and I don't know how to write it.

dreambottle: @MikeMcQuaid they likely wouldn't exist if it's a fresh new install. But the installer wouldn't repair them in case of re-running it on top of the existing installation.

MikeMcQuaid: Ok, in that case you want to`&&` to the `chmod?` logic rather than removing it to check for additional group permissions.

dreambottle: Hi, I merged this to the latest state in the repo and fixed the coding style. Please check. Regarding your initial concern: if these directories don't exist, the script would create them. As for the situation when the permissions are already ok, I think it wouldn't hurt to apply them again.

install
```diff
@@ -226,24 +225,24 @@ end
 
 wait_for_user if STDIN.tty? && !ENV["TRAVIS"]
 
-if File.directory? HOMEBREW_PREFIX
-  sudo "/bin/chmod", "g+rwx", *group_chmods unless group_chmods.empty?
-  sudo "/bin/chmod", "u+rwx", *user_chmods unless user_chmods.empty?
-  sudo "/usr/sbin/chown", ENV["USER"], *chowns unless chowns.empty?
-  sudo "/usr/bin/chgrp", "admin", *chgrps unless chgrps.empty?
```
MikeMcQuaid: Not sure why you've moved these; no need to run them if `HOMEBREW_PREFIX` didn't exist.

dreambottle: Can you explain why not? I see this flow as:
1. Ensure that HOMEBREW_PREFIX dir is created;
2. Ensure all required dirs in HOMEBREW_PREFIX are in place (create them if not)
3. Ensure all permissions are fine.

Tell me if I am missing some point.

MikeMcQuaid: If HOMEBREW_PREFIX didn't exist none of those directories have the wrong permissions at this stage so, like before, there's only any need to fix up the ZSH directory permissions after they were created or if they already existed (neither applies here).

PR #76, opened 2016-10-12
install: Allow painless CLT-free installation with --force-curl.

woodruffw:
Also fixes the default core-tap installation location when installing with `curl`.

This fixes a lot of the issues in https://github.com/Homebrew/brew/issues/729#issuecomment-253278662, particularly for users trying to bootstrap Homebrew without simultaneously installing the CLT. 

cc @MikeMcQuaid 
ref Homebrew/brew/issues/729

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,8 @@ before_script:
   - sudo pkgutil --forget com.apple.pkg.CLTools_Executables
 
 script:
+  - ./uninstall -f
+  - ./install --force-curl
   - ./uninstall -d
   - ./uninstall -f
   - ./install
--- a/install
+++ b/install
@@ -78,17 +78,24 @@ class Version
   end
 end
 
+def force_curl?
+  ARGV.include?('--force-curl')
+end
+
 def macos_version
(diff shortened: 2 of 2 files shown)


MikeMcQuaid:
@woodruffw Nice! May be worth adding this to the `.travis.yml` file to make sure we don't break it. Additionally, maybe the `abort` on the second `should_install_command_line_tools?` could be removed and we can fall-back to the no CLT/no Git behaviour?

woodruffw:
Sure thing! I'll make those changes presently.

.travis.yml
```diff
@@ -11,6 +11,7 @@ before_script:
   - sudo pkgutil --forget com.apple.pkg.CLTools_Executables
 
 script:
+  - ./install --force-curl
```
MikeMcQuaid: Think it's probably worth `uninstall -f`ing before this as otherwise we're not testing a clean install.

woodruffw: Fixed!

MikeMcQuaid:
Great work, thanks @woodruffw!

PR #78, opened 2016-10-20
Create LICENSE.txt

elgalu:
Taken from https://github.com/Homebrew/homebrew-core/blob/master/LICENSE.txt


MikeMcQuaid:
Good idea, thanks!

PR #80, opened 2016-10-23
fix zsh dir permissions #72

dreambottle:
Changed as little as possible here

### DIFF
--- a/install
+++ b/install
@@ -142,8 +142,8 @@ def user_only_chmod?(d)
   mode = File.stat(d).mode & 0b111_111_111
   # u = (mode >> 6) & 0b111
   # g = (mode >> 3) & 0b111
-  # o = (mode >> 0) & 0b111 
-  return mode != 0o755
+  # o = (mode >> 0) & 0b111
+  mode != 0755
 end
 
 def chmod?(d)
--- a/install
+++ b/install
@@ -139,10 +139,10 @@ end
 
 def user_only_chmod?(d)
   return false unless File.directory?(d)
-  mode = File.stat(d).mode & 0b111_111_111
-  # u = (mode >> 6) & 0b111
-  # g = (mode >> 3) & 0b111
-  # o = (mode >> 0) & 0b111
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -137,6 +137,15 @@ def git
   @git
 end
 
+def user_only_chmod?(d)
+  return false unless File.directory?(d)
+  mode = File.stat(d).mode & 0b111_111_111
```
MikeMcQuaid: Any way you can do this with non-binary e.g. octal or even hex?

install
```diff
@@ -193,10 +202,9 @@ group_chmods = %w[ bin etc Frameworks include lib sbin share var
                .select { |d| chmod?(d) }
 # zsh refuses to read from these directories if group writable
 zsh_dirs = %w[share/zsh share/zsh/site-functions]
-user_chmods = zsh_dirs
-              .map { |d| File.join(HOMEBREW_PREFIX, d) }
-              .select { |d| chmod?(d) }
-chmods = group_chmods + user_chmods
+           .map { |d| File.join(HOMEBREW_PREFIX, d) }
+user_only_chmods = zsh_dirs.select { |d| user_only_chmod?(d) }
```
MikeMcQuaid: I think you can omit changing the variable and messaging name for this; people don't need to know/care about ZSH's special case.

install
dreambottle: Hex wouldn't be readable:
mode = File.stat(d).mode & 0x1ff
Octal is fine:
mode = File.stat(d).mode & 0777
I prefer binary, because it visually represents every bit flag (rwx) for every group (ugo).
Do you see huge difference between octal and binary?

install
MikeMcQuaid: Ok to be a bit more explicit: please change this variable name back, thanks.

dreambottle:
Implemented the requested changes. Hope this will finally be merged :)

MikeMcQuaid:

PR #81, opened 2016-10-24
not fetching unnecessary tags

alessiop86:
As far as I can see tags are never used by the install script


MikeMcQuaid:
They are used by `brew update` so it makes sense to fetch them.

PR #84, opened 2016-10-30
Only invalidate `sudo` if it wasn’t active before `install`.

reitermarkus:
The `sudo` timestamp should only be invalidated if `sudo` was activated first inside the `install` script. If the timestamp was active before the script, it should also be left intact after the script.

### DIFF
--- a/install
+++ b/install
@@ -158,8 +158,14 @@ def chgrp?(d)
   !File.grpowned?(d)
 end
 
-# Invalidate sudo timestamp before exiting
-at_exit { Kernel.system "/usr/bin/sudo", "-k" }
+# Invalidate sudo timestamp before exiting (if it wasn't active before).
+begin
+  $stderr.reopen("/dev/null")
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
What's the use-case? I don't think it warrants the extra complexity for this edge-case, personally.

reitermarkus:
The use case is setup scripts (dotfiles), which install Homebrew before running other setup tasks that require `sudo`.

A simpler solution would be to just remove the `at_exit` block, which would also be in line with virtually every other command line utility out there. I have never seen/used anything that just straight-up invalidates the sudo timestamp.

MikeMcQuaid:
> I have never seen/used anything that just straight-up invalidates the sudo timestamp.

My experience has been the opposite, personally. I'm open to doing this if the diff can be significantly slimmer.

reitermarkus:
Should be slim enough now.

MikeMcQuaid:
Thanks @reitermarkus!

PR #86, opened 2016-11-01
Always use latest CLT version (last in list).

reitermarkus:
Fixes https://github.com/Homebrew/brew/issues/1393

### DIFF
--- a/install
+++ b/install
@@ -262,7 +262,7 @@ if should_install_command_line_tools?
   # This temporary file prompts the 'softwareupdate' utility to list the Command Line Tools
   clt_placeholder = "/tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress"
   sudo "/usr/bin/touch", clt_placeholder
-  clt_label = `softwareupdate -l | grep -B 1 -E "Command Line (Developer|Tools)" | awk -F"*" '/^ +\\*/ {print $2}' | sed 's/^ *//' | head -n1`.chomp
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -262,7 +262,7 @@ if should_install_command_line_tools?
   # This temporary file prompts the 'softwareupdate' utility to list the Command Line Tools
   clt_placeholder = "/tmp/.com.apple.dt.CommandLineTools.installondemand.in-progress"
   sudo "/usr/bin/touch", clt_placeholder
-  clt_label = `softwareupdate -l | grep -B 1 -E "Command Line (Developer|Tools)" | awk -F"*" '/^ +\\*/ {print $2}' | sed 's/^ *//' | head -n1`.chomp
+  clt_label = `/usr/sbin/softwareupdate -l | /usr/bin/sed -E '/^.*\*\ *(Command Line (Developer|Tools).*)\ *$/!d;$!d;s//\1/'`.chomp
```
MikeMcQuaid: No need to hardcode these paths. Can you explain a bit what the regex is matching?

reitermarkus: The regex matches all lines starting with whitespace followed by a star, and extracts everything from `Command Line (Developer|Tools)` to then end of the line, without trailing whitespace.

All lines which are not matched by the regex are deleted (`!d`), and only the last line of the remaining ones is kept (`$!d`) and replaced (`s//\1/`),

MikeMcQuaid: Can you keep the `grep` usage? I think it makes this more readable.

PR #88, opened 2016-11-10
add GCC tip for installing

DarrenCattle:
very simple troubleshooting tip would save time for some users

### DIFF
--- a/install
+++ b/install
@@ -82,6 +82,7 @@ end
 def wait_for_user
   puts
   puts "Press RETURN to continue or any other key to abort"
+  puts "If the program is stopping here, try gcc to accept Xcode command-line tools"
   c = getc
   # we test for \r and \n because some stuff does \r instead
   abort unless (c == 13) || (c == 10)


install
```diff
@@ -82,6 +82,7 @@ end
 def wait_for_user
   puts
   puts "Press RETURN to continue or any other key to abort"
+  puts "If the program is stopping here, try gcc to accept Xcode command-line tools"
```
bfontaine: Shouldn’t that be somewhere else than in `wait_for_user`? This method should "wait for user" and nothing else IMHO.

MikeMcQuaid:
Can you elaborate a bit more on this? We've never seen this behaviour. What do you mean by "stopping here"?

To help us debug this issue can you explain:
- What you were trying to do (and why)
- What happened
- What you expected to happen
- Step-by-step reproduction instructions (with as minimal input data as possible)

MikeMcQuaid:
Passing on this as no answers to the questions.

Issue #87, opened 2016-11-09
xcodebuild -license hint is hidden since commit c2c00a4, installation seems to fail


Lekensteyn:
With commit  c2c00a4, installation does not proceed after pressing Enter. The first time, this is visible, following that no more messages are shown besides the RETURN prompt:
```
hostname:~ User$ /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
==> This script will install:
/usr/local/bin/brew
/usr/local/share/doc/homebrew
/usr/local/share/man/man1/brew.1
/usr/local/share/zsh/site-functions/_brew
/usr/local/etc/bash_completion.d/brew
/usr/local/Homebrew
==> The following new directories will be created:
/usr/local/Cellar
/usr/local/Homebrew
/usr/local/Frameworks
/usr/local/bin
/usr/local/etc
/usr/local/include
/usr/local/lib
/usr/local/opt
/usr/local/sbin
/usr/local/share
/usr/local/share/zsh
/usr/local/share/zsh/site-functions
/usr/local/var

Press RETURN to continue or any other key to abort
==> /usr/bin/sudo /bin/mkdir -p /usr/local/Cellar /usr/local/Homebrew /usr/local/Frameworks /usr/local/bin /usr/local/etc /usr/local/include /usr/local/lib /usr/local/opt /usr/local/sbin /usr/local/share /usr/local/share/zsh /usr/local/share/zsh/site-functions /usr/local/var
Password:
==> /usr/bin/sudo /bin/chmod g+rwx /usr/local/Cellar /usr/local/Homebrew /usr/local/Frameworks /usr/local/bin /usr/local/etc /usr/local/include /usr/local/lib /usr/local/opt /usr/local/sbin /usr/local/share /usr/local/share/zsh /usr/local/share/zsh/site-functions /usr/local/var
==> /usr/bin/sudo /bin/chmod 755 /usr/local/share/zsh /usr/local/share/zsh/site-functions
==> /usr/bin/sudo /usr/sbin/chown User /usr/local/Cellar /usr/local/Homebrew /usr/local/Frameworks /usr/local/bin /usr/local/etc /usr/local/include /usr/local/lib /usr/local/opt /usr/local/sbin /usr/local/share /usr/local/share/zsh /usr/local/share/zsh/site-functions /usr/local/var
[...]

PR #90, opened 2016-11-11
Revert "Only invalidate `sudo` if it wasn’t active before `install`."

MikeMcQuaid:
Reverts Homebrew/install#84

https://github.com/reitermarkus/install/blob/c2c00a4634b425588ee150dbb7978dd12c034c5e/install#L163 means that `stderr` output is never displayed. CC @reitermarkus 

Closes https://github.com/Homebrew/install/issues/87

### DIFF
--- a/install
+++ b/install
@@ -158,14 +158,8 @@ def chgrp?(d)
   !File.grpowned?(d)
 end
 
-# Invalidate sudo timestamp before exiting (if it wasn't active before).
-begin
-  $stderr.reopen("/dev/null")
-  Kernel.system "/usr/bin/sudo", "-n", "-v"
-  at_exit { Kernel.system "/usr/bin/sudo", "-k" } unless $?.success?
-ensure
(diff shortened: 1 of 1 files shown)


reitermarkus:
I don't get why `$stderr.reopen(STDERR)` wouldn't be working.

MikeMcQuaid:
Because `$stderr == STDERR`. I should have spotted that the first time, sorry. `>2` should suffice.

MikeMcQuaid:
> >2 should suffice.

Er, `2>/dev/null`

reitermarkus:
> Because `$stderr == STDERR`.

I thought that `STDERR` would always point to `/dev/fd/2`, since it's a constant, but apparently not.

MikeMcQuaid:
@reitermarkus Me too. I missed this too. Today We Learned 😁

reitermarkus:
> Today We Learned

Indeed we did. 😄

PR #91, opened 2016-11-11
Fix `STDERR` not printing because of #84.

reitermarkus:


install
```diff
@@ -158,8 +158,15 @@ def chgrp?(d)
   !File.grpowned?(d)
 end
 
-# Invalidate sudo timestamp before exiting
-at_exit { Kernel.system "/usr/bin/sudo", "-k" }
+# Invalidate sudo timestamp before exiting (if it wasn't active before).
+begin
+  err = $stderr.dup
+  $stderr.reopen("/dev/null")
+  Kernel.system "/usr/bin/sudo", "-n", "-v"
```
MikeMcQuaid: Let's just use `2>/dev/null` instead of all this.

reitermarkus: Okay.

MikeMcQuaid:
Thanks @reitermarkus!

PR #108, opened 2017-09-12
Allow running installer on 10.5

MikeMcQuaid:
It requires `--insecure` but if we can make it work: we may as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: ruby
 os: osx
-env: OSX=10.11
-osx_image: xcode7.3
+osx_image: xcode8.3
 rvm: system
 
 # Remove Xcode and CommandLineTools included in default macOS image.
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: ruby
 os: osx
-env: OSX=10.11
-osx_image: xcode7.3
-rvm: system
+env: OSX=10.10
+osx_image: xcode6.4
+rvm: 1.8.6-p287
```
reitermarkus: Why can't this stay at `system`, are you “simulating” 10.5 with this? And is the `OSX` variable actually doing anything?

MikeMcQuaid: > are you “simulating” 10.5 with this

Kinda, yes.



> And is the OSX variable actually doing anything?

It's describing the version in the image for quick reference in the Travis UI.

reitermarkus: It's describing the version in the image for quick reference in the Travis UI.

> Ah, I see.

reitermarkus:
https://github.com/Homebrew/install/pull/108/files#diff-19ad89bc3e3c9d7ef68b89523eff1987L94 should be `.map { |p| p.to_i }` for Ruby 1.8.

PR #110, opened 2017-09-22
install: always perform full clone.

MikeMcQuaid:
This means we can always use `git describe` to get the version accurately.

Fixes https://github.com/Homebrew/brew/issues/3187


ilovezfs:
Let's hope no one accidentally checks in a large binary :see_no_evil:

PR #112, opened 2017-11-02
Enables HOMEBREW_PREFIX and HOMEBREW_REPOSITORY to be the same value.

z2kviswanathan:
Hi People at Brew,

This is a FYI PR.

The install documentation states I am allowed to change HOMEBREW_PREFIX. This patch handles cases where both of these values can be set to the same value.

The direct comparison isn't correct, because Pathname.realpath() is apparently the way to go, but it involves importing 'pathname'. This change works for me.

### DIFF
--- a/install
+++ b/install
@@ -284,6 +284,10 @@ Xcode.app or running:
     sudo xcodebuild -license
 EOABORT
 
+[HOMEBREW_REPOSITORY].each do |hbr|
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> The install documentation states I am allowed to change HOMEBREW_PREFIX.

This is not supported by the installer intentionally, sorry.

PR #122, opened 2018-03-24
add "--depth=1" to git fetch

lambdaq:
This might speed things up.


MikeMcQuaid:
It will speed up the initial clone but then the performance drops off and also it breaks some of the automatic version generation code. Sorry but thanks!

PR #124, opened 2018-05-24
Use -k flag for curl on 10.7 and below

vinyldarkscratch:
While trying to solve https://github.com/Homebrew/homebrew-core/issues/28058, I noticed that `curl` isn't getting the `-k` flag on 10.7.  This little update should fix the problem.

### DIFF
--- a/install
+++ b/install
@@ -313,7 +313,7 @@ Dir.chdir HOMEBREW_REPOSITORY do
     # -m to stop tar erroring out if it can't modify the mtime for root owned directories
     # pipefail to cause the exit status from curl to propagate if it fails
     curl_flags = "fsSL"
-    curl_flags += "k" if macos_version < "10.6"
+    curl_flags += "k" if macos_version <= "10.7"
     core_tap = "#{HOMEBREW_PREFIX}/Homebrew/Library/Taps/homebrew/homebrew-core"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Are you sure this fixes the issue? When 10.7 has the CLT installed it will also have `git`.

vinyldarkscratch:
(Rubocop once Travis CI seems to be complaining about things that my PR doesn’t even affect...  😫)

To be completely honest, I have no way of personally testing it to make sure it does the trick.  However, based upon the communication and the reported error (which is a TLS protocol error on this line), and the solution described on brew.sh, this should fix everything.

Now that I think about it though, probably should test that theory out.  😛 I’ll have the user try out the patch, and correct it if need be!

MikeMcQuaid:
> To be completely honest, I have no way of personally testing it to make sure it does the trick. However, based upon the communication and the reported error (which is a TLS protocol error on this line), and the solution described on brew.sh, this should fix everything.

I don't think it will because that `else` path is only if `git` isn't installed and, if that's the case, that's a bigger issue with e.g. the CLT not being installed.

vinyldarkscratch:
Yeah, sounds like it's a bigger issue (this is what I get for writing code at 6:00 in the morning).  >~<

MikeMcQuaid:
Thanks for trying anyway! 🙇

PR #125, opened 2018-07-04
fix command line tools checking on 10.14

irvin:
Install command line tools for Xcode 10 won't create `/usr/include/` folder neither `iconv.h` file, that will make install script stuck on downloading-installing procedure on 

```
==> Installing the Command Line Tools (expect a GUI popup):
==> /usr/bin/sudo /usr/bin/xcode-select --install
xcode-select: error: command line tools are already installed, use "Software Update" to install updates
Failed during: /usr/bin/sudo /usr/bin/xcode-select --install
```

### DIFF
--- a/install
+++ b/install
@@ -110,7 +110,12 @@ end
 def should_install_command_line_tools?
   return false if force_curl?
   return false if macos_version < "10.9"
-  !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git")
+  if macos_version > "10.13"
+    !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git")
+  else
+    !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git") ||
+      !File.exist?("/usr/include/iconv.h")
+  end
 end
 
 def git
--- a/install
+++ b/install
@@ -110,8 +110,7 @@ end
 def should_install_command_line_tools?
   return false if force_curl?
   return false if macos_version < "10.9"
-  !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git") ||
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -110,8 +110,7 @@ end
 def should_install_command_line_tools?
   return false if force_curl?
   return false if macos_version < "10.9"
-  !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git") ||
-    !File.exist?("/usr/include/iconv.h")
+  !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git")
```
wildjcrt:
```
if macos_version > "10.13"
  !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git")
else
  !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git") ||
    !File.exist?("/usr/include/iconv.h")
end

**Comment by MikeMcQuaid** [comment]
We don't support 10.14 yet so still require you to manually install the package that does provide `/usr/include`. I'm open to this PR being modified to install that package for now.

File: install (hunk 1 above)
Comment (MitsunChieh) [reply install]:
```
  return !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git") if macos_version >= "10.14"
  !File.exist?("/Library/Developer/CommandLineTools/usr/bin/git") ||
    !File.exist?("/usr/include/iconv.h")
```

MikeMcQuaid:
Build is failing here I'm afraid.

irvin:
@MikeMcQuaid yes we found a better solution to this issue, thanks.

MikeMcQuaid:
Changed my mind on this, thanks @irvin!

PR #127, opened 2018-07-06
Ask command line tools to install header files to /usr/include

irvin:
On 10.14, CLT installation won't install header files to /usr/include automatically,
which will affect many package like Node.js and most pips rely on zlib, and prevent brew to detect CLT properly. trigger this installation package will fix it.

### DIFF
--- a/install
+++ b/install
@@ -272,7 +272,7 @@ if should_install_command_line_tools?
   
   if macos_version > "10.13"
     # Tell CLT to install header files to /usr/include
-    sudo "sudo installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg -target /"
+    sudo "installer", "-pkg", "/Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg", "-target", "/"
   end
 end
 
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -269,6 +269,11 @@ if should_install_command_line_tools?
   sudo "/usr/sbin/softwareupdate", "-i", clt_label
   sudo "/bin/rm", "-f", clt_placeholder
   sudo "/usr/bin/xcode-select", "--switch", "/Library/Developer/CommandLineTools"
+  
+  if macos_version > "10.13"
+    # Tell CLT to install header files to /usr/include
+    sudo "sudo installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg -target /"
```
MikeMcQuaid: Shouldn't need the additional `sudo`. Can you separate each of the arguments into separate strings? Otherwise looks good, thanks!

MikeMcQuaid:
Let's hold off on this for a bit; it looks like https://github.com/Homebrew/brew/pull/4335 will get merged which will mean this isn't required.

irvin:
cool, we still need to make sure installer can detect CLT properly. 
If Homebrew/brew#4335 merged then we will need #125.

irvin:
We are still waiting for https://github.com/Homebrew/brew/pull/4335 , thanks good bot.

MikeMcQuaid:
It was merged, thanks @irvin!

PR #135, opened 2018-08-18
install: add cask analytics message.

MikeMcQuaid:
Part of https://github.com/Homebrew/brew/pull/4620.


reitermarkus:
Also needs

```
system git, "config", "--local", "--replace-all", "homebrew.caskanalyticsmessage", "true"
```

PR #137, opened 2018-08-23
can force no install xcode CLI with NO_CLI=true in env

revilossor:
I was having trouble with the install script - it kept failing trying to install the Xcode CLI ( which is already on my machine - but in a different location )

This PR will allow this: 

NO_CLI=true /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"

which will force the should_install_command_line_tools func to return false

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,12 @@
 
 More installation information and options at http://docs.brew.sh/Installation.html.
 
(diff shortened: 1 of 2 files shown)


MikeMcQuaid:
Sorry, we're not interested in providing customisation to the install script.

PR #153, opened 2018-11-01
/usr/bin/xcode-select 路径修改

fangyagao:
/Library/Developer/CommandLineTools  -> /Applications/Xcode.app/Contents/Developer


MikeMcQuaid:
This is not correct, these are different things.

PR #154, opened 2018-11-01
Use git:// protocol when run on OS X < 10.9

EricFromCanada:
Use `git://` or gracefully request it be installed instead of failing to download from GitHub because of missing TLS 1.2 on OS X < 10.9. Also updates the URLs where old versions of Xcode and Command Line Tools can be found.

Paired with Homebrew/brew#5238.

### DIFF
--- a/install
+++ b/install
@@ -4,10 +4,13 @@
 # anywhere you like.
 HOMEBREW_PREFIX = "/usr/local".freeze
 HOMEBREW_REPOSITORY = "/usr/local/Homebrew".freeze
+HOMEBREW_CORE_TAP = "/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core".freeze
 HOMEBREW_CACHE = "#{ENV["HOME"]}/Library/Caches/Homebrew".freeze
 HOMEBREW_OLD_CACHE = "/Library/Caches/Homebrew".freeze
 BREW_REPO = "https://github.com/Homebrew/brew".freeze
+BREW_REPO_INSECURE = "git://github.com/Homebrew/brew.git".freeze
 CORE_TAP_REPO = "https://github.com/Homebrew/homebrew-core".freeze
+CORE_TAP_REPO_INSECURE = "git://github.com/Homebrew/homebrew-core.git".freeze
 
 # TODO: bump version when new macOS is released
 MACOS_LATEST_SUPPORTED = "10.14".freeze
@@ -137,14 +140,11 @@ def git
   end
 
   return unless @git
-
-  # Github only supports HTTPS fetches on 1.7.10 or later:
-  # https://help.github.com/articles/https-cloning-errors
-  `#{@git} --version` =~ /git version (\d\.\d+\.\d+)/
-  return if Regexp.last_match(1).nil?
-  return if Version.new(Regexp.last_match(1)) < "1.7.10"
-
   @git
+end
+
+def can_macos_use_tls12?
+  macos_version >= "10.9"
 end
 
 def user_only_chmod?(path)
@@ -170,7 +170,7 @@ def chgrp?(path)
 end
 
 # Invalidate sudo timestamp before exiting (if it wasn't active before).
-Kernel.system "/usr/bin/sudo -n -v 2>/dev/null"
+Kernel.system "/usr/bin/sudo -n -v 2>/dev/null" unless macos_version < "10.7"
 at_exit { Kernel.system "/usr/bin/sudo", "-k" } unless $CHILD_STATUS.success?
 
 # The block form of Dir.chdir fails later if Dir.CWD doesn't exist which I
@@ -178,20 +178,33 @@ at_exit { Kernel.system "/usr/bin/sudo", "-k" } unless $CHILD_STATUS.success?
 Dir.chdir "/usr"
 
 ####################################################################### script
-abort "See Linuxbrew: http://linuxbrew.sh/" if RUBY_PLATFORM.to_s.downcase.include?("linux")
-abort "Mac OS X too old, see: https://github.com/mistydemeo/tigerbrew" if macos_version < "10.5"
+abort "See Linuxbrew: #{Tty.underline}http://linuxbrew.sh/#{Tty.reset}" if RUBY_PLATFORM.to_s.downcase.include?("linux")
+abort "Mac OS X too old, see: #{Tty.underline}https://github.com/mistydemeo/tigerbrew#{Tty.reset}" if macos_version < "10.5"
 abort "Don't run this as root!" if Process.uid.zero?
 abort <<-EOABORT unless `dsmemberutil checkmembership -U "#{ENV["USER"]}" -G admin`.include? "user is a member"
 This script requires the user #{ENV["USER"]} to be an Administrator.
 EOABORT
-# Tests will fail if the prefix exists, but we don't have execution
-# permissions. Abort in this case.
+
+# Tests will fail if the prefix exists, but we don't have execution permissions.
+# Abort in this case.
 abort <<-EOABORT if File.directory?(HOMEBREW_PREFIX) && (!File.executable? HOMEBREW_PREFIX)
 The Homebrew prefix, #{HOMEBREW_PREFIX}, exists but is not searchable. If this is
 not intentional, please restore the default permissions and try running the
 installer again:
     sudo chmod 775 #{HOMEBREW_PREFIX}
 EOABORT
+
+# Require preinstalled git binary if TLS 1.2 not built in.
+if !can_macos_use_tls12? && !git
+  puts "This version of Mac OS X requires Git to perform the installation."
+  if macos_version == "10.5"
+    abort "Install from: #{Tty.underline}http://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/git-osx-installer/git-1.7.5.4-i386-leopard.dmg#{Tty.reset}"
+  elsif macos_version == "10.6"
+    abort "Install from: #{Tty.underline}http://storage.googleapis.com/google-code-archive-downloads/v2/code.google.com/git-osx-installer/git-1.8.4.2-intel-universal-snow-leopard.dmg#{Tty.reset}"
+  else
+    abort "Install Xcode or the Command Line Tools: #{Tty.underline}https://developer.apple.com/download/more/#{Tty.reset}"
+  end
+end
 
 # TODO: bump version when new macOS is released
 if macos_version > MACOS_LATEST_SUPPORTED ||
@@ -212,7 +225,7 @@ if macos_version > MACOS_LATEST_SUPPORTED ||
   puts <<-EOS
 This installation may not succeed.
 After installation, you will encounter build failures and other breakages.
-Please create pull-requests instead of asking for help on Homebrew's
+Please create pull requests instead of asking for help on Homebrew's
 GitHub, Discourse, Twitter or IRC. As you are running this #{what},
 you are responsible for resolving any issues you experience.
 
@@ -361,7 +374,7 @@ Dir.chdir HOMEBREW_REPOSITORY do
   system git, "init", "-q"
 
   # "git remote add" will fail if the remote is defined in the global config
-  system git, "config", "remote.origin.url", BREW_REPO
+  system git, "config", "remote.origin.url", can_macos_use_tls12? ? BREW_REPO : BREW_REPO_INSECURE
   system git, "config", "remote.origin.fetch", "+refs/heads/*:refs/remotes/origin/*"
 
   # ensure we don't munge line endings on checkout
@@ -372,6 +385,12 @@ Dir.chdir HOMEBREW_REPOSITORY do
   system(*args)
 
   system git, "reset", "--hard", "origin/master"
+
+  unless can_macos_use_tls12?
+    system git, "clone", "--depth", "1", CORE_TAP_REPO_INSECURE, HOMEBREW_CORE_TAP
+    system git, "config", "--file", "#{HOMEBREW_CORE_TAP}/.git/config", "core.autocrlf", "false"
+    ENV["HOMEBREW_BOTTLE_DOMAIN"] = "http://homebrew.bintray.com"
+  end
 
   system "ln", "-sf", "#{HOMEBREW_REPOSITORY}/bin/brew", "#{HOMEBREW_PREFIX}/bin/brew"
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Figuring out approach in https://github.com/Homebrew/brew/pull/5238 and then we'll update/merge this 👍

install
```diff
@@ -145,11 +150,13 @@ def git
 
   return unless @git
 
-  # Github only supports HTTPS fetches on 1.7.10 or later:
-  # https://help.github.com/articles/https-cloning-errors
-  `#{@git} --version` =~ /git version (\d\.\d+\.\d+)/
-  return if Regexp.last_match(1).nil?
-  return if Version.new(Regexp.last_match(1)) < "1.7.10"
+  if macos_version >= "10.9"
+    # Github only supports HTTPS fetches on 1.7.10 or later:
```
MikeMcQuaid: `GitHub` :trollface:

install
```diff
@@ -199,6 +206,11 @@ not intentional, please restore the default permissions and try running the
 installer again:
     sudo chmod 775 #{HOMEBREW_PREFIX}
 EOABORT
+# Clone using git:// if TLS 1.2 not built in.
+abort <<-EOABORT if macos_version < "10.9" && !git
+This version of Mac OS X requires git to complete the installation.
+Install Xcode or the Command Line Tools: https://developer.apple.com/download/more/
```
MikeMcQuaid: Might want to tailor that message for 10.5/10.6 where Xcode/CLT doesn't install Git. Check the Homebrew/brew code for a guide to what specifically to recommend (or I can provide additional pointers).

install
```diff
@@ -369,7 +381,7 @@ Dir.chdir HOMEBREW_REPOSITORY do
     system git, "init", "-q"
 
     # "git remote add" will fail if the remote is defined in the global config
-    system git, "config", "remote.origin.url", BREW_REPO
+    system git, "config", "remote.origin.url", (macos_version >= "10.9") ? BREW_REPO : BREW_REPO_INSECURE
```
MikeMcQuaid: Might want to make the `macos_version` bit a function?

install
```diff
@@ -389,15 +417,14 @@ Dir.chdir HOMEBREW_REPOSITORY do
     # pipefail to cause the exit status from curl to propagate if it fails
```
MikeMcQuaid: Could consider eliminating the `curl`-only path entirely if it's requiring a modern HTTPS client anyway?

install
EricFromCanada: I believe it's only Xcode 3 that doesn't install Git; Xcode 4.2 does when installed on Snow Leopard.

install
EricFromCanada: If it was only there for use by older OSes, then yeah I guess.

install
MikeMcQuaid: It's been an extremely long time but I believe we only ever supported/recommended Xcode 3 on Snow Leopard. Something about the app store or being paid or something.

install
MikeMcQuaid: @EricFromCanada I think it's there originally for OSs where they didn't have Git preinstalled but HTTPS from GitHub still worked. Now that Git is more backwards compatible than Curl for accessing GitHub I think added complexity isn't worth it.

install
EricFromCanada: It was with v4 that they switched to the self-contained, store-delivered `/Applications/Xcode.app` arrangement that caused some consternation at the time.

MikeMcQuaid: Xcode from the app store? How outrageous 😆! Regardless because of 10.5 I reckon it's worth handling the Git case. In an ideal world we recommend CLT when that's available otherwise Xcode otherwise https://git-scm.com/download/mac

install
```diff
@@ -177,7 +181,7 @@ def chgrp?(path)
 end
 
 # Invalidate sudo timestamp before exiting (if it wasn't active before).
-Kernel.system "/usr/bin/sudo -n -v 2>/dev/null"
+Kernel.system "/usr/bin/sudo -n -v 2>/dev/null" unless macos_version < "10.7"
```
EricFromCanada: This isn't supported on 10.5 and prints an error on 10.6.

install
```diff
@@ -145,11 +147,13 @@ def git
...
-  # https://help.github.com/articles/https-cloning-errors
-  `#{@git} --version` =~ /git version (\d\.\d+\.\d+)/
-  return if Regexp.last_match(1).nil?
-  return if Version.new(Regexp.last_match(1)) < "1.7.10"
+  if can_use_https?
+    # GitHub only supports HTTPS fetches on 1.7.10 or later:
+    # https://help.github.com/articles/https-cloning-errors
+    `#{@git} --version` =~ /git version (\d\.\d+\.\d+)/
+    return if Regexp.last_match(1).nil?
+    return if Version.new(Regexp.last_match(1)) < "1.7.10"
```
MikeMcQuaid: Seems a bit weird to return `nil` here if there's a valid Git but we're not going to do HTTPS cloning. Could perhaps assume given the Git versions that if `can_use_https` is fine then the Git version is too or add another function `can_git_use_https?`?

install
```diff
@@ -199,6 +203,17 @@ not intentional, please restore the default permissions and try running the
 installer again:
     sudo chmod 775 #{HOMEBREW_PREFIX}
 EOABORT
+# Clone using git:// if TLS 1.2 not built in.
+unless can_use_https? || git
```
MikeMcQuaid: Can this if `if ! .. && !`; a bit easier to follow.

install
```diff
@@ -4,15 +4,20 @@
...
+BREW_REPO_INSECURE = "git://github.com/Homebrew/brew.git".freeze
 CORE_TAP_REPO = "https://github.com/Homebrew/homebrew-core".freeze
+CORE_TAP_REPO_INSECURE = "git://github.com/Homebrew/homebrew-core.git".freeze
 
 # TODO: bump version when new macOS is released
 MACOS_LATEST_SUPPORTED = "10.14".freeze
 # TODO: bump version when new macOS is released
 MACOS_OLDEST_SUPPORTED = "10.12".freeze
+# TODO: keep in sync with https://github.com/Homebrew/homebrew-portable-ruby/releases/latest
+PORTABLE_RUBY_INSECURE = "http://homebrew.bintray.com/bottles-portable-ruby/portable-ruby-2.3.7.leopard_64.bottle.tar.gz".freeze
```
MikeMcQuaid: This maybe warrants a Homebrew/brew change instead (where we fall back to HTTP on old versions)? Going to be a bit of a nightmare keeping this up-to-date.

install
```diff
@@ -364,45 +379,42 @@ EOABORT
...
-    system "ln", "-sf", "#{HOMEBREW_REPOSITORY}/bin/brew", "#{HOMEBREW_PREFIX}/bin/brew"
-
-    system "/bin/mkdir", "-p", core_tap
-    Dir.chdir core_tap do
-      system "/bin/bash -o pipefail -c '" \
-        "/usr/bin/curl -#{curl_flags} #{CORE_TAP_REPO}/tarball/master | " \
-        "/usr/bin/tar xz -m --strip 1'"
+      "cd #{HOMEBREW_CACHE} && /usr/bin/curl -OfsSL #{PORTABLE_RUBY_INSECURE}'"
+    system "/bin/mkdir", "-p", HOMEBREW_CORE_TAP
+    Dir.chdir HOMEBREW_CORE_TAP do
```
MikeMcQuaid: Could do a `git clone` here instead?

install
```diff
@@ -364,45 +379,42 @@ EOABORT
...
 
-    system "#{HOMEBREW_PREFIX}/bin/brew", "update", "--force"
-  else
-    # -m to stop tar erroring out if it can't modify the mtime for root owned directories
-    # pipefail to cause the exit status from curl to propagate if it fails
-    curl_flags = "fsSL"
-    curl_flags += "k" if macos_version < "10.6"
-    core_tap = "#{HOMEBREW_PREFIX}/Homebrew/Library/Taps/homebrew/homebrew-core"
+  unless can_use_https?
+    system git, "config", "--file", "#{HOMEBREW_REPOSITORY}/.git/config", "remote.origin.url", BREW_REPO
```
MikeMcQuaid: Does `brew doctor` complain if this isn't changed back?

install
```diff
@@ -364,45 +379,42 @@ EOABORT
...
+      "cd #{HOMEBREW_CACHE} && /usr/bin/curl -OfsSL #{PORTABLE_RUBY_INSECURE}'"
+    system "/bin/mkdir", "-p", HOMEBREW_CORE_TAP
+    Dir.chdir HOMEBREW_CORE_TAP do
+      system git, "init", "-q"
+      system git, "config", "remote.origin.url", CORE_TAP_REPO_INSECURE
+      system git, "config", "remote.origin.fetch", "+refs/heads/*:refs/remotes/origin/*"
+      system git, "config", "core.autocrlf", "false"
+      system git, "fetch", "--force", "--depth=1", "origin", "refs/heads/master:refs/remotes/origin/master"
+      system git, "reset", "--hard", "origin/master"
+      system git, "config", "--file", "#{HOMEBREW_CORE_TAP}/.git/config", "remote.origin.url", CORE_TAP_REPO
```
MikeMcQuaid: Does `brew doctor` complain if this isn't changed back?

install
EricFromCanada: Replaced with setting `HOMEBREW_BOTTLE_DOMAIN` instead.

install
```diff
@@ -199,6 +192,17 @@ not intentional, please restore the default permissions and try running the
 installer again:
     sudo chmod 775 #{HOMEBREW_PREFIX}
 EOABORT
+# Clone using git:// if TLS 1.2 not built in.
```
MikeMcQuaid: Pendatry: empty line above this one?

MikeMcQuaid: Also: is there code missing heee? It doesn’t seem to match what the code immediately below does.

install
```diff
@@ -364,45 +368,31 @@ EOABORT
...
-    system "/bin/mkdir", "-p", core_tap
-    Dir.chdir core_tap do
-      system "/bin/bash -o pipefail -c '" \
-        "/usr/bin/curl -#{curl_flags} #{CORE_TAP_REPO}/tarball/master | " \
-        "/usr/bin/tar xz -m --strip 1'"
-    end
+  unless can_macos_use_tls12?
+    system git, "clone", "--depth", "1", CORE_TAP_REPO_INSECURE, HOMEBREW_CORE_TAP
+    system git, "config", "--file", "#{HOMEBREW_CORE_TAP}/.git/config", "core.autocrlf", "false"
+    ENV["HOMEBREW_BOTTLE_DOMAIN"] = "http://homebrew.bintray.com"
```
MikeMcQuaid: Very clever 👏🏻

install
EricFromCanada: Moved to `brew update` in Homebrew/brew#5238.

install
EricFromCanada: I've updated the comment.

EricFromCanada:
@MikeMcQuaid I tested this on:
- plain 10.5 – 10.8: quits with message saying how to install git
- 10.5 + Xcode + prepackaged git: installs, although curl's auto-install fails at openssl: `makedepend: error:  "omit-frame-pointer" is not present`
- 10.6 + Xcode + prepackaged git: installs, but openssl fails as described in Homebrew/brew#2766
- 10.7 + CLT: installs, auto-installs curl & git, `brew doctor` is clean
- 10.8 + CLT: installs, auto-installs curl & git, `brew doctor` is clean
- plain 10.9: auto-installs CLT, curl & git, `brew doctor` is clean

Note also the current state of Homebrew/brew#5238, which adds a block to `brew update` for checking & updating the remotes' URLs if necessary, which I also tested.

Speaking of Homebrew/brew#2766… it's still a mystery as to why Tigerbrew has no problem compiling OpenSSL 1.0.2j. Even when I change Homebrew's formula and compiler to match to a point where the logs differ only by timestamps, it still fails with the "SHA1 mismatch" error only for Homebrew.

MikeMcQuaid:
> * 10.6 + Xcode + prepackaged git: installs, but openssl fails as described in [Homebrew/brew#2766](https://github.com/Homebrew/brew/pull/2766)
> Speaking of [Homebrew/brew#2766](https://github.com/Homebrew/brew/pull/2766)…

This doesn't seem to be the relevant issue?

> * 10.5 + Xcode + prepackaged git: installs, although curl's auto-install fails at openssl: `makedepend: error:  "omit-frame-pointer" is not present`

Mind opening issues for both the above and CCing me on them? You may need to note I specifically requested the issues opened and I'll look at fixing them so other maintainers don't 😭 at the 10.5 😆.

---

Otherwise: do you consider this ready to merge? Great work here again 👏.

EricFromCanada:
Oops. Wrong repo… it's Homebrew/homebrew-core#2766. So far it looks like compiling with clang on 10.6 causes the SHA1 test failures, but it's fine if built by hand with llvm-gcc, which isn't in Homebrew's list of compilers (probably because it was removed as of Xcode 5). I'll post a PR once I get it working. After that I can look at the 10.5 issue.

But I'd consider these changes good to go.

MikeMcQuaid:
@EricFromCanada Does one of the versioned GCCs work for either package? If so, we could add the relevant `fails_with` so it's used, perhaps.

EricFromCanada:
@MikeMcQuaid I've found that OpenSSL installs on 10.6 after installing `apple-gcc42` which becomes the default compiler on 10.6 due to `HOMEBREW_CC` being set to `gcc-4.2` on that release (though that formula has unresolved issues of its own: Homebrew/brew#5331), and adding this to the `openssl` formula's install block: `inreplace "Makefile", "makedepend", "gcc-4.2" if MacOS.version <= :snow_leopard`. After that `curl` does install, but there's still some certificate issues to iron out due `/usr/local/etc/openssl/cert.pem` being blank, causing HTTPS downloads to still fail.

MikeMcQuaid:
Thanks for digging in @EricFromCanada! I was thinking about any of the `gcc@*` formula. It may be it's worth just 🔥ing the apple-gcc42 formula if it doesn't work and no-one noticed.

EricFromCanada:
All the other gcc formulae would need work to compile, but even then we have a chicken-and-egg issue where without a working `openssl` and `curl`, it can't download those sources. But that shouldn't be necessary as I've found fixes for both using the built-in tools.

MikeMcQuaid:
@EricFromCanada Good point. Nice work!

PR #162, opened 2019-01-06
install: reject Snow Leopard and Leopard installations.

MikeMcQuaid:
These will be deprecated in future so let's stop getting any new installs now.

See some related deprecations in https://github.com/Homebrew/brew/pull/5477.

### DIFF
--- a/install
+++ b/install
@@ -170,7 +170,7 @@ def chgrp?(path)
 end
 
 # Invalidate sudo timestamp before exiting (if it wasn't active before).
-Kernel.system "/usr/bin/sudo -n -v 2>/dev/null" unless macos_version < "10.7"
+Kernel.system "/usr/bin/sudo -n -v 2>/dev/null"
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -170,7 +170,6 @@ def chgrp?(path)
 end
 
 # Invalidate sudo timestamp before exiting (if it wasn't active before).
-Kernel.system "/usr/bin/sudo -n -v 2>/dev/null" unless macos_version < "10.7"
```
reitermarkus: This is still needed, only remove the `unless`.

RandomDSdevel:
@MikeMcQuaid:  

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;How much, if any, does @mistydemeo know about this?  She may want to add support for these OS X versions to [Tigerbrew](https://github.com/mistydemeo/tigerbrew) (though that might necessitate a rename…)

RandomDSdevel:
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Right, OK; this doesn't actually affect me, but I was still curious anyway.

PR #163, opened 2019-01-08
Support CLT installs outside of /Library

indirect:
On some systems, XCode.app contains a full set of command-line tools
within itself, without command line tools being installed into /Library.

This change skips installing the command line tools if `xcrun` is
already able to return the path to a usable `git`.

### DIFF
--- a/install
+++ b/install
@@ -121,6 +121,9 @@ end
 def should_install_command_line_tools?
   return false if macos_version < "10.9"
 
+  exe = `xcrun -find git 2>/dev/null`.chomp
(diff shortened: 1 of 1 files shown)


indirect:
Huh. That's news to me. The documentation I was able to find [explicitly says](https://github.com/Homebrew/brew/blob/master/docs/Installation.md#3) that an XCode install _without_ the command line tools is "supported":

> You can install Xcode, the CLT, or both; Homebrew supports all three configurations.

indirect:
Thanks for the additional info! I'll stick with manual installations. You might want to update the docs to note that XCode alone is no longer supported.

MikeMcQuaid:
It is supported. That doesn't mean that our installer needs to not install the CLT, though.

PR #167, opened 2019-01-21
Fix  Bug: missing HOMEBREW_CORE_TAP

ahuigo:
Fix the bug that missing HOMEBREW_CORE_TAP with newly installation.

https://github.com/Homebrew/brew/issues/5570#event-2085902879

### DIFF
--- a/install
+++ b/install
@@ -379,10 +379,12 @@ Dir.chdir HOMEBREW_REPOSITORY do
 
   system git, "reset", "--hard", "origin/master"
 
-  unless can_macos_use_tls12?
+  if can_macos_use_tls12?
     system git, "clone", "--depth", "1", CORE_TAP_REPO_INSECURE, HOMEBREW_CORE_TAP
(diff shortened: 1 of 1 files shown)


install
```diff
@@ -379,10 +379,12 @@ Dir.chdir HOMEBREW_REPOSITORY do
 
   system git, "reset", "--hard", "origin/master"
 
-  unless can_macos_use_tls12?
+  if can_macos_use_tls12?
```
MikeMcQuaid: This is not a correct change; this is making modern versions of macOS use `CORE_TAP_REPO_INSECURE`

ahuigo: Is this change ok?
```suggestion
  if can_macos_use_tls12?
    system git, "clone", "--depth", "1", CORE_TAP_REPO, HOMEBREW_CORE_TAP
  else
    system git, "clone", "--depth", "1", CORE_TAP_REPO_INSECURE, HOMEBREW_CORE_TAP
  end

```

MikeMcQuaid: It's not necessary as it's done by Homebrew itself.

PR #169, opened 2019-01-23
install: disable pre-Mavericks.

MikeMcQuaid:
This will be no longer supported in Homebrew 2.0.0.

This should not be merged until 2.0.0 will be released imminently.


MikeMcQuaid:
Actually, let's ship this now so they don't get a nasty surprise in a few weeks.

PR #172, opened 2019-01-24
install: HOMEBREW_CACHE is not an array

dbramwell:
Fix broken code treating HOMEBREW_CACHE as an array


dbramwell:
Hi, having issues installing brew today. This fixes it.

PR #174, opened 2019-01-25
Add install.sh

sjackman:
On macOS run…
```sh
/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
```
and on Linux run…
```sh
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install.sh)"
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,0 +1,7 @@
+#!/bin/sh
+set -eu
+case "$(uname -s)" in
+  Darwin) exec /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)" "$@" ;;
+  Linux)  exec /bin/sh -c "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install.sh)" "$@" ;;
+  *) echo "Error: Your operating system is not supported: $(uname -s)" >&2; exit ;;
+esac


sjackman:
@MikeMcQuaid Take 2 for review. Thoughts?

MikeMcQuaid:
@sjackman That's a clever approach, nice 👍 . How confident are you in it being rock-solid (e.g. will fail in a non-confusing way regardless of Ruby installation/current shell) in comparison to the existing version?

sjackman:
PR https://github.com/Homebrew/install/pull/176 updates the installation instructions for Linux.

sjackman:
I'd like for the default installation instructions work on both macOS and Linux. Currently if a Linux user goes to https://brew.sh and runs the installation command there, they may likely see the error…
```console
$ /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
bash: /usr/bin/ruby: No such file or directory
```
and give up there, if they don't have Ruby installed on their system, assuming its a dependency.

Two options I see are…
1. Change the default installation instructions to use `/bin/sh` as this PR had previously
2. Add a separate installation command to https://brew.sh for Linux immediately below the macOS version

Thoughts, Mike?

MikeMcQuaid:
@sjackman Much like the discussions on the title: I don't think the macOS and Linux install methods warrant equal footing on the website. We do link specifically to Linux instructions and that seems sufficient to me if there's no way to do this without degrading the macOS script. Remember the people who will be hit by this are those who both try to run the macOS installer _and_ don't read the Linux instructions on this site. This would have been previously hit by users on Mountain Lion and below who had the same instructions/failures and none of them complained or reported this issue.

sjackman:
Fair enough, but it requires two clicks to get to the Linuxbrew installation instructions from https://brew.sh:

1. There are more installation options [here](https://docs.brew.sh/Installation) (required for Linux and Windows 10 Subsystem for Linux).
2. Check out the [Linuxbrew installation documentation](https://docs.brew.sh/Linuxbrew).

I'd like to get that down to one click. I suggest replacing (1) with:

[Install Homebrew on Linux](https://docs.brew.sh/Linuxbrew) and Windows 10 Subsystem for Linux. There are more installation options [here](https://docs.brew.sh/Installation).

MikeMcQuaid:
> I'd like to get that down to one click. I suggest replacing (1) with:
> 
> [Install Homebrew on Linux](https://docs.brew.sh/Linuxbrew) and Windows 10 Subsystem for Linux. There are more installation options [here](https://docs.brew.sh/Installation).

👍 but also link `and Windows 10 Subsystem for Linux` and maybe add `more installation options for macOS or Linux here`?

MikeMcQuaid:
@sjackman https://github.com/Homebrew/install/pull/177 should make the Linux failure case a bit better. Additionally, probably want to update the instructions on the README, too: https://github.com/Homebrew/install/blob/master/README.md

sjackman:
I've opened PR https://github.com/Homebrew/brew.sh/pull/315 to add a link to the installation instructions on Linux.

sjackman:
I've reverted this PR back to its original form, a `install.sh` trampoline script:
```sh
#!/bin/sh
set -eu
case "$(uname -s)" in
  Darwin) exec /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)" "$@" ;;
  Linux)  exec /bin/sh -c "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install.sh)" "$@" ;;
  *) echo "Error: Your operating system is not supported: $(uname -s)" >&2; exit ;;
esac
```

MikeMcQuaid:
> I'd rather we display either a prominent link to Linux install instructions

We've got this now in https://github.com/Homebrew/brew.sh/pull/315

> The problem with this is that for those folk who want to manually download and inspect the script it effectively adds a "redirect" through this script.

This is still my concern here (particularly given how much shit we already get for our install script method).

sjackman:
My preference is the `install.sh` trampoline. I'm fine closing this PR for now, and waiting for any actual bug reports from any confused Linux users without `/usr/bin/ruby` installed.

PR #176, opened 2019-01-28
Update the installation instructions for Linux

sjackman:


install
```diff
@@ -155,9 +155,11 @@ Dir.chdir "/usr"
...
 ####################################################################### script
 if RUBY_PLATFORM.to_s.downcase.include?("linux")
-  <<-EOABORT
-  You're on Linux! Use the Linuxbrew installer:
-    #{Tty.underline}http://linuxbrew.sh/#{Tty.reset}"
+  abort <<-EOABORT
+  To install Linuxbrew, see the installation instructions at
+    #{Tty.underline}https://docs.brew.sh/Linuxbrew#{Tty.reset}
+  or paste at a terminal prompt
+    sh -c "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install.sh)"
```
MikeMcQuaid: I think it's probably just worth recommending the latter.

sjackman: Done

MikeMcQuaid:
Nice work @sjackman!

sjackman:
Thanks for merging, Mike!

PR #177, opened 2019-01-28
install: use /usr/bin/ruby.

MikeMcQuaid:
This should make the Linux failure case a little nicer and SIP protects us from El Capitan onwards.


sjackman:
This shebang line was never actually used by anyone (except perhaps developers), since the installation instructions at https://brew.sh are `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`, the `/usr/bin/ruby` interpreter is used, not the one specified by the shebang line. I'm in favour of this PR all the same. Thanks, Mike.

PR #178, opened 2019-01-28
README.md: Link to Linux installation instructions

sjackman:


README.md
```diff
@@ -8,6 +8,7 @@
 /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
 ```
 
+Install Homebrew on [Linux and Windows 10 Subsystem for Linux](https://docs.brew.sh/Linuxbrew).
```
MikeMcQuaid: Maybe make this an H3 for added prominence?

```markdown
### Linux and Windows 10 Subsystem for Linux
```

and then the one-liner instructions?

sjackman: How do you feel about doing the same for https://brew.sh?

MikeMcQuaid: I've rather leave it as-is given the two-column thing it's got going. I did just think, though, that `Paste that in a Terminal prompt.` could be replaced with `Paste that in a macOS Terminal prompt.` for added clarity

sjackman: Updated this PR with H3.

sjackman: > could be replaced with Paste that in a macOS Terminal prompt.

Good idea. Done.

sjackman:
Thanks for the review and merging, Mike.

PR #180, opened 2019-02-17
git fetch with --depth=1

iron9light:


MikeMcQuaid:
We intentionally do a full fetch to make contribution easier and avoid the repository becoming slowly unshallowed (which is less performant). Thanks anyway!

PR #213, opened 2019-06-28
Gracefully handle missing USER env var

joehorsnell:
I actually ran into [this issue][1] on `Linuxbrew/install` when testing inside
Docker and noticed that [#201][2] had been merged in `Homebrew/install`.

This is a copy of the [PR][3] I did to `Linuxbrew/install` and I guess is a bit
more forgiving?

[1]: https://github.com/Linuxbrew/install/issues/48
[2]: https://github.com/Homebrew/install/pull/201
[3]: https://github.com/Linuxbrew/install/pull/62

### DIFF
--- a/install
+++ b/install
@@ -145,7 +145,7 @@ def chgrp?(path)
   !File.grpowned?(path)
 end
 
-# USER is always set so provide a fall back for the installer and subprocesses.
+# USER isn't always set so provide a fall back for the installer and subprocesses.
 ENV["USER"] ||= `id -un`.chomp
 
 # Invalidate sudo timestamp before exiting (if it wasn't active before).
--- a/install
+++ b/install
@@ -145,9 +145,10 @@ def chgrp?(path)
   !File.grpowned?(path)
 end
 
-def user
-  @user ||= ENV["USER"] || `id -un`.chomp
-end
+# The USER env var is not guaranteed to be available, so fall back to id if not
+# provided. We destructively set it in the environment to make it available to
+# any sub-processes called.
+ENV["USER"] = `id -un`.chomp unless ENV["USER"]
 
 # Invalidate sudo timestamp before exiting (if it wasn't active before).
 Kernel.system "/usr/bin/sudo -n -v 2>/dev/null"
@@ -172,8 +173,8 @@ elsif macos_version < "10.9"
   abort "Your OS X version is too old"
 elsif Process.uid.zero?
   abort "Don't run this as root!"
-elsif !`dsmemberutil checkmembership -U "#{user}" -G admin`.include?("user is a member")
-  abort "This script requires the user #{user} to be an Administrator."
+elsif !`dsmemberutil checkmembership -U "#{ENV["USER"]}" -G admin`.include?("user is a member")
+  abort "This script requires the user #{ENV["USER"]} to be an Administrator."
 elsif File.directory?(HOMEBREW_PREFIX) && (!File.executable? HOMEBREW_PREFIX)
   abort <<-EOABORT
 The Homebrew prefix, #{HOMEBREW_PREFIX}, exists but is not searchable. If this is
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Homebrew formulae assume that `USER` will be set correctly so I don't think this is the right fix. Some probably have a similar check in Homebrew/brew for the `USER` variable.

joehorsnell:
>Homebrew formulae assume that USER will be set correctly so I don't think this is the right fix. Some probably have a similar check in Homebrew/brew for the USER variable.

That's a good point @MikeMcQuaid, I hadn't considered that. How about [this](https://github.com/Linuxbrew/install/pull/62/files) approach of setting the `USER` env var if not present and adding a similar thing to Homebrew itself, so that `USER` is always available to formulae when called?

On another note, have you given any thought to merging `Linuxbrew/install` back into `Homebrew/install`? It seems that the differences are fewer than the similarities and `Homebrew/install` could fairly easily be made to `DoTheRightThing` on both platforms?

Any interest in a PR for that (also ping @sjackman)?

sjackman:
> How about [this](https://github.com/Linuxbrew/install/pull/62/files) approach of setting the `USER` env var if not present and adding a similar thing to Homebrew itself, so that `USER` is always available to formulae when called?

I'm into this. I'd suggest doing that here: https://github.com/Homebrew/brew/blob/686760d100d1a518863805247a606cc1cef12a87/Library/Homebrew/brew.sh#L6

sjackman:
> On another note, have you given any thought to merging Linuxbrew/install back into Homebrew/install? It seems that the differences are fewer than the similarities and Homebrew/install could fairly easily be made to DoTheRightThing on both platforms?

Linuxbrew/install has always supported both macOS and Linux. I would like to merge the two repositories. It'd be a big step toward winding down the Linuxbrew GitHub organization.

MikeMcQuaid:
> That's a good point @MikeMcQuaid, I hadn't considered that. How about [this](https://github.com/Linuxbrew/install/pull/62/files) approach of setting the `USER` env var if not present and adding a similar thing to Homebrew itself, so that `USER` is always available to formulae when called?

This seems like a better fix.

install
```diff
@@ -145,6 +145,10 @@ def chgrp?(path)
   !File.grpowned?(path)
 end
 
+def user
+  @user ||= ENV["USER"] || `id -un`.chomp
```
MikeMcQuaid: How about not making this a function but instead doing
```ruby
ENV["USER"] || `id -un`.chomp
```
so subprocesses can benefit?

MikeMcQuaid:
> That's a good point @MikeMcQuaid, I hadn't considered that. How about [this](https://github.com/Linuxbrew/install/pull/62/files) approach of setting the `USER` env var if not present and adding a similar thing to Homebrew itself, so that `USER` is always available to formulae when called?

This seems like a better fix.

install
```diff
@@ -145,6 +145,11 @@ def chgrp?(path)
   !File.grpowned?(path)
 end
 
+# The USER env var is not guaranteed to be available, so fall back to id if not
+# provided. We destructively set it in the environment to make it available to
+# any sub-processes called.
+ENV["USER"] = `id -un`.chomp unless ENV["USER"]
```
MikeMcQuaid: Use `||=` instead of `=` and then you can skip the `unless ...`

joehorsnell: Squash now @MikeMcQuaid? Or do you keep fixup/wip commits in your history?

install
```diff
@@ -145,6 +145,11 @@ def chgrp?(path)
   !File.grpowned?(path)
 end
 
+# The USER env var is not guaranteed to be available, so fall back to id if not
+# provided. We destructively set it in the environment to make it available to
```
sjackman: I feel that `destructively` is a bit misleading, because it's only set if it's not already set. I'd be inclined to remove this word.

install
sjackman: I prefer to squash.

MikeMcQuaid:
Thanks again @joehorsnell!

install
joehorsnell: Yep, I agree. I went to change this and saw that @MikeMcQuaid had fixed it - thanks 👍.

PR #215, opened 2019-07-17
install: add `.cleaned` file to cache

EricFromCanada:
This prevents cleanup from being run immediately after the first action on a fresh installation.

Also updated the macOS version warning message to match brew's in `diagnostic.rb`.


MikeMcQuaid:
Hmm, this feels a bit weird to have to do manually. What do you think about adjusting the `brew cleanup` code to e.g. check the date the directory was created or something similar?

EricFromCanada:
Sure, I should be able to pull that off, soon-ish.

EricFromCanada:
@MikeMcQuaid Original fix is re-filed as Homebrew/brew#6320.

MikeMcQuaid:
@EricFromCanada sorry to yo-yo I think the previous `.cleaned` solution here may be the best after all 😭.

EricFromCanada:
@MikeMcQuaid Alright, that should do it.

MikeMcQuaid:
Great work again @EricFromCanada!

PR #216, opened 2019-07-24
Update install: use shallow cloning

zaoqi:
it is faster than cloning the repository's whole Git history


MikeMcQuaid:
This is faster on the initial clone but ends up slower and slower over time.

zaoqi:
@MikeMcQuaid Why?

MikeMcQuaid:
@zaoqi That's just how Git works. Shallow clones are primarily useful for one-off clones e.g. CI environments.

PR #234, opened 2019-10-09
Use github.com URLs

sholladay:
This is just a proposal to use **github.com** URLs instead of **githubusercontent.com**.

1. It's shorter, makes the full line easier to read.
2. It looks more legit, less like some sort of phishing attack.
3. It's what the GitHub UI uses, so it's probably a slightly more stable interface.

I haven't changed them everywhere yet, just wanted to spark a conversation. If you like it, I will do it everywhere to keep things consistent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 ## Install Homebrew
 
 ```bash
-/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
If they work without having to modify what's passed to `curl` (as they seem to from your PR and my local testing): I agree this is a worthwhile change to make globally (i.e. in this repo, Homebrew/brew, Homebrew/install).

PR #242, opened 2019-11-16
Remove `--local` from `git config` commands as it's the default

issyl0:
- The `--local` option doesn't exist in early versions of git (~1.7),
  but its behaviour is the default (saving in the current repo's .git
  directory).
- As it's the default across versions, we don't need to specify it
  everywhere. The `global` and `system` flags are still legitimate.
- Reported in https://github.com/Linuxbrew/install/issues/78, and we'll
  merge these changes into that fork.

### DIFF
--- a/install
+++ b/install
@@ -371,8 +371,8 @@ puts <<-EOS
 EOS
 
 Dir.chdir HOMEBREW_REPOSITORY do
-  system "git", "config", "--local", "--replace-all", "homebrew.analyticsmessage", "true"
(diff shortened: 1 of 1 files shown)


issyl0:
Thanks for the approval. I don’t have permissions on this repo so you’ll have to merge this for me, please.

sjackman:
I pinged Mike for a review.

MikeMcQuaid:
See https://github.com/Homebrew/brew/pull/6747#issuecomment-555980437

In this case given the functionality is a no-op: I'm happy to remove it.

PR #247, opened 2019-12-03
Adopt LICENSE.txt from Homebrew/brew

### LINKED ISSUE #246: Clarify license wording/formatting
This repo has a `LICENSE.txt` file that GitHub cannot identify as a known license type, though it appears to be nearly identical to the BSD 2-Clause License, which is used on many other Homebrew repos (Homebrew/brew, Homebrew/homebrew-core, Homebrew/homebrew-cask). Having GitHub clearly identify the license type would ease the path toward contributing to this repo, in my opinion.

The important difference seems to be the use of "author" instead of "copyright holder and contributor". If this license was intended to be BSD 2-Clause, would it make sense to simply copy the `LICENSE.txt` file from Homebrew/brew to this repo? If the maintainers agree, I'm happy to make such a pull request.

mikesep:
The existing license file was close to the 2-clause BSD license, but the
text didn't match exactly. This commit overwrites this repo's
LICENSE.txt file with the one from Homebrew/brew.

Closes #246.

### DIFF
--- a/LICENSE.txt
+++ b/LICENSE.txt
@@ -1,22 +1,25 @@
-Copyright 2009-2016 Homebrew contributors.
+BSD 2-Clause License
+
+Copyright (c) 2009-present, Homebrew contributors
+All rights reserved.
 
 Redistribution and use in source and binary forms, with or without
-modification, are permitted provided that the following conditions
-are met:
+modification, are permitted provided that the following conditions are met:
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @mikesep!

PR #254, opened 2020-01-09
Allow custom BREW_REPO using env var

elquimista:
This allows installation of Homebrew within organizations that have firewalls or similar situations that can't reach the official Github repo.

```sh
$ BREW_REPO=http://gitlab.local:4000/Homebrew/brew.git \
  HOMEBREW_TAP_DEFAULT_REMOTE_BASE=http://gitlab.local:4000 \
  ruby -e "$(curl -fsSL http://gitlab.local:4000/Homebrew/install/raw/master/install)"
```

### DIFF
--- a/install
+++ b/install
@@ -5,7 +5,7 @@
 HOMEBREW_PREFIX = "/usr/local".freeze
 HOMEBREW_REPOSITORY = "/usr/local/Homebrew".freeze
 HOMEBREW_CACHE = "#{ENV["HOME"]}/Library/Caches/Homebrew".freeze
-BREW_REPO = "https://github.com/Homebrew/brew".freeze
+BREW_REPO = ENV.fetch("BREW_REPO", "https://github.com/Homebrew/brew").freeze
 
 # TODO: bump version when new macOS is released
 MACOS_LATEST_SUPPORTED = "10.15".freeze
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
> This allows installation of Homebrew within organizations that have firewalls or similar situations that can't reach the official Github repo.

They can use the other installation methods: https://docs.brew.sh/Installation#alternative-installs

elquimista:
@MikeMcQuaid I understand we can use the other installation methods you suggested and that looks something like this on macOS:
```sh
$ sudo mkdir /usr/local/Homebrew
$ sudo chown $(id -u):admin /usr/local/Homebrew
$ git clone http://gitlab.local:4000/Homebrew/brew.git /usr/local/Homebrew
$ ln -s /usr/local/Homebrew/bin/brew /usr/local/bin/brew
$ echo 'eval "$(brew shellenv)"' >> ~/.zshrc
```
This works but I guess this is what this `install` repo does already so why can't we just use regular one-liner (with different repo url)? Plus this PR doesn't introduce any new variable at all - `BREW_REPO` was already there.

Also setting `HOMEBREW_CORE_GIT_REMOTE` didn't help - let's continue discussion about it on https://github.com/Homebrew/brew/pull/6918

MikeMcQuaid:
> This works but I guess this is what this `install` repo does already so why can't we just use regular one-liner (with different repo url)?

Because the maintainers will have to support this behaviour indefinitely.

elquimista:
What kind of potential burden do you foresee with this behavior in terms of maintenance?

MikeMcQuaid:
@elquimista Sorry, it's a no for now. I don't want to get into a back-and-forth of justifying it.

PR #256, opened 2020-01-23
Bash installer

### LINKED ISSUE #217: Port installer to Bash
The installer should be ported (with identical logic on macOS to begin with) to Bash for:

1. future versions of macOS that don't include Ruby
2. avoid failing with messed up `ruby` in the users' `$PATH`
3. allow using the same installer for Homebrew on Linux (CC @Homebrew/linux).

The best strategy to begin with would be a line-by-line port.

Ideally the same thing would be done for the uninstaller but this isn't essential for a first PR.

itchyny:
Hi, I looked into the installer and translated into bash one line by one. Needs some more works (~not sure how to compare version string.. maybe I'll implement this with arrays~ parameter substitution works). I love Homebrew, thank you all.

## Why?
- resolves #217 
- `Future versions of macOS won’t include scripting language runtimes by default, and might require you to install additional packages. If your software depends on scripting languages, it’s recommended that you bundle the runtime within the app. (49764202)` https://developer.apple.com/documentation/macos_release_notes/macos_catalina_10_15_release_notes

## but shell script is hard to write
I introduced [shellcheck](https://github.com/koalaman/shellcheck) integration. This tool catches various common mistakes in shell script and there's no reason to follow it.

## a few notes
- Intentional changes
  - Basically I ported keeping the behavior of the original script but I intentionally introduced some changes.
    - Since user_chmods are already chmod 755, there's no need to chmod zsh_dirs bc13488cf700f1d6070c6f713189336a39fe5ce8
    - Print a newline after the donation link 589c65344a25e3fb4256a0c4b462bcc46cf068c3
    - Print audible bell only when the stdout is tty 4f0094f99a6b77019ad0af0f8615031e875f58ff
- Always check the length of the array when iterating through it
  - `for x in ${arr[@]}` splits an element containing a space,  `for x in "${a[@]}"` aborts with `unbound variable` when `set -u`, `"${a[@]-}"` emits an empty string when the array is empty, `"${a[@]+"${a[@]}"}"` (see https://stackoverflow.com/questions/7577052) works actually but the syntax is ugly. Checking the length is much simpler and readable.
- Including Linuxbrew to this script is not that difficult (maybe). Note that the flags of `stat` is not portable.


.github/workflows/tests.yml
```diff
@@ -31,8 +31,8 @@ jobs:
         export PATH=/usr/bin:/bin:/usr/local/bin
         ruby uninstall -d >/dev/null
         ruby uninstall -f >/dev/null
-        ruby install --force-curl
+        bash install.sh
```
reitermarkus: This step is now duplicated.

install
```diff
@@ -1,381 +1,3 @@
 #!/usr/bin/ruby
```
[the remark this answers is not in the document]
itchyny: Can this be a message to stderr?

README.md
```diff
@@ -3,7 +3,7 @@
 ## Install Homebrew
 
 ```bash
-/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+sh -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```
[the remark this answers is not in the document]
itchyny: Fixed by 2bf5f50bb1b861a328736c3bb5c01885c91f93e3.

.github/workflows/tests.yml
itchyny: Fixed by 2bf5f50bb1b861a328736c3bb5c01885c91f93e3.

README.md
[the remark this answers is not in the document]
itchyny: Fixed by c6160cd5d0f0b864d2e348dd9b85486d8844de93.

install
```diff
@@ -1,381 +1,3 @@
...
-Dir.chdir HOMEBREW_REPOSITORY do
-  system "git", "config", "--replace-all", "homebrew.analyticsmessage", "true"
-  system "git", "config", "--replace-all", "homebrew.caskanalyticsmessage", "true"
-end
-
-ohai "Next steps:"
-puts "- Run `brew help` to get started"
-puts "- Further documentation: "
-puts "    #{Tty.underline}https://docs.brew.sh#{Tty.reset}"
+system "sh", "-c", "sh -c \"$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)\""
```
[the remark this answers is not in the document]
itchyny: Fixed by c6160cd5d0f0b864d2e348dd9b85486d8844de93.

install.sh
```diff
@@ -0,0 +1,457 @@
...
+EOABORT
+)"
+elif version_gt "$macos_version" "$MACOS_LATEST_SUPPORTED" || \
+  version_lt "$macos_version" "$MACOS_OLDEST_SUPPORTED"; then
+  who="We"
+  what=""
+  if version_gt "$macos_version" "$MACOS_LATEST_SUPPORTED"; then
+    what="pre-release version"
+  else
+    who="$who (and Apple)"
```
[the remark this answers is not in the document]
itchyny: Fixed by ac3294e9f1e66e35a5e6d66bae04fd2bba756a05.

install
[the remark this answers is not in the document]
reitermarkus: Yes.

install
```diff
@@ -1,381 +1,8 @@
...
-Dir.chdir HOMEBREW_REPOSITORY do
-  system "git", "config", "--replace-all", "homebrew.analyticsmessage", "true"
-  system "git", "config", "--replace-all", "homebrew.caskanalyticsmessage", "true"
-end
-
-ohai "Next steps:"
-puts "- Run `brew help` to get started"
-puts "- Further documentation: "
-puts "    #{Tty.underline}https://docs.brew.sh#{Tty.reset}"
+system "bash", "-c", 'bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"'
```
reitermarkus: Use `exec` here.

install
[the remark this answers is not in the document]
reitermarkus: Maybe set an environment variable here and then use `warn` in the Bash script.

install
itchyny: Fixed by b8d3d1fc600919230ad1dd373dc9821ec17f24ea.

install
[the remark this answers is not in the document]
itchyny: Fixed by a8104cf0eabb684e5f886527908fcbc822cfaf1d.

sjackman:
Thank you very much for this work, @itchyny! The installer for Homebrew on Linux supports both macOS and Linux. Would you be interested in adding support for Linux to this PR?
See https://github.com/Linuxbrew/install/blob/master/install
and https://github.com/Linuxbrew/install/blob/master/install.sh

itchyny:
I'm actually thought that I can, but we should do this in separate PRs. Translation into bash and making it work in Linux are different things.

MikeMcQuaid:
> I'm actually thought that I can, but we should do this in separate PRs.

Agreed 👍

.github/workflows/reviewdog.yml
```diff
@@ -0,0 +1,15 @@
+name: reviewdog
```
MikeMcQuaid: Would be preferable to integrate this into the existing `tests.yml` and just run `brew install shellcheck` there, thanks. You may even be able to just use `brew style`.

.github/workflows/tests.yml
```diff
@@ -31,8 +31,6 @@ jobs:
         export PATH=/usr/bin:/bin:/usr/local/bin
         ruby uninstall -d >/dev/null
         ruby uninstall -f >/dev/null
-        ruby install --force-curl
```
MikeMcQuaid: Let's keep this argument.

README.md
```diff
@@ -3,7 +3,7 @@
 ## Install Homebrew
 
 ```bash
-/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"
+bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```
MikeMcQuaid:
```suggestion
/usr/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```

itchyny: What's the point here? On my environment bash locates at /bin/bash, and /usr/bin/bash does not exist. It's unlikely that /bin is not in $PATH.

.github/workflows/tests.yml
itchyny: Fixed by d6289ea9b573dda3c2804fe6dad6c336c92b7925

.github/workflows/reviewdog.yml
itchyny: Fixed by b38ead0ba0d665e4d7d833a21b7ca448c08a8806

README.md
MikeMcQuaid: Sorry, typo, I meant `/bin/bash`, fixed. We want to use the specific system Bash rather than a random on in the PATH.

.github/workflows/tests.yml
reitermarkus: Where was this even used before? I don't see it in the Ruby script.

itchyny:
Is there anything else to fix?

MikeMcQuaid:
@itchyny Apologies for the delay. This is on my list to review in depth and, as I hope you understand, we need to be very careful before merging a large change like this to the installer. I'll get to it as soon as I can.

MikeMcQuaid:
@itchyny Did a full review of this, some testing and made some tweaks/fixes where I saw them. As a starting point: this is an incredible first contribution, I'm really impressed 👏. I think it's now at a stage where I'm happy to have this merged. I just want to have a few other maintainers give their 👀 on it and ensure I only merge this at a time where I can be around to make any quick fixes should problems pop up. Let me know what you think of any of my changes. Again: great work!

@Homebrew/maintainers would appreciate as many of you as possible taking a look at this and offering any thoughts. Thanks!

install.sh
```diff
@@ -0,0 +1,459 @@
...
+echo
+
+# Use the shell's audible bell.
+if [[ -t 1 ]]; then
+  printf "\a"
+fi
+
+# Use an extra newline and bold to avoid this being missed.
+ohai "Homebrew has enabled anonymous aggregate formulae and cask analytics."
+echo "$(cat <<EOS
```
vitorgalvao: Why `echo` and `cat` at once? `cat` alone should suffice.

install.sh
```diff
@@ -0,0 +1,459 @@
+#!/bin/bash
+set -u
+
+# This script installs to /usr/local only. To install elsewhere (which is
+# unsupported) you can untar https://github.com/Homebrew/brew/tarball/master
+# anywhere you like.
+HOMEBREW_PREFIX="/usr/local"
```
vitorgalvao: Bash has *a lot* of pitfalls and places where things can go wrong if we’re distracted.

I’d feel safer if every variable would be `readonly` when possible (i.e. if it’s not you know it will change) and used single quotes unless double quotes are necessary.

install.sh
```diff
@@ -0,0 +1,459 @@
...
+fi
+tty_mkbold() { tty_escape "1;$1"; }
+tty_underline="$(tty_escape "4;39")"
+tty_blue="$(tty_mkbold 34)"
+tty_red="$(tty_mkbold 31)"
+tty_bold="$(tty_mkbold 39)"
+tty_reset="$(tty_escape 0)"
+
+shell_join() {
+  local arg
```
vitorgalvao: Similarly, `local` variables should be `local -r` (local and readonly).

install.sh
```diff
@@ -0,0 +1,459 @@
...
+HOMEBREW_CACHE="${HOME}/Library/Caches/Homebrew"
+BREW_REPO="https://github.com/Homebrew/brew"
+
+# TODO: bump version when new macOS is released
+MACOS_LATEST_SUPPORTED="10.15"
+# TODO: bump version when new macOS is released
+MACOS_OLDEST_SUPPORTED="10.13"
+
+# no analytics during installation
+export HOMEBREW_NO_ANALYTICS_THIS_RUN=1
```
vitorgalvao: To make a variable readonly and exported at the same time, `declare -rx` works.

install.sh
```diff
@@ -0,0 +1,459 @@
...
+  trap '/usr/bin/sudo -k' EXIT
+fi
+
+# Things can fail later if `pwd` doesn't exist.
+# Also sudo prints a warning message for no good reason
+cd "/usr" || exit 1
+
+####################################################################### script
+if [[ "$OSTYPE" == "linux-gnu" ]]; then
+  abort "$(cat <<'EOABORT'
```
vitorgalvao: Again, it’s weird to me that a function is called and then `cat` is called inside. In this case, there’s also the case that `EOABORT` is quoted, which is inconsistent with the rest of the code.

install.sh
```diff
@@ -0,0 +1,459 @@
...
+}
+version_ge() {
+  [[ "${1%.*}" -gt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -ge "${2#*.}" ]]
+}
+version_lt() {
+  [[ "${1%.*}" -lt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -lt "${2#*.}" ]]
+}
+
+should_install_command_line_tools() {
+  if version_gt "$macos_version" "10.13"; then
```
vitorgalvao: All variables should be `${var}` instead of `$var`. That’s one of those simple things that can save us from a world of trouble by doing everywhere, and we definitely do not want this script to fail due to mistakes.

In addition, it makes the script consistent throughout, since several cases already use the former notation.

install.sh
```diff
@@ -0,0 +1,459 @@
...
+  if version_gt "$macos_version" "$MACOS_LATEST_SUPPORTED"; then
+    what="pre-release version"
+  else
+    who+=" (and Apple)"
+    what="old version"
+  fi
+  ohai "You are using macOS ${macos_version}."
+  ohai "${who} do not provide support for this ${what}."
+
+  echo "$(cat <<EOS
```
vitorgalvao: Again, I don’t get the use of `cat`, as `echo` alone would suffice. As we can make it look nicer by keeping the indentation and removing it on output, like so:

```bash
echo "
  Some message
  With ${variables}
" sed -E 's/^ {2}//'
```

Replacing the `2` with however many spaces we have at the start of the line.

sjackman: I personally prefer simpler and less magical (no undenting).
```sh
cat <<EOS
Hello, world!
EOS
```

vitorgalvao: I disagree that the `sed` call is either complex or magical, especially when you consider the script as a whole, but I won’t bikeshed it. Not using `echo` and `cat` at the same time should be considered, though.

There’s also

```bash
cat <<-EOS
	Hello
	World
EOS
```

The `<<-` removes the indentation (not as good as Ruby’s `<<~`), but it only works with tabs.

dawidd6:
Initial Linux support on top of this PR branch: https://github.com/dawidd6/install/tree/install-sh-linux.

install.sh
MikeMcQuaid: Undenting is also probably not worth the complexity, agreed.

install.sh
```diff
@@ -0,0 +1,459 @@
...
+
+# Things can fail later if `pwd` doesn't exist.
+# Also sudo prints a warning message for no good reason
+cd "/usr" || exit 1
+
+####################################################################### script
+if [[ "$OSTYPE" == "linux-gnu" ]]; then
+  abort "$(cat <<'EOABORT'
+  To install Homebrew on Linux, paste at a terminal prompt:
+    sh -c "$(curl -fsSL https://raw.githubusercontent.com/Linuxbrew/install/master/install.sh)"
```
vitorgalvao: We should be explicit (`/bin/bash -c`). `/bin/sh` is often a link to something else (e.g. in Ubuntu it [used to point to `bash` and later `dash](https://wiki.ubuntu.com/DashAsBinSh)`), which might mean unexpected breakage when we least expect it.

dawidd6: But it's a `sh` script actually, not `bash`.

https://github.com/Linuxbrew/install/blob/master/install.sh

Look at the shebang.

zbeekman: @dawidd6 in this PR it's `#!/bin/bash`. On most systems `/bin/sh` runs bash or some other interpreter in compatibility mode, IIRC.

sjackman: When you run `sh -c "…"` the shebang line is ignored. The script will be run using `sh`, not with whatever the shebang line specifies. We should either remove Bash-isms from the script and test with `dash`, or change the installation instructions to `bash -c "…"`.

install.sh
reitermarkus: Or we could just `echo` every line separately, which also aligns nicely with `ohai`, as seen below.

MikeMcQuaid:
@vitorgalvao and others: I appreciate the style suggestions but as this PR is otherwise ready to go (and I'm limited in the windows in which I can merge it and it's limiting other things going in until it's done) I'm going to merge as-is and accept PRs to tweak style. Ideally all style checks will be enforced by `shellcheck` or similar.

The `sh -c` I'm also leaving as-is as it's what was in the existing code (which I think is worthing sticking as close to as possible for this PR) and it should go away as soon as we get Linux support in here.

MikeMcQuaid:
Thanks so much for your first, incredible contribution! Without people like you submitting PRs we couldn't run this project. You rock, @itchyny!

itchyny:
Great, thank you all for throughout reviews! I'm honored and I love Homebrew!

PR #259, opened 2020-02-12
Support custom brew repository and branch for development

nachshon:
Add support in the installer for a custom branch and git repository

### DIFF
--- a/install
+++ b/install
@@ -5,7 +5,16 @@
 HOMEBREW_PREFIX = "/usr/local".freeze
 HOMEBREW_REPOSITORY = "/usr/local/Homebrew".freeze
 HOMEBREW_CACHE = "#{ENV["HOME"]}/Library/Caches/Homebrew".freeze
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Sorry, we don't want to support this in the official installer.

nachshon:
> Sorry, we don't want to support this in the official installer.

Why not? We run an organisational mirror (behind a firewall) and having to modify installers locally adds a large maintenance burden when updating from upstream.

At the same time I am working on a similar pull request for the "brew" repository so that we can work from an alternative repository. Should I not do that?

Thanks,

MikeMcQuaid:
> We run an organisational mirror (behind a firewall) and having to modify installers locally adds a large maintenance burden when updating from upstream.

If your organisation has custom requirements I suggest you use a custom installer based on details from https://docs.brew.sh/Installation.

> At the same time I am working on a similar pull request for the "brew" repository so that we can work from an alternative repository. Should I not do that?

Read the documentation, please (`man brew`). It's already supported.

PR #260, opened 2020-02-16
Update README.md

Danrancan:
1) Updated uninstall instructions to take ownership and execute the uninstall script "raw" file downloaded from github.
2) Updated uninstall instructions to appeal to noobs and those unfamiliar with the command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,4 +18,10 @@ Install Homebrew on Linux and Windows 10 Subsystem for Linux: https://docs.brew.
 /usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall)"
 ```
 
-Download the uninstall script and run `./uninstall --help` to view more uninstall options.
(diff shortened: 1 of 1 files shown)


itchyny:
Note that `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall)" -- --help` prints the help of the uninstaller.

MikeMcQuaid:
I would suggest instead doing `/usr/bin/ruby ./uninstall --help` and then it can be a single step again.

Danrancan:
I think just noting that `/usr/bin/ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall)" --help` one liner is probably the easiest for newbs and the most newb friendly.

itchyny:
Please note that `-- --help`, not just `--help`.

MikeMcQuaid:
Either is fine.

PR #267, opened 2020-03-02
Linux support for the installer

dawidd6:


install.sh
```diff
@@ -1,12 +1,36 @@
...
 set -u
 
-# This script installs to /usr/local only. To install elsewhere (which is
-# unsupported) you can untar https://github.com/Homebrew/brew/tarball/master
+# First check if the OS is Linux.
+if [[ "$(uname)" = "Linux" ]]; then
+    HOMEBREW_ON_LINUX=true
+else
+    HOMEBREW_ON_LINUX=false
+fi
```
MikeMcQuaid: Use two-character indentation only.

install.sh
```diff
@@ -1,12 +1,36 @@
 #!/bin/bash
 set -u
 
-# This script installs to /usr/local only. To install elsewhere (which is
-# unsupported) you can untar https://github.com/Homebrew/brew/tarball/master
+# First check if the OS is Linux.
+if [[ "$(uname)" = "Linux" ]]; then
+    HOMEBREW_ON_LINUX=true
+else
+    HOMEBREW_ON_LINUX=false
```
MikeMcQuaid: Would strongly suggest not setting this on macOS so you can do `[ -z "$HOMEBREW_ON_LINUX" ]`. Alternatively, might be a bit more readable to do `MACOS` (and `LINUX`) or `HOMEBREW_ON_MACOS`.

install.sh
```diff
@@ -127,26 +154,24 @@ exists_but_not_writable() {
...
   [[ " $(id -G "$USER") " != *" $(get_group "$1") "*  ]]
 }
 
-# USER isn't always set so provide a fall back for the installer and subprocesses.
-if [[ -z "$USER" ]]; then
-  USER="$(chomp "$(id -un)")"
-  export USER
-fi
+# USER isn't always set so always get it from `id` for the installer and subprocesses.
+USER="$(chomp "$(id -un)")"
```
MikeMcQuaid: Why this change if `USER` is already set?

install.sh
```diff
@@ -294,15 +317,15 @@ if [[ "${#chowns[@]}" -gt 0 ]]; then
...
+  ohai "The following existing directories will have their group set to ${tty_underline}${GROUP}${tty_reset}:"
   printf "%s\n" "${chgrps[@]}"
 fi
 if [[ "${#mkdirs[@]}" -gt 0 ]]; then
   ohai "The following new directories will be created:"
   printf "%s\n" "${mkdirs[@]}"
 fi
 
-if should_install_command_line_tools; then
+if ! "$HOMEBREW_ON_LINUX" && should_install_command_line_tools; then
```
MikeMcQuaid: Instead I'd suggest `should_install_command_line_tools` return false if `HOMEBREW_ON_LINUX` is set. This will simplify all the callers.

install.sh
dawidd6: Because if it isn't set, the script will fail when checking if it exists, because of `set -u` line at the top of the script.

MikeMcQuaid: Could put this check above that `set`; want to respect the existing variable if it's there.

itchyny: Or you can use `if [[ -z "${USER-}" ]]`.

dawidd6: That's neat, thanks @itchyny!

.github/workflows/tests.yml
```diff
@@ -4,7 +4,7 @@ on:
     branches: master
   pull_request: []
 jobs:
-  tests:
+  tests-macos:
```
MikeMcQuaid: Would be good/more consistent with Homebrew/brew to use a single `tests` here with conditionals. I can do that if necessary!

dawidd6: Initially I've done it like that, but then went for separate jobs.

Will do then.

sjackman:
Thank you, Dawid! ❤️

PR #269, opened 2020-03-03
Update tests.yml

imbsky:
- `[]` is not required.
- It's dangerous to use master, so please specify a tag for any Action.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -2,7 +2,7 @@ name: GitHub Actions CI
 on:
   push:
     branches: master
-  pull_request: []
+  pull_request:
 jobs:
   tests:
     strategy:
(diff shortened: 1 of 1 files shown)


imbsky:
Updated.

.github/workflows/tests.yml
```diff
@@ -24,7 +24,7 @@ jobs:
     - name: Install RubyGems
       run: |
         gem install --user bundler
-        bundle install --retry=3 --jobs=4
+        bundle install --retry=3 --jobs=$(getconf _NPROCESSORS_ONLN)
```
imbsky: Actions VMs are not that powerful, so 4 is too many. So, specify the number of processors on the machine. This allows the same code to be used in the future even if the Actions VM becomes more powerful.

MikeMcQuaid:
I don't see the need for any of these changes. As-is they are consistent with the other Homebrew organisation repos, too.

.github/workflows/tests.yml
EricFromCanada: Homebrew uses its own worker nodes.

MikeMcQuaid: Not for this repo, just FYI.

MikeMcQuaid:
Passing on this for now, sorry!

PR #270, opened 2020-03-04
install.sh: print linux specific info after completion

dawidd6:
Port of this section: https://github.com/Linuxbrew/install/blob/2cc4266960a388510cff7325d4172e3344d7eecf/install#L449-L464

cc @sjackman

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -484,6 +484,40 @@ EOS
   execute "git" "config" "--replace-all" "homebrew.caskanalyticsmessage" "true"
 )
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Some shellcheck CI failures (which could be pulled into a new step for readability?). Otherwise looks good.

MikeMcQuaid:
Thanks again @dawidd6!

sjackman:
Thank you, Dawid! 🌟

PR #271, opened 2020-03-04
workflows: new shellcheck and PATH setting steps

dawidd6:
As @MikeMcQuaid said here: https://github.com/Homebrew/install/pull/270#issuecomment-594464647

Shellcheck has now its own step and PATH is set in a separate step too.

I also deleted the last `uninstall` line, as it seems to be not much needed?


MikeMcQuaid:
Thanks again @dawidd6!

PR #272, opened 2020-03-04
install.sh: check sudo and ask for prefix

dawidd6:
Port: https://github.com/Linuxbrew/install/blob/2cc4266960a388510cff7325d4172e3344d7eecf/install#L212-L217

With this change, the installer asks interactively (not on CI) where to install Homebrew (`/home/linuxbrew/.linuxbrew` with sudo or `~/.linuxbrew` without sudo).

cc @sjackman

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -6,7 +6,9 @@ if [[ "$(uname)" = "Linux" ]]; then
   HOMEBREW_ON_LINUX=1
 fi
 
-# This script installs to /usr/local (macOS) or /home/linuxbrew/.linuxbrew (Linux) only.
+# On macOS, this script installs to /usr/local only.
+# On Linux, it installs to /home/linuxbrew/.linuxbrew if you have sudo access
+# and ~/.linuxbrew otherwise.
 # To install elsewhere (which is unsupported)
 # you can untar https://github.com/Homebrew/brew/tarball/master
 # anywhere you like.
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -53,6 +54,19 @@ tty_red="$(tty_mkbold 31)"
 tty_bold="$(tty_mkbold 39)"
 tty_reset="$(tty_escape 0)"
 
+have_sudo() {
```
MikeMcQuaid: This naming (or `execute_sudo`) feels a bit strange given sudo is probably always available but just not accessible and that this returns true unconditionally on macOS.

install.sh
```diff
@@ -187,6 +206,29 @@ fi
 cd "/usr" || exit 1
 
 ####################################################################### script
+if [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
+  if [[ -n "${CI-}" ]] || [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] || [[ -w "/home/linuxbrew" ]] || [[ -w "/home" ]]; then
```
MikeMcQuaid: What's this for?

dawidd6: I have no idea, just ported it from `Linuxbrew/install`. @sjackman might know it.

install.sh
dawidd6: `have_sudo_access` might be a better fit here. Thoughts?

MikeMcQuaid: Perfect!

MikeMcQuaid: Thoughts: checking for access on macOS too makes sense, I think, and just erroring out if you don't have `sudo` there?

MikeMcQuaid:
Thanks again @dawidd6!

sjackman:
Fantastic! Thanks again, Dawid! 🏆

PR #276, opened 2020-03-07
Add Linux uninstall support

gromgit:
Basically three differences:
1. Parameterize default Homebrew path
2. GNU find vs. BSD find
3. .DS_Store not applicable in Linux

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,8 +46,7 @@ jobs:
           echo "::add-path::/home/linuxbrew/.linuxbrew/bin:/usr/bin:/bin"
         fi
 
-    - name: Test uninstaling GitHub Actions macOS Homebrew
-      if: runner.os == 'macOS'
+    - name: Test uninstaling GitHub Actions Homebrew
       run: |
         ruby uninstall -d >/dev/null
         ruby uninstall -f >/dev/null
@@ -56,10 +55,8 @@ jobs:
       run: |
         bash install.sh
         brew install ack
-        if [ "$RUNNER_OS" = "macOS" ]; then
-          ruby uninstall -f >/dev/null
-          bash install.sh
-        fi
+        ruby uninstall -f >/dev/null
+        bash install.sh
 
     - name: Install Shellcheck
       run: brew install shellcheck
@@ -68,7 +65,6 @@ jobs:
       run: shellcheck *.sh
 
     - name: Test uninstalling newly installed Homebrew
-      if: runner.os == 'macOS'
       run: |
         ruby uninstall -d >/dev/null
         ruby uninstall -f >/dev/null
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -46,7 +46,8 @@ jobs:
           echo "::add-path::/home/linuxbrew/.linuxbrew/bin:/usr/bin:/bin"
         fi
 
-    - name: Test uninstaling GitHub Actions Homebrew
+    - name: Test uninstalling GitHub Actions macOS Homebrew
+      if: runner.os == 'macOS'
       run: |
         ruby uninstall -d >/dev/null
         ruby uninstall -f >/dev/null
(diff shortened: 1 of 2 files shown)


dawidd6:
You can also tweak the CI config to run the uninstall on Linux too.

uninstall
```diff
@@ -16,6 +16,9 @@ options = {
 # global status to indicate whether there is anything wrong.
 @failed = false
 
+@is_mac = RUBY_PLATFORM.to_s.downcase.include? "darwin"
```
MikeMcQuaid:
```suggestion
@is_mac = !RUBY_PLATFORM[/darwin/].nil?
```

uninstall
```diff
@@ -224,7 +227,13 @@ paths = %w[Frameworks bin etc include lib opt sbin share var]
...
         .map(&:to_s)
 if paths.any?
-  args = %w[-E] + paths + %w[-regex .*/info/([^.][^/]*\.info|dir)]
+  if @is_mac
+    # BSD find
+    args = %w[-E] + paths + %w[-regex .*/info/([^.][^/]*\.info|dir)]
+  else
+    # GNU find
+    args = paths + %w[-regextype posix-extended -regex .*/info/([^.][^/]*\.info|dir)]
+  end
```
MikeMcQuaid:
```suggestion
  args = if @is_mac
    # BSD find
    %w[-E] + paths + %w[-regex .*/info/([^.][^/]*\.info|dir)]
  else
    # GNU find
    paths + %w[-regextype posix-extended -regex .*/info/([^.][^/]*\.info|dir)]
  end
```

uninstall
```diff
@@ -272,7 +281,7 @@ paths = %w[bin etc include lib opt sbin share var
         .map { |p| HOMEBREW_PREFIX/p }
         .select(&:exist?)
         .map(&:to_s)
-if paths.any?
+if @is_mac && paths.any?
   args = paths + %w[-name .DS_Store]
```
MikeMcQuaid:
```suggestion
if paths.any?
  args = paths
  args += %w[-name .DS_Store] if @is_mac 
```

.github/workflows/tests.yml
```diff
@@ -46,8 +46,7 @@ jobs:
           echo "::add-path::/home/linuxbrew/.linuxbrew/bin:/usr/bin:/bin"
         fi
 
-    - name: Test uninstaling GitHub Actions macOS Homebrew
-      if: runner.os == 'macOS'
```
dawidd6: Leave this line. It's for uninstalling Homebrew which is preinstalled on macOS but not on Ubuntu.

uninstall
gromgit: I can see now that there's a bug in my original change, but shouldn't the logic in this section be:

```
if paths.any?

  # Delete .DS_Store on macOS only...
  if @is_mac
    args = paths + %w[-name .DS_Store]
    if options[:dry_run]
      args << "-print"
    else
      args << "-delete"
    end
    puts "Would delete:" if options[:dry_run]
    sudo "/usr/bin/find", *args
  end

  # ...but remove all empty subdirs on all OSes
  args = paths + %w[-depth -type d -empty]
  if options[:dry_run]
    args << "-print"
  else
    args += %w[-exec rmdir {} ;]
  end
  puts "Would remove directories:" if options[:dry_run]
  sudo "/usr/bin/find", *args

end
```

.github/workflows/tests.yml
gromgit: Noted. Will revert in a subsequent commit (just noticed a speling error in the original).

MikeMcQuaid: My bad! If you could fix those it'd be great ❤️

uninstall
MikeMcQuaid: Yup, rereading it again: I agree with your newly suggested version 👍

gromgit:
Any other issues? If not, I can force-push a squashed commit, then PR the bash version. Thanks much!

MikeMcQuaid:
Looks great, thanks again @gromgit!

PR #278, opened 2020-03-10
Translate uninstaller to Bash

gromgit:
For clearer 1-to-1 correspondence with the original Ruby code,
I've adopted the naming convention:

  Class::method -> Class_method()


uninstall.sh
```diff
@@ -0,0 +1,348 @@
...
+
+### Shell equivalents to standard Ruby methods
+
+Kernel_abort() {
+  [[ $1 ]] && echo "$1" 2>/dev/null
+  exit 1
+}
+
+String_strip() {
+  local s; for s in "$@"; do
```
dawidd6: `local s` might be better to be in a separate line. Readability and consistency with `Dir_children`.

uninstall.sh
```diff
@@ -0,0 +1,348 @@
...
+fi
+Tty_mkbold() { Tty_escape "1;${1:-39}"; }
+Tty_blue=$(Tty_mkbold 34)
+Tty_red=$(Tty_mkbold 31)
+Tty_bold=$(Tty_mkbold 39)
+Tty_reset=$(Tty_escape 0)
+
+Array_shell_s() {
+  local result=("$1"); shift
+  local s; for s in "$@"; do
```
dawidd6: Same as above.

uninstall.sh
gromgit: Good point, will fix.

uninstall.sh
gromgit: Good point, will fix.

gromgit:
Clarification request from a complete Ruby newbie. In the following code from the Ruby `uninstall`:
```ruby
homebrew_files.each do |file|
  if options[:dry_run]
    puts "Would delete #{file}"
  else
    begin
      FileUtils.rm_rf(file)
    rescue => e
      warn "Failed to delete #{file}"
      puts e.message
      @failed = true
    end
  end
end
```
Can `FileUtils.rm_rf()` throw an exception? If I'm reading [the relevant RubyDoc](https://ruby-doc.org/stdlib-2.4.1/libdoc/fileutils/rdoc/FileUtils.html#method-c-rm_rf) correctly, it never does, which explains a significant difference with the bash equivalent:
```sh
for file in "${homebrew_files[@]}"; do
  if [[ $opt_dry_run ]]; then
    echo "Would delete ${file}"
  else
    err=$(rm -fr "$file" 2>&1)
    if [[ $err ]]; then
      warn "Failed to delete ${file}"
      echo "$err"
      failed=true
    fi
  fi
done
```
When both instance try to remove a protected directory like `/usr/local/Cellar`, the Ruby code seems to silently ignore the failure, while the bash code registers an `rm` "Permission denied" failure.

uninstall.sh
```diff
@@ -0,0 +1,350 @@
...
+shopt -s extglob
+
+# Set up temp dir
+tmpdir=/tmp/uninstall.$$
+mkdir -p "$tmpdir" || Kernel_abort "Unable to create temp dir '$tmpdir'"
+trap '
+  rm -fr "$tmpdir"
+  # Invalidate sudo timestamp before exiting
+  /usr/bin/sudo -k
+' exit
```
itchyny: I think it's common to use UPPER CASE for traps.

uninstall.sh
```diff
@@ -0,0 +1,350 @@
...
+if [[ -t 1 ]]; then
+  Tty_escape() { printf "\033[%sm" "$1"; }
+else
+  Tty_escape() { :; }
+fi
+Tty_mkbold() { Tty_escape "1;${1:-39}"; }
+Tty_blue=$(Tty_mkbold 34)
+Tty_red=$(Tty_mkbold 31)
+Tty_bold=$(Tty_mkbold 39)
+Tty_reset=$(Tty_escape 0)
```
itchyny: I believe case consistency with installer.sh matters.

uninstall.sh
```diff
@@ -0,0 +1,350 @@
...
+
+Array_shell_s() {
+  local result=("$1"); shift
+  local s
+  for s in "$@"; do
+    result+=("${s// /\\ }")
+  done
+  echo -n "${result[*]}"
+}
```
itchyny: Why is this implementation different from `shell_join` in installer.sh?

itchyny:
Do you have plan to update the `uninstall` script to use `uninstall.sh` like the installer? After that, I think we can remove the Ruby things (Gemfile, Rubocop, CI), but I'd like to opinions from main maintainers.

uninstall.sh
```diff
@@ -0,0 +1,350 @@
...
+  if [[ $failed == true ]]; then
+    warn "Homebrew partially uninstalled (but there were steps that failed)!"
+    echo "To finish uninstalling rerun this script with \`sudo\`."
+  else
+    ohai "Homebrew uninstalled!"
+  fi
+fi
+
+Dir_children "$HOMEBREW_REPOSITORY" "$HOMEBREW_PREFIX" \
+| sort -u > "$tmpdir/residual_files"
```
itchyny: Let's add indentation here.

uninstall.sh
```diff
@@ -0,0 +1,350 @@
...
+    -d|--dry-run) opt_dry_run=1;;
+    -h|--help) usage;;
+    *) warn "Unrecognized option: '$1'"; usage 1;;
+  esac
+  shift
+done
+
+if [[ ${#homebrew_prefix_candidates[@]} -eq 0 ]]; then # Attempt to locate Homebrew unless `--path` is passed
+  prefix=$(brew --prefix)
+  [[ $prefix ]] && homebrew_prefix_candidates+=("$prefix")
```
itchyny: I think it's preferable to write `-n` explicitly.

uninstall.sh
gromgit: I've seen it both ways in about equal measure, but OK, I'll fix that.

uninstall.sh
gromgit: OK. As I mentioned in the commit text, this was to parallel "the Ruby way" for easier translation, but I'll change that.

uninstall.sh
gromgit: Mostly because it didn't click in my mind that they were actually the same function. Will replace with `shell_join`.

uninstall.sh
gromgit: OK.

uninstall.sh
gromgit: Sorry, where exactly?

gromgit:
> Do you have plan to update the `uninstall` script to use `uninstall.sh` like the installer? After that, I think we can remove the Ruby things (Gemfile, Rubocop, CI), but I'd like to opinions from main maintainers.

Good point. I'd not actually looked at `install`, but yeah, `uninstall` should probably do the same.

uninstall.sh
gromgit: Along with any other "twinned" functions.

MikeMcQuaid:
> Do you have plan to update the `uninstall` script to use `uninstall.sh` like the installer? After that, I think we can remove the Ruby things (Gemfile, Rubocop, CI), but I'd like to opinions from main maintainers.

Yep, also think this would be good.

uninstall.sh
MikeMcQuaid: Thanks both of you. Agreed that consistency with install.sh would be good for now.

uninstall.sh
MikeMcQuaid: Agreed that consistency with install.sh would be good for now.

uninstall.sh
MikeMcQuaid:
```suggestion
  | sort -u > "$tmpdir/residual_files"
```
presumably

gromgit:
OK, made all the suggested changes, including replacing `uninstall` with its shim variant.

itchyny:
Would you change the p/r title contains the word uninstaller?

gromgit:
> Would you change the p/r title contains the word uninstaller?

Done.

uninstall.sh
gromgit: OK, done.

uninstall.sh
```diff
@@ -0,0 +1,388 @@
...
+  Linux)
+    ostype=linux
+    homebrew_prefix_default=/home/linuxbrew/.linuxbrew
+  ;;
+  Darwin)
+    ostype=macos
+    homebrew_prefix_default=/usr/local
+    # Stock macOS doesn't have a realpath, so we make do with builtin Python
+    realpath() {
+      /usr/bin/python -c 'import os,sys;print(os.path.realpath(sys.argv[1]))' "$1"
```
itchyny: I'm afraid introducing dependency on _builtin Python_ is not a good option. Translation to Bash started from the deprecation of scripting languages (see [49764202](https://developer.apple.com/documentation/macos_release_notes/macos_catalina_10_15_release_notes)). I think Python will be installed while installing the command line tools but is it guaranteed to keep it after uninstallation? @MikeMcQuaid What do you think?

uninstall.sh
```diff
@@ -0,0 +1,388 @@
...
+    *) warn "Unrecognized option: '$1'"; usage 1;;
+  esac
+  shift
+done
+
+if [[ ${#homebrew_prefix_candidates[@]} -eq 0 ]]; then # Attempt to locate Homebrew unless `--path` is passed
+  prefix=$(brew --prefix)
+  [[ -n $prefix ]] && homebrew_prefix_candidates+=("$prefix")
+  # shellcheck disable=SC2230  # "which" is a legitimate fallback
+  prefix=$(command -v brew) || prefix=$(which brew) || prefix=""
```
itchyny: I see this is a direct translation from the original code, but I'm not sure we still need this fallback code (`command -v brew` fails but `which brew` succeed?). In ruby it invokes the executable (maybe `/usr/bin/command`), but in bash it's a builtin function.

uninstall.sh
gromgit: > [49764202](https://developer.apple.com/documentation/macos_release_notes/macos_catalina_10_15_release_notes))

Oh. That's not good. Will find a pure-bash replacement.

uninstall.sh
gromgit: Good point, will fix.

uninstall.sh
MikeMcQuaid: agreed with @itchyny. This is the method I tend to use: https://stackoverflow.com/a/35996728/5355

gromgit: OK, here's the version I'm inclined to go with. It takes care not to output anything except the canonicalized pathname, or nothing at all if it can't be canonicalized for whatever reason, especially link chains 100 or more levels deep (usually a symlink cycle):
```sh
realpath() {
  local dir base newpath
  dir=$(dirname "$1")
  base=$(basename "$1")
  (
    cd -P "$dir" &> /dev/null || exit 1
    redirs=0
    while [[ -L $base && $((redirs++)) -lt 100 ]]; do
      newpath=$(readlink "$base" 2>/dev/null) || exit 1
      dir=$(dirname "$newpath")
      base=$(basename "$newpath")
      cd -P "$dir" &> /dev/null || exit 1
    done
    if [[ -L $base ]]; then
      exit 1
    else
      echo "$(pwd -P)/$base"
    fi
  )
}
```
Comments?

MikeMcQuaid: That seems excessively complicated. `cd` and `pwd` with an argument should be sufficient.

gromgit: That would only work if the entire path points to an existing directory, so it'll fail when we test, say, `$HOMEBREW_PREFIX/bin/brew`, as is already done in the uninstaller.

MikeMcQuaid: If it doesn't exist it can/will return an empty string which can itself be filtered out.

gromgit: Are you talking about:
```
cd "$(dirname "$target")" && echo "$(pwd -P)/$(basename "$target")"
```
? I guess that might work.

MikeMcQuaid: Aye, thanks!

gromgit:
Very sorry for the long delay, had to deal with some personal issues. I've fixed everything previously discussed, and also squashed a bug related to cleaning up Info files.

MikeMcQuaid:
No worries @gromgit, hope you and your family are all OK. Any thoughts @itchyny @dawidd6?

MikeMcQuaid:
Merged, thanks @gromgit! Would welcome more PRs from you or anyone else to clean up and make the installer/uninstaller (somehow?) share code and generally be more consistent.

PR #280, opened 2020-03-11
Separately fetch commits and tags

jonchang:
This is necessary for older gits, which will only fetch the tags and not
e.g., the `master` ref.

Tested on git 1.8.3

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -482,6 +482,7 @@ ohai "Downloading and installing Homebrew..."
   # ensure we don't munge line endings on checkout
   execute "git" "config" "core.autocrlf" "false"
 
+  execute "git" "fetch" "origin" "--force"
   execute "git" "fetch" "origin" "--tags" "--force"
 
   execute "git" "reset" "--hard" "origin/master"


MikeMcQuaid:
> This is necessary for older gits, which will only fetch the tags and not
> e.g., the `master` ref.
> 
> Tested on git 1.8.3

Could this be gated for older Gits? Its a bit of a slow operation to do unconditionally.

jonchang:
> > This is necessary for older gits, which will only fetch the tags and not
> > e.g., the `master` ref.
> > Tested on git 1.8.3
> 
> Could this be gated for older Gits? Its a bit of a slow operation to do unconditionally.

If the tags are already fetched via `git fetch origin` (the default in later gits), the command `git fetch origin --tags` basically does nothing.

```
% time git fetch origin -q
git fetch origin -q  3.81s user 1.04s system 43% cpu 11.095 total
% time git fetch origin --tags
git fetch origin --tags  0.06s user 0.06s system 6% cpu 1.714 total
```

MikeMcQuaid:
Ok, thanks, seems fine then!

PR #281, opened 2020-03-11
Add set -e to exit upon errors

notklaatu:
Testing on Linux, running install.sh on a system that does not yet have Git installed results in apparent success.

Adding ``set -e`` ensures that the install script exits when errors are thrown.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,5 +1,6 @@
 #!/bin/bash
 set -u
+set -e
 
 # First check if the OS is Linux.
 if [[ "$(uname)" = "Linux" ]]; then


MikeMcQuaid:
This isn't going to work as-is because the entire script assumes that errors are handled differently. I do think a `set -e` approach is probably a sensible one but it'll need a much wider refactor.

notklaatu:
What kind of refactor do you have in mind? Do you mean the way errors are handled?

Are you open to just adding some ``which`` checks for vital commands (git being the most significant, but I think it's arguable that ``gcc`` is also necessary because the suggested ``brew install gcc`` command, after install, fails on Linux, as far as I can tell). I'd be interested in adding/changing that.

MikeMcQuaid:
> What kind of refactor do you have in mind? Do you mean the way errors are handled?

A refactor such that `set -e` works as you've intended. As you can see: this fails CI as-is.

> Are you open to just adding some `which` checks for vital commands (git being the most significant, but I think it's arguable that `gcc` is also necessary because the suggested `brew install gcc` command, after install, fails on Linux, as far as I can tell). I'd be interested in adding/changing that.

Yes, that would be fine too. I think the only `which` checks necessary in this script are those actually used by this script e.g. `git`. Homebrew/brew is where checks for e.g. `gcc` should be present (and I'm pretty sure they already are when they are needed).

notklaatu:
The install.sh script returns success on Linux even when it fails to install, so something's not working as expected. I'll see if I can dig into it and make some changes. 

Is the CI testing on both Linux and Mac?

MikeMcQuaid:
> The install.sh script returns success on Linux even when it fails to install, so something's not working as expected. I'll see if I can dig into it and make some changes.

Yes, I'm not claiming it's without bugs 😄. It does require a larger change than this to fix that bug, though.

notklaatu:
Looks like ''git'' isn't checked for on Linux, only on Mac versions greater-than $some_version.  I'll mess around with it to see what the most elegant way to add a check for Linux and re-submit. Thanks for the feedback and guidance.

PR #282, opened 2020-03-12
check for Git, abort if not present

notklaatu:
Check to ensure Git is installed before continuing. Abort upon failure.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -213,7 +213,11 @@ cd "/usr" || exit 1
 
 ####################################################################### script
 if should_install_git; then
-    abort "You must install Git before installing Homebrew."
+    abort "$(cat <<EOABORT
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -206,6 +212,10 @@ fi
 cd "/usr" || exit 1
 
 ####################################################################### script
+if should_install_git; then
+    abort "You must install Git before installing Homebrew."
```
jonchang: Should this link to https://docs.brew.sh/Installation?

notklaatu: Great idea. Added link.

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @notklaatu!

PR #283, opened 2020-03-12
allow override of install prefix, separate PREFIX from PREFIX_DEFAULT

notklaatu:
allow override of install prefix, separate PREFIX from PREFIX_DEFAULT to retain required(?)/expected dir structure

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -22,7 +22,8 @@ if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
   CHGRP="/usr/bin/chgrp"
   GROUP="admin"
 else
-  HOMEBREW_PREFIX_DEFAULT="/home/linuxbrew/.linuxbrew"
+  HOMEBREW_PREFIX="${HOMEBREW_PREFIX:-/home}"
+  HOMEBREW_PREFIX_DEFAULT="${HOMEBREW_PREFIX}/linuxbrew/.linuxbrew"
   HOMEBREW_CACHE="${HOME}/.cache/Homebrew"
 
   STAT="stat --printf"


MikeMcQuaid:
Can you explain what the advantage is here over the existing code? I wouldn't say `linuxbrew/.linuxbrew` is a necessary subdirectory/prefix format.

notklaatu:
The advantages, the way I see it:

HOMEBREW_PREFIX allows the user to set with an env var what the prefix will be (ie, /opt or /usr/local or whatever). The user doesn't need to know or care about what directories Homebrew gets put into beyond the prefix.

HOMEBREW_PREFIX_DEFAULT allows the default for the Hombrew-specific directory structure to be controlled by the Homebrew project. I don't 

As is, the user has no control aside from editing the install script.

And if it's set like this:

```
  HOMEBREW_PREFIX="${HOMEBREW_PREFIX:-/home/linuxbrew}"
  HOMEBREW_PREFIX_DEFAULT="${HOMEBREW_PREFIX}/.linuxbrew"
```

Then the user has to know to set the prefix to something that includes a subdirectory (or else the contents of .linuxbrew either gets put into a hidden directory in the prefix, or if left out entirely, then its subdirs get put directly into prefix. I feel like people who are used to configuring options for installs understand that a prefix is the top-level system dir, and so expect the application to have its own separate top-level dir.

Providing an env var override makes it possible to override the default, but I'd be interested in writing some ``--option`` parsing to make this more discoverable. Just not sure whether you want options that may not apply to Mac installs.

MikeMcQuaid:
> HOMEBREW_PREFIX_DEFAULT allows the default for the Hombrew-specific directory structure to be controlled by the Homebrew project. I don't

Looks like you missed something here.

> then its subdirs get put directly into prefix

Yep, this is intentional/desired. The `.linuxbrew` stuff exists only for the full path to match those in bottles.

> Providing an env var override makes it possible to override the default, but I'd be interested in writing some `--option` parsing to make this more discoverable. Just not sure whether you want options that may not apply to Mac installs.

I'm not sure setting an environment variable that changes the behaviour of the installer script without prompting is a good idea. Having a `--prefix=...` or similar would work but I think the argument parsing will be a pain.

I'm not sure there's the (widespread) desire to programmatically have the Linux installer install to a non-default location?

PR #287, opened 2020-03-15
✨ support set brew_repo by environment variable

stormslowly:
in china due to GFW, cloning or fetching  from github directly is not stable.
usually, we ues a mirror of brew repo.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -30,14 +30,8 @@ else
   CHGRP="/bin/chgrp"
   GROUP="$(id -gn)"
 fi
-DEFAULT_BREW_REPO="https://github.com/Homebrew/brew"
 
-BREW_REPO="$DEFAULT_BREW_REPO"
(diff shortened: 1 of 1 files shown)


stormslowly:
in this commit, Chinese user can install as 

```
BREW_REPO=https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/brew.git HOMEBREW_CORE_GIT_REMOTE=https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/homebrew-core.git  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
```
not only make the installation success，but also fast！

MikeMcQuaid:
For initial installation this script should continue to use the default URL (to avoid setting a random environment variable on the system changing the Homebrew installation behaviour).

PR #288, opened 2020-03-15
Fixed compatibility with Ruby 2.0 (macOS <= 10.12)

Bo98:
Fixes Homebrew/brew#7180.


MikeMcQuaid:
Good call, thanks @Bo98!

PR #290, opened 2020-03-20
Fix terminal device

maxim-belkin:
Fix bug introduced in b2dd833c.
I also suggest using `stty -g` to save terminal state and then restore to it.

CC @sjackman


MikeMcQuaid:
Good catch, thanks @maxim-belkin!

maxim-belkin:
Thanks Mike!

PR #294, opened 2020-04-14
Fix compatibility with Ruby 2.0 for the uninstaller

itchyny:
This is uninstaller counterpart of #288. I believe the consistency matters here too.


MikeMcQuaid:
Thanks @itchyny!

PR #296, opened 2020-04-14
Fix unbound variable error on quick abort

itchyny:
I see `/bin/bash: line 5: $1: unbound variable` error on aborting the uninstaller quickly.


MikeMcQuaid:
can this be fixed in `abort()`?

itchyny:
Sounds reasonable.

MikeMcQuaid:
Thanks again @itchyny!

PR #297, opened 2020-04-14
Remove Ruby tools

itchyny:
Now that we only have 20 LOC of Ruby scripts and they are unlikely to grow, we can remove the Ruby tools and setup in the CI.


MikeMcQuaid:
Great work, thanks again @itchyny!

PR #301, opened 2020-04-15
Pass down cmd arguments to bash, fixes #299

### LINKED ISSUE #299: ruby version of `uninstall --force` now freezes

The travis log linked below runs on 10.13.6 (as can be seen in the log file)

Our CI is using the recommended procedure to automatically uninstall a preinstalled homebrew:

```
curl -fsSOL https://raw.githubusercontent.com/Homebrew/install/master/uninstall
sudo ruby uninstall --force
```

For example: https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/r.rb#L568-L576

As of yesterday running this script in CI freezes at the user prompt:

```
Warning: This script will remove:
/Users/travis/Library/Caches/Homebrew/
/Users/travis/Library/Logs/Homebrew/
/usr/local/Caskroom/
/usr/local/Cellar/
/usr/local/bin/brew -> /usr/local/bin/brew
Are you sure you want to uninstall Homebrew? This will remove your installed packages! [y/N] 
```

Example: https://travis-ci.org/github/ropensci/xslt/jobs/675283572

From the commits I infer that the uninstall script was ported to bash (which is great) but perhaps it would be possible to keep the ruby version working as well for existing deployments?

Uninstall homebrew without prompting user confirmation.

jeroen:
This should pass down the ruby command line parameters to bash. The `man bash` docs say:

```
  -c string If the -c option is present, then commands are read from string. If
     there are arguments after the string, they are assigned to the positional 
     parameters, starting with $0.
```

So therefore we first add parameter `$0` (which is normally then filename of the script you are invoking) and call it `uninstall` (it can ban be anything). And after that we append the command line arguments in the same order as we got them.

You can test it like so:

```sh
ruby uninstall --help
```

Or so:

```sh
ruby uninstall --dry-run --force
```

### DIFF
--- a/uninstall
+++ b/uninstall
@@ -7,4 +7,4 @@ Bash. Please migrate to the following command:
 
 EOS
 
-Kernel.exec "/bin/bash", "-c", '/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall.sh)"'
+Kernel.exec "/bin/bash", "-c", '/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/uninstall.sh)"' + ' uninstall ' +  ARGV.join(" ")


MikeMcQuaid:
Thanks @jeroen!

PR #308, opened 2020-05-01
install.sh: check Glibc version

maxim-belkin:
Don't attempt to install on Linux systems with Glibc older than 2.13.
😢

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -219,7 +219,7 @@ old_glibc() {
 if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && old_ruby && old_glibc; then
     abort "$(cat <<EOFABORT
 Homebrew requires Ruby 2.6 which was not found on your system. 
-Homebrew can use vendored Ruby but it requires Glibc version 2.13 or newer.
+Homebrew's portable Ruby requires Glibc version 2.13 or newer
 Your Glibc version is too old.
 See ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#linuxwsl-requirements${tty_reset}
 Consider installing Ruby 2.6 and poiting Homebrew to its executable with:
--- a/install.sh
+++ b/install.sh
@@ -238,7 +238,7 @@ if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && no_usable_ruby && outdated_glibc
 then
     abort "$(cat <<-EOFABORT
 	Homebrew requires Ruby $REQUIRED_RUBY_VERSION which was not found on your system.
-	Homebrew's portable Ruby requires Glibc version $REQUIRED_GLIBC_VERSION or newer,
+	Homebrew portable Ruby requires Glibc version $REQUIRED_GLIBC_VERSION or newer,
 	and your Glibc version is too old.
 	See ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#requirements${tty_reset}
 	Install Ruby $REQUIRED_RUBY_VERSION and add its location to your PATH.
--- a/install.sh
+++ b/install.sh
@@ -39,6 +39,10 @@ MACOS_LATEST_SUPPORTED="10.15"
 # TODO: bump version when new macOS is released
 MACOS_OLDEST_SUPPORTED="10.13"
 
+# For Homebrew on Linux
+required_ruby_version=2.6
+required_glibc_version=2.13
+
 # no analytics during installation
 export HOMEBREW_NO_ANALYTICS_THIS_RUN=1
 export HOMEBREW_NO_ANALYTICS_MESSAGE_OUTPUT=1
@@ -200,32 +204,36 @@ file_not_grpowned() {
   [[ " $(id -G "$USER") " != *" $(get_group "$1") "*  ]]
 }
 
-old_ruby() {
-  if builtin command -v ruby &>/dev/null; then
-    local ruby_version
-    ruby_version=$(ruby -e 'puts RUBY_VERSION' | grep -o '^[0-9]\+\.[0-9]\+')
-    version_lt "$ruby_version" 2.6
-  else
-    return 0
-  fi
+# Please sync with 'test-ruby()' in 'Library/Homebrew/utils/ruby.sh' from Homebrew/brew repository.
+test-ruby () {
+  [[ ! -x $1 ]] && { echo "false"; return 1; }
+  "$1" --enable-frozen-string-literal --disable=gems,did_you_mean,rubyopt -rrubygems -e \
+    "puts Gem::Version.new(RUBY_VERSION.to_s.dup).to_s.split('.').first(2) == \
+          Gem::Version.new('$required_ruby_version').to_s.split('.').first(2)" 2>/dev/null
 }
 
-old_glibc() {
+usable_ruby() {
+  local ruby_exec
+  ruby_exec=$(which ruby)
+  test "$(test-ruby "$ruby_exec")" == true
+}
+
+usable_glibc() {
   local glibc_version
   glibc_version=$(ldd --version | head -n1 | grep -o '[0-9.]*$' | grep -o '^[0-9]\+\.[0-9]\+')
-  version_lt "$glibc_version" 2.13
+  version_ge "$glibc_version" $required_glibc_version
 }
 
-if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && old_ruby && old_glibc; then
-    abort "$(cat <<EOFABORT
-Homebrew requires Ruby 2.6 which was not found on your system. 
-Homebrew's portable Ruby requires Glibc version 2.13 or newer
-Your Glibc version is too old.
-See ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#linuxwsl-requirements${tty_reset}
-Consider installing Ruby 2.6 and poiting Homebrew to its executable with:
-    export HOMEBREW_RUBY_PATH=/path/to/ruby
-EOFABORT
-)"
+if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && ! usable_ruby && ! usable_glibc
+then
+    abort "$(cat <<-EOFABORT
+	Homebrew requires Ruby $required_ruby_version which was not found on your system.
+	Homebrew's portable Ruby requires Glibc version $required_glibc_version or newer
+	and your Glibc version is too old.
+	See ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#linuxwsl-requirements${tty_reset}
+	Consider installing Ruby $required_ruby_version and adding its location to PATH.
+	EOFABORT
+    )"
 fi
 
 # USER isn't always set so provide a fall back for the installer and subprocesses.
--- a/install.sh
+++ b/install.sh
@@ -40,8 +40,8 @@ MACOS_LATEST_SUPPORTED="10.15"
 MACOS_OLDEST_SUPPORTED="10.13"
 
 # For Homebrew on Linux
-required_ruby_version=2.6
-required_glibc_version=2.13
+REQUIRED_RUBY_VERSION=2.6  # https://github.com/Homebrew/brew/pull/6556
+REQUIRED_GLIBC_VERSION=2.13  # https://docs.brew.sh/Homebrew-on-Linux#requirements
 
 # no analytics during installation
 export HOMEBREW_NO_ANALYTICS_THIS_RUN=1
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -200,6 +200,30 @@ file_not_grpowned() {
   [[ " $(id -G "$USER") " != *" $(get_group "$1") "*  ]]
 }
 
+glibc_version_check() {
+  local glibc_version dots ndots
+  glibc_version=$(ldd --version | grep '^ldd' | grep -o '\([0-9.]*\)$')
```
sjackman, line 205:
```suggestion
  local glibc_version=$(ldd --version | head -n1 | grep -o '[0-9.]*$' | grep -o '^[0-9]*\.[0-9]*')
```

install.sh
```diff
@@ -200,6 +200,30 @@ file_not_grpowned() {
...
+  dots=${glibc_version//[^.]}
+  ndots=${#dots}
+  while (( ndots > 1 )); do
+    glibc_version=${glibc_version%.*}
+    dots="${glibc_version//[^.]}"
+    ndots=${#dots}
+  done
```
sjackman, line 212: Remove this loop.

install.sh
```diff
@@ -200,6 +200,30 @@ file_not_grpowned() {
...
+  if version_lt "$glibc_version" 2.13; then
+    abort "$(cat <<EOFABORT
+Your Glibc version ($glibc_version) is too old.
```
sjackman, line 215:
```suggestion
Your Glibc version $glibc_version is too old.
```

install.sh
```diff
@@ -200,6 +200,30 @@ file_not_grpowned() {
...
+Minimum supported version is 2.13.
```
sjackman, line 216:
```suggestion
Glibc version 2.13 or newer is required.
```

install.sh
```diff
@@ -200,6 +200,30 @@ file_not_grpowned() {
...
+See: ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#linuxwsl-requirements${tty_reset}
```
sjackman, line 217:
```suggestion
See ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#linuxwsl-requirements${tty_reset}
```

install.sh
```diff
@@ -200,6 +200,30 @@ file_not_grpowned() {
   [[ " $(id -G "$USER") " != *" $(get_group "$1") "*  ]]
 }
 
+glibc_version_check() {
+  local glibc_version dots ndots
```
sjackman, line 204:
```suggestion
  local glibc_version
```

sjackman:
Thanks for this PR, Maxim!

MikeMcQuaid:
> glibc 2.13 is a dependency of `portable-ruby`. This check should be made conditional on `ruby` not being available in `PATH`.

If that's the case I think this logic should be in Homebrew/brew too.

maxim-belkin:
> glibc 2.13 is a dependency of `portable-ruby`. This check should be made conditional on `ruby` not being available in `PATH`.

Done!

install.sh
```diff
@@ -200,6 +200,34 @@ file_not_grpowned() {
   [[ " $(id -G "$USER") " != *" $(get_group "$1") "*  ]]
 }
 
+old_ruby() {
+  if builtin command -v ruby &>/dev/null; then
+    local ruby_version
+    ruby_version=$(ruby <<< 'puts RUBY_VERSION' | grep -o '^[0-9]\+\.[0-9]\+')
```
maxim-belkin, line 206: Changed  `^[0-9]*\.[0-9]*` to `^[0-9]\+\.[0-9]\+` to disallow zero-length matches.

install.sh
```diff
@@ -200,6 +200,34 @@ file_not_grpowned() {
   [[ " $(id -G "$USER") " != *" $(get_group "$1") "*  ]]
 }
 
+old_ruby() {
+  if builtin command -v ruby &>/dev/null; then
+    local ruby_version
+    ruby_version=$(ruby -e 'puts RUBY_VERSION' | grep -o '^[0-9]\+\.[0-9]\+')
```
MikeMcQuaid, line 206: Use logic like this: https://github.com/Homebrew/brew/blob/53fe2c94dbfecf78192fd05b7c3adc245e67f8d6/Library/Homebrew/utils/ruby.sh#L56

install.sh
```diff
@@ -200,6 +200,34 @@ file_not_grpowned() {
...
+}
+
+if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && old_ruby && old_glibc; then
+    abort "$(cat <<EOFABORT
+Homebrew requires Ruby 2.6 which was not found on your system. 
+Homebrew can use vendored Ruby but it requires Glibc version 2.13 or newer.
+Your Glibc version is too old.
+See ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#linuxwsl-requirements${tty_reset}
+Consider installing Ruby 2.6 and poiting Homebrew to its executable with:
+    export HOMEBREW_RUBY_PATH=/path/to/ruby
```
MikeMcQuaid, line 226: This isn't necessary/right. Please use similar messaging to this: https://github.com/Homebrew/brew/blob/53fe2c94dbfecf78192fd05b7c3adc245e67f8d6/Library/Homebrew/utils/ruby.sh#L69-L73

install.sh
maxim-belkin: What advantage does that approach have?

MikeMcQuaid: Consistency is the main advantage. Otherwise there are cases where the version in this PR will fail and the version in Homebrew/brew will not (or vice-versa).

maxim-belkin:
> This issue has been automatically marked as stale because it has not had recent activity. It will be closed if no further activity occurs.

Thanks for the (automatic) reminder! I'll get back to this soon.

maxim-belkin:
@MikeMcQuaid @sjackman this is ready for a second round of reviews.

install.sh
```diff
@@ -219,7 +219,7 @@ old_glibc() {
 if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && old_ruby && old_glibc; then
     abort "$(cat <<EOFABORT
 Homebrew requires Ruby 2.6 which was not found on your system. 
-Homebrew can use vendored Ruby but it requires Glibc version 2.13 or newer.
+Homebrew's portable Ruby requires Glibc version 2.13 or newer
```
sjackman:
```suggestion
Homebrew's portable Ruby requires Glibc version 2.13 or newer.
```

install.sh
```diff
@@ -200,6 +204,38 @@ file_not_grpowned() {
   [[ " $(id -G "$USER") " != *" $(get_group "$1") "*  ]]
 }
 
+# Please sync with 'test-ruby()' in 'Library/Homebrew/utils/ruby.sh' from Homebrew/brew repository.
+test-ruby () {
+  [[ ! -x $1 ]] && { echo "false"; return 1; }
```
sjackman, line 209:
```suggestion
  if [[ ! -x "$1" ]]; then 
    return 1
  fi
```

install.sh
```diff
@@ -200,6 +204,38 @@ file_not_grpowned() {
...
+  "$1" --enable-frozen-string-literal --disable=gems,did_you_mean,rubyopt -rrubygems -e \
+    "puts Gem::Version.new(RUBY_VERSION.to_s.dup).to_s.split('.').first(2) == \
+          Gem::Version.new('$required_ruby_version').to_s.split('.').first(2)" 2>/dev/null
```
sjackman, line 212: Capture the output and compare its output to the word `true` and return `0` if equal and `1` otherwise.

install.sh
```diff
@@ -200,6 +204,38 @@ file_not_grpowned() {
...
+}
+
+usable_ruby() {
+  local ruby_exec
+  ruby_exec=$(which ruby)
+  test "$(test-ruby "$ruby_exec")" == true
```
sjackman, line 218:
```suggestion
  test-ruby "$(command -v ruby)"
```

install.sh
```diff
@@ -200,6 +204,38 @@ file_not_grpowned() {
...
+}
+
+usable_glibc() {
+  local glibc_version
+  glibc_version=$(ldd --version | head -n1 | grep -o '[0-9.]*$' | grep -o '^[0-9]\+\.[0-9]\+')
+  version_ge "$glibc_version" $required_glibc_version
```
sjackman, line 224:
```suggestion
  version_ge "$glibc_version" "$required_glibc_version"
```
Quotes for consistency and style.

install.sh
```diff
@@ -200,6 +204,38 @@ file_not_grpowned() {
...
+}
+
+if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && ! usable_ruby && ! usable_glibc
+then
+    abort "$(cat <<-EOFABORT
+	Homebrew requires Ruby $required_ruby_version which was not found on your system.
+	Homebrew's portable Ruby requires Glibc version $required_glibc_version or newer
```
sjackman, line 231:
```suggestion
	Homebrew's portable Ruby requires Glibc version $required_glibc_version or newer,
```

install.sh
sjackman: I suggest merging `usable_ruby` and `test_ruby` into a single function. I forgot now… is it supposed to iterate through `PATH` and find the first usable Ruby anywhere in `PATH`? In case say Ruby 2.7 is at the head of `PATH`, but Ruby 2.6 is in there somewhere deeper.

install.sh
```diff
@@ -200,6 +204,38 @@ file_not_grpowned() {
   [[ " $(id -G "$USER") " != *" $(get_group "$1") "*  ]]
 }
 
+# Please sync with 'test-ruby()' in 'Library/Homebrew/utils/ruby.sh' from Homebrew/brew repository.
+test-ruby () {
```
sjackman, line 208:
```suggestion
test_ruby () {
```
I believe our convention is to use `_` rather than `-` in function names.

install.sh
maxim-belkin: I'm 👍 for this change (except that there is no need for `"` around `$1`) but I'd like to keep this function identical to its sibling in `utils/ruby.sh` in Homebrew/brew.

install.sh
maxim-belkin: Same as above -- I actually implemented what you suggested here but decided that keeping it identical to its sibling in Homebrew/brew is more important.

install.sh
maxim-belkin: `which ruby` finds the first `ruby` in user's PATH. I can change it to mimick what we do in `utils/ruby.sh` to check all Ruby versions.

install.sh
maxim-belkin: Quotes around `required_glibc_version` are not necessary because we set its value ourselves, it does not have spaces, and this is not expected to change. 
What style do you mean?

install.sh
maxim-belkin: I copied this function directly from `Library/Homebrew/utils/ruby.sh`. It's better to keep these two functions identical so that `install.sh` does exactly the same check as `ruby.sh`

install.sh
maxim-belkin: I'm 👍  for this change if we also change it in `ruby.sh` in Homebrew/brew.

install.sh
sjackman: > except that there is no need for " around $1

```console
$ function cat_file() { cat $1; }
$ cat_file "foo bar"
cat: foo: No such file or directory
cat: bar: No such file or directory
$ function cat_file() { cat "$1"; }
$ cat_file "foo bar"
Hello, world!
```

I believe the quotes are necessary.

sjackman: > I'd like to keep this function identical to its sibling in utils/ruby.sh in Homebrew/brew.

Want to change them both then?

install.sh
sjackman: Let's change both!

install.sh
sjackman: Yeah, that's probably necessary.

install.sh
sjackman: All `$foo` should be surrounded by quotes `"$foo"`. Each time I see `$foo` I have to ask myself, Is this a bug? 99% of the time the answer is yes.

sjackman: If I see `$required_glibc_version` I have to scroll up to the top of the file to find the definition of `required_glibc_version` to determine whether it's a bug. If I see `"$required_glibc_version"` I can happily continue on my way.

install.sh
maxim-belkin: > Want to change them both then?

Sure!



> > except that there is no need for " around $1
> 
> ```
> $ function cat_file() { cat $1; }
> $ cat_file "foo bar"
> cat: foo: No such file or directory
> cat: bar: No such file or directory
> $ function cat_file() { cat "$1"; }
> $ cat_file "foo bar"
> Hello, world!
> ```
> 
> I believe the quotes are necessary.

Not inside `[[ ]]`:
```sh
cd $(mktemp -d)
mkdir "dir with spaces"
FILE="dir with spaces/test.sh"
echo "echo Hi" > "$FILE"
chmod u+x "$FILE"
if [[ -x $FILE ]]; then
    echo success
fi
```

The reason is that (from `man bash`):

> Word  splitting  and  pathname  expansion  are  not performed on the words between the [[ and ]];

sjackman: All the exceptions in bash are crazy making. 🤪 😠

install.sh
```diff
@@ -200,6 +204,51 @@ file_not_grpowned() {
...
+  version_ge "$glibc_version" "$required_glibc_version"
+}
+
+if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && ! has_usable_ruby && ! usable_glibc
+then
+    abort "$(cat <<-EOFABORT
+	Homebrew requires Ruby $required_ruby_version which was not found on your system.
+	Homebrew's portable Ruby requires Glibc version $required_glibc_version or newer,
+	and your Glibc version is too old.
+	See ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#linuxwsl-requirements${tty_reset}
```
MikeMcQuaid, line 246:
```suggestion
	See ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#requirements${tty_reset}
```

install.sh
```diff
@@ -200,6 +204,51 @@ file_not_grpowned() {
...
+	Consider installing Ruby $required_ruby_version and adding its location to PATH.
```
MikeMcQuaid, line 247:
```suggestion
	Install Ruby $required_ruby_version and add its location to your PATH.
```

install.sh
```diff
@@ -39,6 +39,10 @@ MACOS_LATEST_SUPPORTED="10.15"
 # TODO: bump version when new macOS is released
 MACOS_OLDEST_SUPPORTED="10.13"
 
+# For Homebrew on Linux
+required_ruby_version=2.6
+required_glibc_version=2.13
```
MikeMcQuaid, line 44: Uppercase these and use `"` to match style of variables above. Also: might be worth linking to the document/code that defines these requirements.

install.sh
maxim-belkin: Suggestion implemented. Resolving the conversation to cleanup the page.

install.sh
maxim-belkin: Message updated in 761430e575b0ddf83fdbe2e3d024c95d565339ee.

maxim-belkin:
Had to change 
> Homebrew's portable Ruby

to
> Homebrew portable Ruby

because it was making Bash 3 unhappy (tested with `/bin/bash -e install.sh`)

Other than quotes around 2.6 and 2.13 I made all requested changes. I'm not a big fan of adding `"` and `{}` where they are not necessary because in Bash by default everything is a string and quotation marks bear different meaning than in programming languages with type systems. But I can do so if you feel strongly about it -- let me know.

maxim-belkin:
Idea: we can move added code to `check_requirements_linux.sh`.

maxim-belkin:
> Idea: we can move added code to `check_requirements_linux.sh`.

Silly me.... we can't because of the way we install Homebrew.

MikeMcQuaid:
Great work as usual @maxim-belkin 👏

PR #315, opened 2020-06-02
Run as a standard user

### LINKED ISSUE #312: Unable to install as a standard user with a permissive sudo policy
Hello,

Install Homebrew on macOS 10.15.4 as a standard user with sudo configured to enable the installation.

I have a custom sudo plugin which allows standard users to perform some privileged actions based on various file and process properties, in this case we allow sudo commands to run if the parent process is a verified Homebrew installation script.

You could enable a similar set up with a custom sudoers configuration, although it would be difficult to get right and far too permissive.

Many companies enforce "least privilege" solutions company wide, but they would also like to allow their developers and power users to use the latest Homebrew installer.

<pre>
standard@mac ~ % ./install.sh                                                                      
This script requires the user test to be an Administrator.</pre>

The installation script shouldn't check if the user is a member of the admin group, because they might have sufficient privileges as a standard user. If the sudo command fails it can abort with a useful error message. This is what happens if you simply remove the group membership check from the script.

<pre>
...
==> The following existing directories will have their group set to admin:
/usr/local/bin
/usr/local/etc
...

Password: ...
Need sudo access on macOS!
</pre>

And in the case where sudo is permissive, Homebrew will install successfully.

Any custom sudoers configuration I could write would be too permissive to be used in practice, but would work for testing purposes. I'll happily create one if the above descriptions aren't sufficient.

Thanks!

lakshaykaushik:
Allows the install script to be run when the user is a standard user and not part of the admin group.

Fixes #312

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -72,7 +72,7 @@ have_sudo_access() {
   fi
 
   if [[ -z "${HOMEBREW_ON_LINUX-}" ]] && [[ "$HAVE_SUDO_ACCESS" -ne 0 ]]; then
-    abort "Run the script with a user which has sufficient permissions!"
+    abort "Need sudo access on macOS (e.g. the user $USER to be an Administrator)!"
   fi
 
   return "$HAVE_SUDO_ACCESS"
--- a/install.sh
+++ b/install.sh
@@ -233,14 +233,16 @@ EOABORT
 )"
 fi
 
-if [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
+if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
+ if have_sudo_access; then :
+ fi
+elif [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
   if [[ -n "${CI-}" ]] || [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] || [[ -w "/home/linuxbrew" ]] || [[ -w "/home" ]]; then
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
   else
     trap exit SIGINT
-    sudo_output="$(/usr/bin/sudo -n -l mkdir 2>&1)"
-    sudo_exit_code="$?"
-    if [[ "$sudo_exit_code" -ne 0 ]] && [[ "$sudo_output" = "sudo: a password is required" ]]; then
+    if [[ `/usr/bin/sudo -n -l mkdir 2>&1` == *"mkdir"* ]]; then :
+    else
       ohai "Select the Homebrew installation directory"
       echo "- ${tty_bold}Enter your password${tty_reset} to install to ${tty_underline}${HOMEBREW_PREFIX_DEFAULT}${tty_reset} (${tty_bold}recommended${tty_reset})"
       echo "- ${tty_bold}Press Control-D${tty_reset} to install to ${tty_underline}$HOME/.linuxbrew${tty_reset}"
@@ -255,6 +257,7 @@ if [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
   fi
   HOMEBREW_REPOSITORY="${HOMEBREW_PREFIX}/Homebrew"
 fi
+
 
 if [[ "$UID" == "0" ]]; then
   abort "Don't run this as root!"
@@ -277,7 +280,6 @@ EOABORT
 )"
   elif version_lt "$macos_version" "10.9"; then
     abort "Your OS X version is too old"
-  elif have_sudo_access; then :
   elif version_gt "$macos_version" "$MACOS_LATEST_SUPPORTED" || \
     version_lt "$macos_version" "$MACOS_OLDEST_SUPPORTED"; then
     who="We"
--- a/install.sh
+++ b/install.sh
@@ -234,9 +234,8 @@ EOABORT
 fi
 
 if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
- if have_sudo_access; then :
- fi
-elif [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
+ have_sudo_access
+else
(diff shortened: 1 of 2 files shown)


install.sh
```diff
@@ -72,7 +72,7 @@ have_sudo_access() {
   fi
 
   if [[ -z "${HOMEBREW_ON_LINUX-}" ]] && [[ "$HAVE_SUDO_ACCESS" -ne 0 ]]; then
-    abort "Need sudo access on macOS!"
+    abort "Run the script with a user which has sufficient permissions!"
```
MikeMcQuaid, line 75:
```suggestion
    abort "Need sudo access on macOS (e.g. the user $USER to be an Administrator)!"
```

install.sh
```diff
@@ -277,8 +277,7 @@ EOABORT
 )"
   elif version_lt "$macos_version" "10.9"; then
     abort "Your OS X version is too old"
-  elif ! [[ "$(dsmemberutil checkmembership -U "$USER" -G "$GROUP")" = *"user is a member"* ]]; then
-    abort "This script requires the user $USER to be an Administrator."
+  elif have_sudo_access; then :
```
MikeMcQuaid, line 280: Feels a bit weird being in here. What about calling `have_sudo_access` at line 305 (just above "this script will install")

uninstall.sh
```diff
@@ -87,7 +87,7 @@ have_sudo_access() {
   fi
 
   if [[ -z "${HOMEBREW_ON_LINUX-}" ]] && [[ "$HAVE_SUDO_ACCESS" -ne 0 ]]; then
-    abort "Need sudo access on macOS!"
+    abort "Run the script with a user which has sufficient permissions!"
```
MikeMcQuaid, line 90:
```suggestion
    abort "Need sudo access on macOS (e.g. the user $USER to be an Administrator)!"
```

install.sh
lakshaykaushik: Keeping it at 305 will make it run on linux machines too, line 241 is already checking sudo for linux machines. 
I think for macOS we need to check between line 271 and 304 which is within the if statement.
Am i missing something ?

lakshaykaushik: Typo: not line 241, line 249 i meant

MikeMcQuaid: Yeh, you're right, good catch. I think it would be good to unify the `sudo` checking as much as possible. Not sure why `sudo_output` and `sudo_exit_code` should be checked at all rather than just using `have_sudo_access` there? Would be nice to have one check location that's shared between macOS and Linux. Thoughts?

lakshaykaushik: Right, i think there is bug there as well, in case when SUDO_ASKPASS is set, the sudo checks would fail on line 241. 
I think we should definitely refactor the sudo checking, can we submit the refactoring as a separate PR? For this PR i am thinking about fixing the standard user run issue at line 280?
What do you think?

MikeMcQuaid: I reckon roll it all into this PR (if you're up for that)!

lakshaykaushik: Cool. Will refactor that as well

lakshaykaushik: Refactoring done. Actions pre and post sudo check are different for macOS and Linux, tried to bring them as close as they get without changing the behaviour of the script.
Tested this with linux and macOS, sudo(NOPASSWD and without NOPASSWD), non sudo users and users who are not admins but still have sudo rights. Works fine in all cases now.

install.sh
```diff
@@ -256,6 +258,7 @@ if [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
   HOMEBREW_REPOSITORY="${HOMEBREW_PREFIX}/Homebrew"
 fi
 
+
```
MikeMcQuaid, line 261:
```suggestion
```

install.sh
```diff
@@ -233,14 +233,16 @@ EOABORT
...
+elif [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
   if [[ -n "${CI-}" ]] || [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] || [[ -w "/home/linuxbrew" ]] || [[ -w "/home" ]]; then
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
   else
     trap exit SIGINT
-    sudo_output="$(/usr/bin/sudo -n -l mkdir 2>&1)"
-    sudo_exit_code="$?"
-    if [[ "$sudo_exit_code" -ne 0 ]] && [[ "$sudo_output" = "sudo: a password is required" ]]; then
+    if [[ $(/usr/bin/sudo -n -l mkdir 2>&1) == *"mkdir"* ]]; then :
+    else
```
MikeMcQuaid, line 245:
```suggestion
    if [[ $(/usr/bin/sudo -n -l mkdir 2>&1) != *"mkdir"* ]]; then
```

install.sh
```diff
@@ -233,14 +233,16 @@ EOABORT
 )"
 fi
 
-if [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
+if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
+ if have_sudo_access; then :
+ fi
+elif [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
```
MikeMcQuaid, line 239:
```suggestion
if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
 have_sudo_access
else
```

install.sh
lakshaykaushik: Tested it, works on my local.
Although whoever is running the script as non admin user(with sudo) needs to know that all directories within /usr/local will have their non admin user as owner of these directories. I did not change this as the script prompts the user before making this change.

If this looks fine to you, its ready to be merged.

MikeMcQuaid: Fine with me, great work 👏!

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @lakshaykaushik!

install.sh
steven-joruk: Thanks guys, sorry I haven't had any time to test, but the changes look good :).

PR #317, opened 2020-06-30
Do not install Command Line Tools on ARM Mac

djensenius:
When using Apple Transition Kit installing Command Line Tools is not needed. 

The command line tools are Intel based, and end up breaking things.

*Note*: Brew fails to run because of some Ruby issues. I will try to fix this in another PR.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -10,6 +10,7 @@ fi
 if [[ "$(uname -m)" = "arm64" ]] && [[ "$(uname)" = "Darwin" ]]; then
   HOMEBREW_APPLE_SILICON=1
 fi
+
 # On macOS, this script installs to /usr/local only.
 # On Linux, it installs to /home/linuxbrew/.linuxbrew if you have sudo access
 # and ~/.linuxbrew otherwise.
--- a/install.sh
+++ b/install.sh
@@ -10,8 +10,6 @@ fi
 if [[ "$(uname -m)" = "arm64" ]] && [[ "$(uname)" = "Darwin" ]]; then
   HOMEBREW_APPLE_SILICON=1
 fi
-
-
 # On macOS, this script installs to /usr/local only.
 # On Linux, it installs to /home/linuxbrew/.linuxbrew if you have sudo access
 # and ~/.linuxbrew otherwise.
--- a/install.sh
+++ b/install.sh
@@ -8,7 +8,7 @@ fi
 
 # Check if macOS is ARM
 if [[ "$(uname -m)" = "arm64" ]] && [[ "$(uname)" = "Darwin" ]]; then
-  HOMEBREW_MAC_ARM64=1
+  HOMEBREW_APPLE_SILICON=1
 fi
 
 
@@ -183,7 +183,7 @@ should_install_command_line_tools() {
     return 1
   fi
 
-  if [[ -n "${HOMEBREW_MAC_ARM64-}" ]]; then
+  if [[ -n "${HOMEBREW_APPLE_SILICON-}" ]]; then
     return 1;
   fi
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -6,6 +6,12 @@ if [[ "$(uname)" = "Linux" ]]; then
   HOMEBREW_ON_LINUX=1
 fi
 
+# Check if macOS is ARM
+if [[ "$(uname -m)" = "arm64" ]] && [[ "$(uname)" = "Darwin" ]]; then
+  HOMEBREW_MAC_ARM64=1
+fi
+
+
```
MikeMcQuaid, line 14:
```suggestion
# Check if macOS is ARM
if [[ "$(uname -m)" = "arm64" && "$(uname)" = "Darwin" ]]; then
  HOMEBREW_APPLE_SILICON=1
fi

```

install.sh
```diff
@@ -177,6 +183,10 @@ should_install_command_line_tools() {
     return 1
   fi
 
+  if [[ -n "${HOMEBREW_MAC_ARM64-}" ]]; then
```
MikeMcQuaid, line 186:
```suggestion

  # No CLT available for Apple Silicon (yet)
  if [[ -n "${HOMEBREW_APPLE_SILICON-}" ]]; then
```

djensenius:
Thanks @MikeMcQuaid! Good suggestion. Updated.

install.sh
```diff
@@ -6,6 +6,12 @@ if [[ "$(uname)" = "Linux" ]]; then
   HOMEBREW_ON_LINUX=1
 fi
 
+# Check if macOS is ARM
+if [[ "$(uname -m)" = "arm64" ]] && [[ "$(uname)" = "Darwin" ]]; then
+  HOMEBREW_APPLE_SILICON=1
+fi
+
+
```
MikeMcQuaid, line 14:
```suggestion

```

install.sh
```diff
@@ -6,6 +6,10 @@ if [[ "$(uname)" = "Linux" ]]; then
   HOMEBREW_ON_LINUX=1
 fi
 
+# Check if macOS is ARM
+if [[ "$(uname -m)" = "arm64" ]] && [[ "$(uname)" = "Darwin" ]]; then
+  HOMEBREW_APPLE_SILICON=1
+fi
```
MikeMcQuaid, line 12:
```suggestion
fi

```

One newline needed, not two, not none 😁

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @djensenius!

Bo98:
Just to be clear: this situation is likely temporary but a good fix to have in the meantime.

Bo98:
> Brew fails to run because of some Ruby issues. I will try to fix this in another PR.

You might be experiencing a bug with system Ruby. We've reported it as FB7823402.

djensenius:
For sure, my hope is it is only temporary! 😃

ttimpe:
Thank you.

tresf:
> The command line tools are Intel based, and end up breaking things.

For those of us that jumped the gun and now have accidentally installed the `x86_64` command line tools, is there a recommendation to revert them back?

Bo98:
`rm -rf /Library/Developer/CommandLineTools` should be sufficient.

tresf:
> rm -rf /Library/Developer/CommandLineTools should be sufficient.

The exact commands to fix this:

```bash
sudo rm -rf /Library/Developer/CommandLineTools
sudo xcode-select --switch /Applications/Xcode-beta.app
```

... and then rerun the brew install:

```bash
```

If Ruby is still broken, see the workaround here: https://github.com/Homebrew/brew/issues/7857#issue-647960270 and then repeat the above failing command.

PR #328, opened 2020-09-01
Add multi user support in macOS

GauthamGoli:
* Make `admin` group own directories related to brew

 * Grant read write execute permissions for `admin` group

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -191,6 +191,10 @@ exists_but_not_writable() {
   [[ -e "$1" ]] && ! [[ -r "$1" && -w "$1" && -x "$1" ]]
 }
 
+exists_but_not_group_writable() {
+  [[ -e "$1" ]] && [[ "$(get_permission "$1")" != "775" ]]
+}
+
 get_owner() {
   $STAT "%u" "$1"
 }
@@ -386,11 +390,20 @@ directories=(bin etc include lib sbin share opt var
              var/log var/homebrew var/homebrew/linked
              bin/brew)
 group_chmods=()
+
 for dir in "${directories[@]}"; do
   if exists_but_not_writable "${HOMEBREW_PREFIX}/${dir}"; then
     group_chmods+=("${HOMEBREW_PREFIX}/${dir}")
   fi
 done
+
+if [[ -n "${HOMEBREW_MULTI_USER-}" ]] && exists_but_not_group_writable "${HOMEBREW_PREFIX}/Homebrew"; then
+  group_chmods+=("${HOMEBREW_PREFIX}/Homebrew")
+fi
+
+if [[ -n "${HOMEBREW_MULTI_USER-}" ]]; then
+  directories+=(Homebrew)
+fi
 
 # zsh refuses to read from these directories if group writable
 directories=(share/zsh share/zsh/site-functions)
@@ -459,6 +472,10 @@ fi
 if [[ "${#mkdirs[@]}" -gt 0 ]]; then
   ohai "The following new directories will be created:"
   printf "%s\n" "${mkdirs[@]}"
+fi
+
+if [[ -z "${HOMEBREW_MULTI_USER-}" ]]; then
+  echo "Set HOMEBREW_MULTI_USER env var to allow ${tty_underline}${GROUP}${tty_reset} group users to install packages"
 fi
 
 if should_install_command_line_tools; then


install.sh
```diff
@@ -557,6 +557,17 @@ EOABORT
 )"
 fi
 
+# Multi-user support on macOS
+directories=(/usr/local /Library/Caches/Homebrew /opt/homebrew-cask)
```
MikeMcQuaid, line 561: Shouldn't need `/Library/Caches/Homebrew /opt/homebrew-cask` any more.

install.sh
```diff
@@ -557,6 +557,17 @@ EOABORT
...
+if [[ -z "${HOMEBREW_ON_LINUX-}" ]] && [[ -n "${HOMEBREW_ENABLE_MULTI_USER-}" ]]; then
+  for dir in "${directories[@]}"; do
+    if [[ -d "${dir}" ]]; then
+      execute_sudo "$CHGRP" "-R" "admin" ${dir}/*
```
MikeMcQuaid, line 565: Can't do chown/chgrp/chmod on `/usr/local` itself unfortunately.

MikeMcQuaid (commented):
Nice work so far! This should probably be part of this code: https://github.com/Homebrew/install/blob/771caa06f9e524717b429293d4622a89880ebe0f/install.sh#L373-L514

MichaelDeBoey:
@GauthamGoli any news on this one?

install.sh
```diff
@@ -543,6 +543,17 @@ EOABORT
 )"
 fi
 
+# Multi-user support on macOS
+directories=("$HOMEBREW_REPOSITORY" "$HOMEBREW_CACHE")
```
MikeMcQuaid, line 547: I don't think it makes sense to do this for `HOMEBREW_CACHE` because it's inside the `HOME` of a specific user (and will change depending on which user runs it)

install.sh
```diff
@@ -543,6 +543,17 @@ EOABORT
...
+if [[ -z "${HOMEBREW_ON_LINUX-}" ]] && [[ -n "${HOMEBREW_ENABLE_MULTI_USER-}" ]]; then
```
MikeMcQuaid, line 548: I think this should be done on Linux

install.sh
```diff
@@ -543,6 +543,17 @@ EOABORT
...
+  for dir in "${directories[@]}"; do
+    if [[ -d "${dir}" ]]; then
+      execute_sudo "$CHGRP" "-R" "admin" "${dir}"/*
+      execute_sudo "/bin/chmod" "-R" "g+rwx" "${dir}"/*
```
MikeMcQuaid, line 552: It'd be good to make this part of the `group_chmods` and `chgrps` logic above. I think this could be done by using the `file_not_grpowned` and perhaps `exists_but_not_writable` above. Similarly, on `The following existing directories will be made group writable` it'd be nice to e.g. advertise the `HOMEBREW_ENABLE_MULTI_USER` (perhaps renamed to `HOMEBREW_MULTI_USER`) variable and tell people they can Ctrl-C and rerun with it set in order to use as multiple `$GROUP` users.

MikeMcQuaid (commented):
Nice work!

install.sh
GauthamGoli: @MikeMcQuaid  A cleaner implementation, thanks! Updated the code accordingly

MichaelDeBoey:
Would love to see this one merged!

install.sh
sjackman: Linux doesn't unfortunately have a standard name for the `admin` group that I know of. Which group would we use? The primary group of the current user? Or we could create a `homebrew` group if one doesn't exist. That'd be the cleanest solution.

MikeMcQuaid: `wheel` is the closest thing to a standard here I think.

gromgit: I wish it were:
* `wheel` seems to be used mainly on BSD-derived distros
* POSIX is silent on the matter
* LSB (Linux Standard Base) says `adm` is optional
* Debian and Ubuntu split the difference between `adm` and `staff`

I'm inclined to agree with @sjackman that creating a `homebrew` group is probably the best way forward on _any_ platform.

MikeMcQuaid: I don't think `homebrew` grouping is a good way forward on macOS. You've convinced me it is on Linux, though.

jasl:
Friendly ping, looks CI green, anyone can push this forward? thanks!

sjackman:
@GauthamGoli What's the status on this PR?

install.sh
```diff
@@ -386,12 +390,21 @@ directories=(bin etc include lib sbin share opt var
...
 group_chmods=()
+
 for dir in "${directories[@]}"; do
   if exists_but_not_writable "${HOMEBREW_PREFIX}/${dir}"; then
     group_chmods+=("${HOMEBREW_PREFIX}/${dir}")
   fi
 done
 
+if [[ -n "${HOMEBREW_MULTI_USER-}" ]] && exists_but_not_group_writable "${HOMEBREW_PREFIX}/Homebrew"; then
+  group_chmods+=("${HOMEBREW_PREFIX}/Homebrew")
```
sjackman, line 401:
```suggestion
  group_chmods+=("${HOMEBREW_REPOSITORY}")
```
On native M1 on macOS, `HOMEBREW_PREFIX = HOMEBREW_REPOSITORY`.

install.sh
```diff
@@ -386,12 +390,21 @@ directories=(bin etc include lib sbin share opt var
...
+fi
+
+if [[ -n "${HOMEBREW_MULTI_USER-}" ]]; then
+  directories+=(Homebrew)
```
sjackman, line 405: How should this be changed when `HOMEBREW_PREFIX = HOMEBREW_REPOSITORY`?

gromgit: Since `$HOMEBREW_REPOSITORY` is always under the `$HOMEBREW_PREFIX` hierarchy:
```
if [[ -n "${HOMEBREW_MULTI_USER-}" && "${HOMEBREW_PREFIX}" != ${HOMEBREW_REPOSITORY} ]]; then
  directories+=("${HOMEBREW_REPOSITORY#${HOMEBREW_PREFIX}/}")
fi
```

gromgit (changes requested):
https://github.com/Homebrew/install/blob/8c05dab0e8e6e2e82246de589fcfdff98546902d/install.sh#L380-L382

Suggest adding:
```
# Force a sane umask for multi-user support
if [[ -n "${HOMEBREW_MULTI_USER-}" ]]; then
  umask 022
fi
```

sjackman:
The group sticky bit permission `2775` is very useful for directories.
See https://man7.org/linux/man-pages/man2/chown.2.html#NOTES
and https://linoxide.com/how-tos/stickbit-suid-guid/

>    Ownership of new files
       When a new file is created (by, for example, open(2) or
       mkdir(2)), its owner is made the same as the filesystem user ID
       of the creating process.  The group of the file depends on a
       range of factors, including the type of filesystem, the options
       used to mount the filesystem, and _whether or not the set-group-ID
       mode bit is enabled on the parent directory_.

install.sh
```diff
@@ -191,6 +191,10 @@ exists_but_not_writable() {
   [[ -e "$1" ]] && ! [[ -r "$1" && -w "$1" && -x "$1" ]]
 }
 
+exists_but_not_group_writable() {
+  [[ -e "$1" ]] && [[ "$(get_permission "$1")" != "775" ]]
```
sjackman, line 195:
```suggestion
  [[ -e "$1" ]] && [[ "$(get_permission "$1")" != "2775" ]]
```
See https://github.com/Homebrew/install/pull/328#issuecomment-758859007

sjackman: Note that this comment is intended for Linux. I just tested on macOS, and the _default_ behaviour on macOS is the same as `2775` directories on Linux! 🤯

sjackman: I don't know what bit `02000` (`S_ISGID`) does on Darwin.

jasl:
Friendly ping @GauthamGoli sorry to disturb you,
but do you still working on this?

install.sh
gromgit: > Note that this comment is intended for Linux. I just tested on macOS, and the _default_ behaviour on macOS is the same as `2775` directories on Linux!

That turns out to be [true on most (all?) BSDs](https://docstore.mik.ua/orelly/unix/upt/ch22_05.htm): SysV-based OSes tend to respect the effective GID of the creator, while BSDs tend to respect the parent dir's group.

As for macOS...

`mkdir(2)`:
> The directory's owner ID is set to the process's effective user ID.  The directory's group ID is set to that of the parent directory in which it is created.

`open(2)`:
> When a new file is created, it is given the group of the directory which contains it.

gromgit: > I don't know what bit `02000` (`S_ISGID`) does on Darwin.

Can't find any documentation on that either.

install.sh
```diff
@@ -461,6 +474,10 @@ if [[ "${#mkdirs[@]}" -gt 0 ]]; then
   printf "%s\n" "${mkdirs[@]}"
 fi
 
+if [[ -z "${HOMEBREW_MULTI_USER-}" ]]; then
+  echo "Set HOMEBREW_MULTI_USER env var to allow ${tty_underline}${GROUP}${tty_reset} group users to install packages"
+fi
```
diogoeichert, line 479: Could it be the default behavior instead, to avoid both the extra complexity in implementation and the extra step from the users? 🤔

MikeMcQuaid: Eventually: yes. Now: no.

install.sh
diogoeichert: It's probably there just for compatibility.

MichaelDeBoey:
I think this one needs to be kept open

MikeMcQuaid:
I disagree. Let's close this out until it's updated with the various comments and passes CI. Anyone else is welcome to pick this up and start a new PR.

PR #329, opened 2020-09-03
install.sh: use fewer functions

maxim-belkin:
@jonchang, unless I'm missing something, I think we can replace `should_install_curl` and `should_install_git` with `! command -v ... >/dev/null` without any loss of functionality.


MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @maxim-belkin!

PR #330, opened 2020-09-04
Quick n dirty patched install.sh MacOS11.0 on ARM

jnuyens:
The new generation of Macs run on ARM silicon instead of x86 architecture.
It is, however, possible to run brew and git on MacOS/ARM with the emulation mode.

This can help to prior to running brew and/or git commands:

alias brew='arch -arch x86_64 brew'
alias git='arch -arch x86_64 git'

A better autodetection of the architecture can be added as uname -p on this new kind of machines provides:
uname -p
arm64
uname -r
20.0.0


MikeMcQuaid:
Passing on this until it's not "quick and dirty" and passes CI 😁

PR #331, opened 2020-09-08
Support use custom BREW_REPO url

menduo:
Support use custom BREW_REPO url

So I can use mirrors like:

```console
BREW_REPO=https://mirrors.aliyun.com/homebrew/brew.git /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)" 
```


MikeMcQuaid:
You can do this after installation but we intentionally don't support this during installation.

PR #335, opened 2020-10-07
MacOS Big Sur and above install.sh

kenjiw360:
Pushed macos latest so it supports future versions


MikeMcQuaid:
We don't support Big Sur yet, sorry.

kenjiw360:
okay

kenjiw360:
actually, I changed it and I can do big sur

kenjiw360:
it's not buggy

PR #336, opened 2020-10-15
Use GITHUB_PATH instead of add-path

peaceiris:
https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/

> Action and workflow authors who are setting environment variables via stdout should update any usage of the set-env and add-path workflow commands to use the new environment files.


MikeMcQuaid:
Thanks so much for your first contribution and sponsorship! Without people like you submitting PRs and sponsoring us we couldn't run this project. You rock, @peaceiris!

PR #337, opened 2020-10-16
Update install.sh

melito00:
Need to restore IFS before return

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -224,6 +224,7 @@ no_usable_ruby() {
   IFS=$'\n' # Do word splitting on new lines only
   for ruby_exec in $(which -a ruby); do
     if test_ruby "$ruby_exec"; then
+      IFS=$' \t\n' # Restore IFS to its default value
       return 1
     fi
   done


MikeMcQuaid:
CC @maxim-belkin for thoughts

maxim-belkin:
We don't export IFS and it's changed inside of a function, so I'm pretty sure there is no need to unset it before the `return`.
Are you experiencing an issue / problem, @melito00?

melito00:
Yes, I'm not sure the reason but if IFS is not restored, $STAT was quoted and stat command failed. I cheched $STAT with echo command  some place in script then I found it.

maxim-belkin:
What STAT?

melito00:
https://github.com/Homebrew/install/blob/79ccb10560c855f407431bd52dd654cba3fc4fd9/install.sh#L29

maxim-belkin:
Homebrew/install is not related to this repository. What you observe is a problem with homebrew/install not homebrew/brew.￼ I can confirm this issue but the solution is to fix install.sh script￼￼. I think `stat --printf￼` is interpreted as a single command￼ so the solution is to use eval function on lines 183, 185, and 203 in that file.

reitermarkus:
We should refactor this according to:

https://github.com/koalaman/shellcheck/wiki/SC2013

gromgit:
@reitermarkus Agreed: #338

maxim-belkin:
You're correct, @melito00 -- we need to restore `IFS` because even though this is a global var, we can modify it from within a function in Bash. My bad. I think the fix you propose is the safest one 👍 (the other fix is to use `local IFS=$'\n'`).

@reitermarkus @gromgit, we want this piece to be analogous to what we do in Homebrew/brew -- see `find_ruby` function in `utils/ruby.sh`:
https://github.com/Homebrew/brew/blob/e630602ef6619df5b71f88443ab09a598982608e/Library/Homebrew/utils/ruby.sh#L14-L29
So, if you think that a `while` loop is better, then I'd suggest you first propose a change to `utils/ruby.sh` and then we can merge it here.

maxim-belkin (approved):
👍

MikeMcQuaid:
https://github.com/Homebrew/install/pull/338 made more sense to me 👍🏻

PR #338, opened 2020-10-18
Improved no_usable_ruby() test

gromgit:
This version doesn't mess with IFS, so it's probably more robust, and definitely easier to understand.


MikeMcQuaid (approved):
Looks good, thanks! Could you check if there's similar updates that can be done in Homebrew/brew?

PR #340, opened 2020-10-23
If Xcode installed, no need install CommandLineTools

cntrump:
If Xcode installed, git also installed to `/usr/local/bin`, CommandLineTools is not necessary.


MikeMcQuaid:
It is necessary.

PR #341, opened 2020-10-24
Use sudo -v instead of sudo -l mkdir

hwalinga:
Install script does not prompt for password for user to make choice between root and non-root install. 

If you install homebrew you can at some point hit the question 

```
Select the Homebrew installation directory
Enter your password to install to..
Press Control-D to install to ...". 
```

However, the check to prompt this is `sudo -l mkdir`. The `-l` flag is according to the sudo manpage [1] a check if the user is *allowed* to run command mkdir with sudo rights. Either the user can do this, in which case the prompt is skipped. Or the user cannot do this, in which case the prompt is shown, but useless, since the user would not be able to give useful credentials anyway.

Instead you should be using the flag `-v` which is for checking credentials and *prompting when necessary*. You can use `sudo -n -v` to check if the user *at the moment* can execute sudo commands without interaction (`-n` makes this non-interactive.) And using the flag `-v` without `-n` to check for user credentials and *prompting when necessary*. In which case the prompt make sense and can be correctly skipped with CTRL-D (I have that tested). 

TLDR; `sudo -l mkdir` does not prompt for password. `sudo -v` will prompt for password. 

[1] https://linux.die.net/man/8/sudo

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -68,9 +68,9 @@ have_sudo_access() {
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null 
-  else
-      /usr/bin/sudo -v && /usr/bin/sudo -l mkdir &>/dev/null 
+      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
+    else
+      /usr/bin/sudo -v && /usr/bin/sudo -l mkdir &>/dev/null
     fi
     HAVE_SUDO_ACCESS="$?"
   fi
@@ -289,7 +289,6 @@ else
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
   else
     trap exit SIGINT
-    # If allowed to run sudo for mkdir and required to provide sudo password: prompt:
     if ! /usr/bin/sudo -n -v &>/dev/null; then
       ohai "Select the Homebrew installation directory"
       echo "- ${tty_bold}Enter your password${tty_reset} to install to ${tty_underline}${HOMEBREW_PREFIX_DEFAULT}${tty_reset} (${tty_bold}recommended${tty_reset})"
--- a/install.sh
+++ b/install.sh
@@ -68,9 +68,9 @@ have_sudo_access() {
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
+      /usr/bin/sudo "${args[@]}" -v &>/dev/null
     else
-      /usr/bin/sudo -l mkdir &>/dev/null
+      /usr/bin/sudo -v &>/dev/null
     fi
     HAVE_SUDO_ACCESS="$?"
   fi
@@ -286,7 +286,7 @@ else
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
   else
     trap exit SIGINT
-    if [[ $(/usr/bin/sudo -n -l mkdir 2>&1) != *"mkdir"* ]]; then
+    if ! /usr/bin/sudo -n -v &>/dev/null; then
       ohai "Select the Homebrew installation directory"
       echo "- ${tty_bold}Enter your password${tty_reset} to install to ${tty_underline}${HOMEBREW_PREFIX_DEFAULT}${tty_reset} (${tty_bold}recommended${tty_reset})"
       echo "- ${tty_bold}Press Control-D${tty_reset} to install to ${tty_underline}$HOME/.linuxbrew${tty_reset}"
--- a/install.sh
+++ b/install.sh
@@ -67,26 +67,8 @@ have_sudo_access() {
   fi
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
-
-    if [[ -n "${args[*]-}" ]]; then
-      SUDO="/usr/bin/sudo ${args[*]}"
-    else
-      SUDO="/usr/bin/sudo"
-    fi
-
-    if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
-      ${SUDO} -l mkdir
-    else
-      # We ask for password on Linux with sudo -v
-      # to give options as shown in the
-      # 'Select the Homebrew installation folder'
-      # But first we check if the user is allowed to run mkdir with sudo.
-      # (It is possible that the user is only allowed
-      # to run a subset of commands with sudo, but not mkdir.)
-      ${SUDO} -l mkdir &>/dev/null && ${SUDO} -v
-    fi
+    /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null 
     HAVE_SUDO_ACCESS="$?"
-
   fi
 
   if [[ -z "${HOMEBREW_ON_LINUX-}" ]] && [[ "$HAVE_SUDO_ACCESS" -ne 0 ]]; then
@@ -304,7 +286,7 @@ else
   else
     trap exit SIGINT
     # If allowed to run sudo for mkdir and required to provide sudo password: prompt:
-    if /usr/bin/sudo -l mkdir &>/dev/null && ! /usr/bin/sudo -n -v &>/dev/null; then
+    if ! /usr/bin/sudo -n -v &>/dev/null; then
       ohai "Select the Homebrew installation directory"
       echo "- ${tty_bold}Enter your password${tty_reset} to install to ${tty_underline}${HOMEBREW_PREFIX_DEFAULT}${tty_reset} (${tty_bold}recommended${tty_reset})"
       echo "- ${tty_bold}Press Control-D${tty_reset} to install to ${tty_underline}$HOME/.linuxbrew${tty_reset}"
--- a/install.sh
+++ b/install.sh
@@ -66,19 +66,27 @@ have_sudo_access() {
     args=("-A")
   fi
 
-  if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
-      SUDO_TEST="-l mkdir"
-  else
-      SUDO_TEST="-v"
-  fi
-
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
+
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" "${SUDO_TEST}" &>/dev/null
+      SUDO="/usr/bin/sudo ${args[*]}"
     else
-      /usr/bin/sudo "${SUDO_TEST}" &>/dev/null
+      SUDO="/usr/bin/sudo"
+    fi
+
+    if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
+      ${SUDO} -l mkdir
+    else
+      # We ask for password on Linux with sudo -v
+      # to give options as shown in the
+      # 'Select the Homebrew installation folder'
+      # But first we check if the user is allowed to run mkdir with sudo.
+      # (It is possible that the user is only allowed
+      # to run a subset of commands with sudo, but not mkdir.)
+      ${SUDO} -l mkdir &>/dev/null && ${SUDO} -v
     fi
     HAVE_SUDO_ACCESS="$?"
+
   fi
 
   if [[ -z "${HOMEBREW_ON_LINUX-}" ]] && [[ "$HAVE_SUDO_ACCESS" -ne 0 ]]; then
@@ -292,7 +300,8 @@ else
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
   else
     trap exit SIGINT
-    if ! /usr/bin/sudo -n -v &>/dev/null; then
+    # If allowed to run sudo for mkdir and required to provide sudo password: prompt:
+    if /usr/bin/sudo -l mkdir &>/dev/null && ! /usr/bin/sudo -n -v &>/dev/null; then
       ohai "Select the Homebrew installation directory"
       echo "- ${tty_bold}Enter your password${tty_reset} to install to ${tty_underline}${HOMEBREW_PREFIX_DEFAULT}${tty_reset} (${tty_bold}recommended${tty_reset})"
       echo "- ${tty_bold}Press Control-D${tty_reset} to install to ${tty_underline}$HOME/.linuxbrew${tty_reset}"
--- a/install.sh
+++ b/install.sh
@@ -68,9 +68,14 @@ have_sudo_access() {
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
+      SUDO="/usr/bin/sudo ${args[*]}"
     else
-      /usr/bin/sudo -v && /usr/bin/sudo -l mkdir &>/dev/null
+      SUDO="/usr/bin/sudo"
+    fi
+    if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
+      ${SUDO} -l mkdir &>/dev/null
+    else
+      ${SUDO} -v && ${SUDO} -l mkdir &>/dev/null
     fi
     HAVE_SUDO_ACCESS="$?"
   fi
--- a/install.sh
+++ b/install.sh
@@ -67,7 +67,11 @@ have_sudo_access() {
   fi
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
-    /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null 
+    if [[ -n "${args[*]-}" ]]; then
+      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null 
+  else
+      /usr/bin/sudo -v && /usr/bin/sudo -l mkdir &>/dev/null 
+    fi
     HAVE_SUDO_ACCESS="$?"
   fi
 
--- a/install.sh
+++ b/install.sh
@@ -66,11 +66,17 @@ have_sudo_access() {
     args=("-A")
   fi
 
+  if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
+      SUDO_TEST="-l mkdir"
+  else
+      SUDO_TEST="-v"
+  fi
+
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -v &>/dev/null
+      /usr/bin/sudo "${args[@]}" "${SUDO_TEST}" &>/dev/null
     else
-      /usr/bin/sudo -v &>/dev/null
+      /usr/bin/sudo "${SUDO_TEST}" &>/dev/null
     fi
     HAVE_SUDO_ACCESS="$?"
   fi


hwalinga:
Note sure why this fails on MacOS. As I code it now, everything should be executed exactly the same for MacOS and there should only be changes for Linux.

MikeMcQuaid:
Are you saying that, without this PR, this will never request a `sudo` password on Linux in any circumstance? Thanks!

hwalinga:
> Are you saying that, without this PR, this will never request a `sudo` password on Linux in any circumstance? Thanks!

Without this PR, you will not be prompted the `sudo` password in the setup phase of the script. So, you will just directly be shown "This script will install" and then the first command that has to be executed with `sudo` will trigger the prompt. 

So, how it is in the master: https://termbin.com/z8iz

How it is in this PR: https://termbin.com/5gl0

Also note that with this PR the user is given the choice, and without this PR the install script will either install in /home/linuxbrew or ~/.linuxbrew depending if it can, not if it should.

hwalinga:
I made the checks a little more advanced to account for the multiple situations. The situation I did not include first was that a user can be able to run sudo but is not allowed to do that for mkdir.

Here they are listed:

 * User can run sudo and can run it for mkdir and needs password -> prompt for password with choice
 * User can run sudo can run it for mkdir and doesn't needs password -> Install to /home/linuxbrew, don't prompt
 * User cannot run sudo -> Install to ~/.linuxbrew, don't prompt
 * User can run sudo but cannot run it for mkdir -> Install to ~/.linuxbrew, don't prompt

MikeMcQuaid:
I'd like to bring in @maxim-belkin and @sjackman to confirm if their understanding of this matches yours and generally see what they think of this PR.

sjackman:
```console
$ docker run -it --rm homebrew/brew
$ sed -i 's/NOPASSWD://' /etc/sudoers
$ yes secret | passwd linuxbrew
$ sudo -i -u linuxbrew
$ sudo -l mkdir; echo $?
[sudo] password for linuxbrew: secret
/usr/bin/mkdir
0
$ sudo -k
$ sudo -l mkdir; echo $?
[sudo] password for linuxbrew: ^D
sudo: no password was provided
1
$ rm -r /home/linuxbrew/.linuxbrew
$ chown 555 /home/linuxbrew
$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
==> Select the Homebrew installation directory
- Enter your password to install to /home/linuxbrew/.linuxbrew (recommended)
- Press Control-D to install to /home/linuxbrew/.linuxbrew
- Press Control-C to cancel installation
[sudo] password for linuxbrew: 
```

Works as expected for me.
Can you please provide instructions to reproduce your issue in the `homebrew/brew` Docker image?

hwalinga:
Yes, no problem. Didn't know you had such a docker image. 

```console
$ docker run -it --rm homebrew/brew
$ # NOTICE: ADDING A NOPASSWD LINE HERE (/sbin/poweroff can be anything)
$ sed -i 's/NOPASSWD://; $a linuxbrew ALL=(root) NOPASSWD: /sbin/poweroff' /etc/sudoers  
$ yes secret | passwd linuxbrew
$ sudo -i -u linuxbrew
$ sudo -l mkdir; echo $?
/usr/bin/mkdir
0
$ sudo -k 
$ sudo -l mkdir; echo $?
/usr/bin/mkdir
0
$ rm -rf /home/linuxbrew/.linuxbrew/
$ sudo chown 555 /home/linuxbrew
$ # NOTICE: NO PROMPT FOR PASSWORD ON INSTALL AND AN EVENTUALLY FAILED INSTALL
$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"  
==> This script will install:
...
==> The following new directories will be created:
/home/linuxbrew/.linuxbrew/bin
...
/home/linuxbrew/.linuxbrew/Frameworks

==> /usr/bin/sudo /bin/mkdir -p /home/linuxbrew/.linuxbrew
...
/bin/mkdir: cannot create directory '/home/linuxbrew/.cache': Permission denied
Failed during: /bin/mkdir -p /home/linuxbrew/.cache/Homebrew
```

I had to take a closer look myself why `sudo -l` was initially different on the docker instance in comparison with what I have running locally. I discovered that the difference was that there are NOPASSWD in my sudoers configuration and there are none in the docker instance's. Why exactly this is different I am not sure, apparently prompting password on `sudo -l` is undefined behavior? 

Commonly, distributions do ship with such NOPASSWD configurations for for example shutdown/hibernate or other hardware interactions considered not harmful. Also not hard to imagine sysadmins to include NOPASSWD lines for their users to execute commands which they see no harm in them doing. 

With this in mind, I think including the additional check on `sudo -v` should be done to also execute the correct behavior on these cases.

sjackman:
Thank you for test reproducible test case, Hielke! That's very helpful. I can reproduce your issue. It seems to me like it may be a bug in `sudo`, and one that we'll need to work around in Homebrew, so thanks for your PR!

install.sh
```diff
@@ -67,12 +67,26 @@ have_sudo_access() {
   fi
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
+
```
sjackman, line 70:
```suggestion
```
Please remove this blank line.

install.sh
```diff
@@ -67,12 +67,26 @@ have_sudo_access() {
...
+      # We ask for password on Linux with sudo -v
+      # to give options as shown in the
+      # 'Select the Homebrew installation folder'
+      # But first we check if the user is allowed to run mkdir with sudo.
+      # (It is possible that the user is only allowed
+      # to run a subset of commands with sudo, but not mkdir.)
+      ${SUDO} -l mkdir &>/dev/null && ${SUDO} -v
     fi
     HAVE_SUDO_ACCESS="$?"
+
```
sjackman, line 89:
```suggestion
```
Please remove this blank line.

install.sh
```diff
@@ -67,12 +67,26 @@ have_sudo_access() {
   fi
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
+
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
+      SUDO="/usr/bin/sudo ${args[*]}"
```
sjackman, line 72:
```suggestion
      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo -l mkdir &>/dev/null
```

install.sh
```diff
@@ -67,12 +67,26 @@ have_sudo_access() {
...
+    if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
+      ${SUDO} -l mkdir
+    else
+      # We ask for password on Linux with sudo -v
+      # to give options as shown in the
+      # 'Select the Homebrew installation folder'
+      # But first we check if the user is allowed to run mkdir with sudo.
+      # (It is possible that the user is only allowed
+      # to run a subset of commands with sudo, but not mkdir.)
+      ${SUDO} -l mkdir &>/dev/null && ${SUDO} -v
```
sjackman, line 86:
```suggestion
      /usr/bin/sudo -v && /usr/bin/sudo -l mkdir &>/dev/null
```

install.sh
```diff
@@ -289,7 +303,8 @@ else
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
   else
     trap exit SIGINT
-    if [[ $(/usr/bin/sudo -n -l mkdir 2>&1) != *"mkdir"* ]]; then
+    # If allowed to run sudo for mkdir and required to provide sudo password: prompt:
+    if /usr/bin/sudo -l mkdir &>/dev/null && ! /usr/bin/sudo -n -v &>/dev/null; then
```
sjackman, line 307:
```suggestion
    if ! sudo -n -v &>/dev/null; then
```
We can't use `sudo -l mkdir` until `have_sudo_access` is called, or it could prompt for a password. We could use `sudo -n -l mkdir`, but unfortunately `sudo -n -l mkdir` appears to be buggy, as you've shown me.

sjackman (changes requested):
A few suggested changes to simplify.

hwalinga:
Something fishy going on on Mac. I don't own a Mac so I cannot debug this. The last time I kind of hacked this to pass by specifying this check on HOMEBREW_ON_LINUX, but it should just work identical on Mac as on Ubuntu. They should use the same 'sudo' program, I think, but not entirely sure. 

```
  bash install.sh
  shell: /bin/bash -e {0}
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
Need sudo access on macOS (e.g. the user runner to be an Administrator)!
Error: Process completed with exit code 1.
```

install.sh
```diff
@@ -68,9 +68,9 @@ have_sudo_access() {
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
-    else
-      /usr/bin/sudo -l mkdir &>/dev/null
+      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null 
+  else
```
sjackman, line 72:
```suggestion
    else
```
Fix indentation.

install.sh
```diff
@@ -289,7 +289,8 @@ else
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
   else
     trap exit SIGINT
-    if [[ $(/usr/bin/sudo -n -l mkdir 2>&1) != *"mkdir"* ]]; then
+    # If allowed to run sudo for mkdir and required to provide sudo password: prompt:
```
sjackman, line 292:
```suggestion
```
This comment is stale now. You can delete it.

sjackman (changes requested):

install.sh
```diff
@@ -68,9 +68,9 @@ have_sudo_access() {
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
-    else
-      /usr/bin/sudo -l mkdir &>/dev/null
+      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
```
sjackman, line 71:
```suggestion
      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
```
Remove trailing white space.

install.sh
```diff
@@ -68,9 +68,9 @@ have_sudo_access() {
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
-    else
-      /usr/bin/sudo -l mkdir &>/dev/null
+      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null 
+  else
+      /usr/bin/sudo -v && /usr/bin/sudo -l mkdir &>/dev/null
```
sjackman, line 73:
```suggestion
      /usr/bin/sudo -v && /usr/bin/sudo -l mkdir &>/dev/null
```
Remove trailing white space.

sjackman:
> Something fishy going on on MacOS. I don't own a Mac so I cannot debug this. The last time I kind of hacked this to pass by specifying this check on HOMEBREW_ON_LINUX, but it should just work identical on Mac as on Ubuntu. They should use the same 'sudo' program, I think, but not entirely sure.

It may be issue that is specific to running on macOS CI without a terminal. It seems like perhaps `sudo -l mkdir` does not prompt for a password on macOS CI, but `sudo -v` does, and then it fails because it doesn't have a terminal.

hwalinga:
That seems beyond my reach to fix, you might be able to configure the MacOS CI with a SUDO_ASKPASS?

MikeMcQuaid:
> Something fishy going on on MacOS. I don't own a Mac so I cannot debug this.

In that case: restore the macOS functionality so it continues to run the same commands as it did before.

MikeMcQuaid:
> That seems beyond my reach to fix, you might be able to configure the MacOS CI with a SUDO_ASKPASS?

That will break all other non-interactive installations

hwalinga:
It seems that it is related to the following:

```
By default, if the NOPASSWD tag is applied to any of a user's entries
for the current host, the user will be able to run “sudo -l” without a
password.  ***Additionally, a user may only run “sudo -v” without a pass‐
word if all of the user's entries for the current host have the
NOPASSWD tag.***  This behavior may be overridden via the verifypw and
listpw options.
```

(From the sudoers man page)

But if that isn't true on a default MacOS CI, yes I can see it breaking a lot of other macOS CI pipelines using homebrew. So, let's restore old behavior for MacOS.

MikeMcQuaid:
CI is 🟢 now, nice.

install.sh
MikeMcQuaid: Still want/need this change @sjackman?

hwalinga:
@sjackman You might want to see it again, because I basically reverted your suggestions. So, there is now a SUDO variable again, otherwise the checks on args and HOMEBREW_ON_LINUX become nested without unnecessary duplication.

sjackman:
http://manpages.ubuntu.com/manpages/precise/man5/sudoers.5.html

> By default, if the NOPASSWD tag is applied to any of a user's entries for the current host, the user will be able to run “sudo -l” without a password.  ***Additionally, a user may only run “sudo -v” without a pass‐ word if all of the user's entries for the current host have the NOPASSWD tag.***  This behavior may be overridden via the verifypw and listpw options.

😭 I do not like it, Sam I am.

hwalinga:
Yes `sudo -v` will not prompt if *all* entries have NOPASSWD, but `sudo -l` already does that if *any* have it. So, we improve with the choice for checking both `sudo -l` and `sudo -v`, but I am afraid we won't have a 100 % waterproof system.

sjackman:
For future me, three configurations of `sudo` need to be tested (at least).

### No password needed for any commands

```sudoers
linuxbrew ALL=(ALL) NOPASSWD:ALL
```

### No password needed for some commands

```sudoers
linuxbrew ALL=(ALL) ALL
linuxbrew ALL=(root) NOPASSWD:/sbin/poweroff
```

### Password needed for all commands
```sudoers
linuxbrew ALL=(ALL) ALL
```

### Permissions for only select commands

```sudoers
linuxbrew ALL=(root) NOPASSWD:/sbin/poweroff
```

install.sh
sjackman: No

sjackman (approved):
Thank you very much for your first contribution to Homebrew/install, Hielke! I thank you for sticking with this PR through the review. There were a lot of pointy sharp corners that were tricky to get just right. I appreciate your tenacity! 🏆

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @hwalinga!

andreineculau:
Actually this Github Actions pipeline shows only the Windows job running Ubuntu 18.04 under WSL getting stuck https://github.com/rokmoln/support-firecloud/runs/1452925934?check_suite_focus=true so it must be specific to the WSL environment.

hwalinga:
Fascinating! We are making a tool for MacOS and Linux, and now have a problem with Windows!

But I can see the importance of having that running, but boy that is going to be complicated to reproduce!

Would you be able to get that reproduced using Docker, or something I can use on Linux?

I believe that there is information about this in `/proc/version`, so we should just be able to do a `if Microsoft /proc/version` and maintain old behavior for WSL. 

Should you be able to tell me the contents of `/proc/version`? Maybe it is different on WSL1 and WSL2, I wouldn't know.

jonchang:
Maybe we can just add `windows-latest` to the CI for this repository so that it doesn't break in the future?

hwalinga:
Well, in essence WSL is just another Linux distro. Should not behave that differently. I am also curious if this isn't specific to WSL1. The CI specifies `main-windows-2019-minimal` which I believe only had WSL1? WSL2 should be a lot less buggy.

andreineculau:
> Fascinating! We are making a tool for MacOS and Linux, and now have a problem with Windows!

👏

> But I can see the importance of having that running, but boy that is going to be complicated to reproduce!

I think it will be easy to reproduce by testing also on WSL within Github Actions like @jonchang also noted. I can file a PR.

> maintain old behavior for WSL.
Maybe we figure out something "extra" to do on top of the current solution that works everywhere, rather than introducing another condition. But to answer your question, detecting if running under WSL is agreed to be greping for `Microsoft|WSL` in `/proc/version` as per the WSL gang themselves.

sjackman:
Thanks for the bug report, Andrei. As a workaround, try setting…
```bash
export SUDO_ASKPASS=false
```

andreineculau:
Thanks Shaun. We shall see what happens here https://github.com/rokmoln/support-firecloud/runs/1455165677?check_suite_focus=true . Might be useful for others that experience this. I will go with a "lock" for my setup.

hwalinga:
>I think it will be easy to reproduce by testing also on WSL within Github Actions like @jonchang also noted. I can file a PR.

I can't debug in a CI..., (still helps when we try to make a solution though)

>Maybe we figure out something "extra"

That might be quite hard dependent on what the extra thing is, I am not entirely what the approach would be. I think in essence the problem is that there are different sudoers configurations on different CI pipelines. Homebrew should (and will) work *interactively* on all kinds of different configurations. However, making it work *non-interactively* on all other configurations is a bit trickier, and eventually that can boil down to first parsing out /etc/sudoers and than perform actions. (In essence, this PR made the homebrew install script more interactive.)

However, another solution to the problem what might work is to just check if the script is run interactively or not and perform actions based on that. I think you can check that with `[ -t 1 ]`.

miketimofeev:
@andreineculau @sjackman @MikeMcQuaid  all our Ubuntu builds in https://github.com/actions/virtual-environments are now fail on homebrew installation. Looks like `/home/linuxbrew` directory can't be created after this changes with sudo.
This is the old installation log:
```
2020-11-25T03:28:19.8318203Z     azure-arm: ==> Next steps:
2020-11-25T03:28:19.8321904Z     azure-arm: - Run `brew help` to get started
2020-11-25T03:28:19.8332307Z     azure-arm: - Further documentation:
2020-11-25T03:28:19.8335578Z     azure-arm:     https://docs.brew.sh
2020-11-25T03:28:19.8342516Z     azure-arm: - Install the Homebrew dependencies if you have sudo access:
2020-11-25T03:28:19.8346402Z     azure-arm:     sudo apt-get install build-essential
2020-11-25T03:28:19.8358871Z     azure-arm:     See https://docs.brew.sh/linux for more information
2020-11-25T03:28:19.8363198Z     azure-arm: - Add Homebrew to your PATH in /home/packer/.profile:
2020-11-25T03:28:19.8365714Z     azure-arm:     echo 'eval $(/home/linuxbrew/.linuxbrew/bin/brew shellenv)' >> /home/packer/.profile
2020-11-25T03:28:19.8371902Z     azure-arm:     eval $(/home/linuxbrew/.linuxbrew/bin/brew shellenv)
2020-11-25T03:28:19.8375775Z     azure-arm: - We recommend that you install GCC:
2020-11-25T03:28:19.8381046Z     azure-arm:     brew install gcc
```
And this is current one:
```
2020-11-25T17:12:15.4115724Z     azure-arm: ==> Next steps:
2020-11-25T17:12:15.4117578Z     azure-arm: - Run `brew help` to get started
2020-11-25T17:12:15.4128166Z     azure-arm: - Further documentation:
2020-11-25T17:12:15.4135562Z     azure-arm:     https://docs.brew.sh
2020-11-25T17:12:15.4138943Z     azure-arm: - Install the Homebrew dependencies if you have sudo access:
2020-11-25T17:12:15.4148552Z     azure-arm:     sudo apt-get install build-essential
2020-11-25T17:12:15.4160205Z     azure-arm:     See https://docs.brew.sh/linux for more information
2020-11-25T17:12:15.4165023Z     azure-arm: - Add Homebrew to your PATH in /home/packer/.profile:
2020-11-25T17:12:15.4173081Z     azure-arm:     echo 'eval $(/home/packer/.linuxbrew/bin/brew shellenv)' >> /home/packer/.profile
2020-11-25T17:12:16.3220469Z     azure-arm:     eval $(/home/packer/.linuxbrew/bin/brew shellenv)
2020-11-25T17:12:16.3221611Z     azure-arm: - We recommend that you install GCC:
2020-11-25T17:12:16.3223527Z     azure-arm:     brew install gcc
```

maxim-belkin:
@miketimofeev , what is the difference between the above two snippets?

miketimofeev:
/home/**linuxbrew**/.linuxbrew/bin/brew shellenv
transformed to
/home/**packer**/.linuxbrew/bin/brew shellenv
packer is the default user for image-generation

immae1:
actual i have also a strange behavior in my ci pipline which executes a ansible task that install brew on linux. 

export SUDO_ASKPASS=false doesn't works for me.  

it seems that the interactive mode is blocking the execution. So my pipeline step will resume forever without output and hangs.  In the past this ansible task worked like a charm! 

Here this is my ansible task: 
```
- name: Install Brew the package manager for community software
  shell: echo -ne '\n\n\n' | bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
  become_user: "{{ devUser }}"

```

maxim-belkin:
@immae1 , add one more `\n`: root's password is requested 3 times, then you need to `Press RETURN to continue or any other key to abort`

maxim-belkin:
ah, my bad. it won't work right now. I'll have a closer look at this

immae1:
my ansible log output with export SUDO_ASKPASS =false is:
`
2020-11-25T20:07:09.4976605Z fatal: [13.94.248.75]: FAILED! => {"changed": true, "cmd": "export SUDO_ASKPASS=false; echo -ne '\\n\\n\\n\\n' | bash -c \"$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)\"", "delta": "0:00:00.592909", "end": "2020-11-25 21:07:09.408193", "msg": "non-zero return code", "rc": 1, "start": "2020-11-25 21:07:08.815284", "stderr": "sudo: unable to run false: No such file or directory\nsudo: no password was provided\nenvironment: line 117: -A: command not found", "stderr_lines": ["sudo: unable to run false: No such file or directory", "sudo: no password was provided", "environment: line 117: -A: command not found"], "stdout": "==> This script will install:\n/home/linuxbrew/.linuxbrew/bin/brew\n/home/linuxbrew/.linuxbrew/share/doc/homebrew\n/home/linuxbrew/.linuxbrew/share/man/man1/brew.1\n/home/linuxbrew/.linuxbrew/share/zsh/site-functions/_brew\n/home/linuxbrew/.linuxbrew/etc/bash_completion.d/brew\n/home/linuxbrew/.linuxbrew/Homebrew\n==> The following existing directories will be made writable by user only:\n/home/linuxbrew/.linuxbrew/share/zsh\n/home/linuxbrew/.linuxbrew/share/zsh/site-functions\n==> -A /bin/chmod u+rwx /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions\nFailed during: -A /bin/chmod u+rwx /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions", "stdout_lines": ["==> This script will install:", "/home/linuxbrew/.linuxbrew/bin/brew", "/home/linuxbrew/.linuxbrew/share/doc/homebrew", "/home/linuxbrew/.linuxbrew/share/man/man1/brew.1", "/home/linuxbrew/.linuxbrew/share/zsh/site-functions/_brew", "/home/linuxbrew/.linuxbrew/etc/bash_completion.d/brew", "/home/linuxbrew/.linuxbrew/Homebrew", "==> The following existing directories will be made writable by user only:", "/home/linuxbrew/.linuxbrew/share/zsh", "/home/linuxbrew/.linuxbrew/share/zsh/site-functions", "==> -A /bin/chmod u+rwx /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions", "Failed during: -A /bin/chmod u+rwx /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions"]}`

hwalinga:
@immae1 Thanks for the log. It exposes a small unrelated bug with the install script that makes this kind of usage impossible. 

I made a PR that solves this: https://github.com/Homebrew/install/pull/354

andreineculau:
With #354 out of the way, we should go back to the main thread 😉

sjackman:
If you have sudo access on your CI system, can you edit `/etc/sudoers` to add `NOPASSWD:` to every entry?

hwalinga:
#355 might be a little bit a more robust solution to the problem of running non-interactively on CI.

andreineculau:
I ran the script with tracing on `set -x` and it gets stuck on

https://github.com/Homebrew/install/blob/master/install.sh#L78

```
2020-11-25T21:04:29.8704155Z + should_install_command_line_tools
2020-11-25T21:04:29.8704679Z + [[ -n 1 ]]
2020-11-25T21:04:29.8705136Z + return 1
2020-11-25T21:04:29.8705567Z + [[ -t 0 ]]
2020-11-25T21:04:29.8706124Z + [[ -d /home/linuxbrew/.linuxbrew ]]
2020-11-25T21:04:29.8706926Z + execute_sudo /bin/mkdir -p /home/linuxbrew/.linuxbrew
2020-11-25T21:04:29.8707568Z + args=("$@")
2020-11-25T21:04:29.8708036Z + local -a args
2020-11-25T21:04:29.8708518Z + [[ -n '' ]]
2020-11-25T21:04:29.8708966Z + have_sudo_access
2020-11-25T21:04:29.8709468Z + local -a args
2020-11-25T21:04:29.8709939Z + [[ -n '' ]]
2020-11-25T21:04:29.8710351Z + [[ -z '' ]]
2020-11-25T21:04:29.8710783Z + [[ -n '' ]]
2020-11-25T21:04:29.8711230Z + SUDO=/usr/bin/sudo
2020-11-25T21:04:29.8711702Z + [[ -z 1 ]]
2020-11-25T21:04:29.8712152Z + /usr/bin/sudo -v
2020-11-25T21:17:09.4310720Z ++ /usr/bin/sudo -k
2020-11-25T21:17:09.4733819Z ##[error]The operation was canceled.
```

Full logs https://github.com/rokmoln/support-firecloud/commit/21ca38fc9303f16002e4559f9bb70572590eaa54/checks/1455713990/logs

As much as I'd love to be wrong, while I thought (and tested) that #354 has a side-effect of fixing this issue, it doesn't. Apparently clicking in the PR diff "View raw", doesn't show you the file contents as they would look like in the target branch (i.e. master of this repo + the diff) but how it looks like in the origin repo (i.e. @hwalinga's repo), so I basically tested a version of the installer before #341 was merged, but with the fix for `-A`.

So the issue of stuck build still stands.

---

@sjackman on some I do, on some I don't, so no go.

hwalinga:
@andreineculau 

#354 only was a solution to successfully using the `export SUDO_ASKPASS=false` trick.

See #355 for a more robust solution for CI/non-interactive usage.

MikeMcQuaid:
Reverted this in https://github.com/Homebrew/install/pull/356 until we figure out a way to make CI catch these cases.

Issue #342, opened 2020-10-30
install.sh: broken no_usable_ruby


maxim-belkin:
- [x] your problem was from running the official `install` or `uninstall` script?
- [ ] after installation: ran `brew config` and `brew doctor` and included their output with your issue? If you couldn't install: provided your OS version with the output of your issue?

## What you were trying to do (and why)

Install Homebrew after #338

## What happened (include command output)

quick reproducer (extracted two functions from install.sh + added a few debugging messages):

```sh
# Please sync with 'test_ruby()' in 'Library/Homebrew/utils/ruby.sh' from Homebrew/brew repository.
test_ruby () {
  if [[ ! -x $1 ]]
  then
    return 1
  fi

  "$1" --enable-frozen-string-literal --disable=gems,did_you_mean,rubyopt -rrubygems -e \
    "abort if Gem::Version.new(RUBY_VERSION.to_s.dup).to_s.split('.').first(2) != \
              Gem::Version.new('$REQUIRED_RUBY_VERSION').to_s.split('.').first(2)" 2>/dev/null
}

no_usable_ruby() {
  local ruby_exec
  which -a ruby | while read -r ruby_exec; do
    if test_ruby "$ruby_exec"; then
      echo $ruby_exec works
      return 1
    else
      echo $ruby_exec does not work
    fi
  done
  echo still here
  return 0
}

no_usable_ruby
```

on a Mac with /usr/bin/ruby == ruby 2.6.3p62 outputs
```
/Users/mbelkin/.rbenv/shims/ruby does not work
/usr/local/opt/ruby/bin/ruby does not work
/usr/bin/ruby works
still here
```

## What you expected to happen

not see `still here` in the above message (well, install Homebrew, actually).

## Step-by-step reproduction instructions (by running `brew` commands)

See "What happened" section above.

Also, we need to update the title of this section because one can't run `brew` commands until after they install Homebrew.

MikeMcQuaid:
Is this because the `return` is leaving the `while` and not the function?

maxim-belkin:
the body of the `which -a ruby | while read...` block is executed in a subshell, so a `return` can't exit from the function.

MikeMcQuaid:
Doh 🤦🏻

gromgit:
Good point. Should've used a process substitution instead.

PR #343, opened 2020-10-30
Revert "Improved no_usable_ruby() test"

MikeMcQuaid:
Reverts Homebrew/install#338
Fixes https://github.com/Homebrew/install/issues/342

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -221,11 +221,13 @@ test_ruby () {
 
 no_usable_ruby() {
   local ruby_exec
-  which -a ruby | while read -r ruby_exec; do
+  IFS=$'\n' # Do word splitting on new lines only
+  for ruby_exec in $(which -a ruby); do
     if test_ruby "$ruby_exec"; then
       return 1
     fi
   done
+  IFS=$' \t\n' # Restore IFS to its default value
   return 0
 }
 


maxim-belkin:
we need to add `IFS=$' \t\n' # Restore IFS to its default value` before `return 1` too.

MikeMcQuaid:
@maxim-belkin Please open a pull request for that or any other refactoring given it wasn't there before this revert.

gromgit:
Alternative 1: since process substitutions are supported in Bash 3, the pipeline can be replaced as follows:
```
no_usable_ruby() {
  local ruby_exec
  while read -r ruby_exec; do
    if test_ruby "$ruby_exec"; then
      return 1
    fi
  done < <(which -a ruby)
  return 0
}
```
Alternative 2: Setting `shopt -s lastpipe` keeps the pipeline's last command line in the current shell process, so my original patch would work as-is, and it has no effect on the only other pipeline in `install.sh`. (I actually have that in almost all my scripts for that reason, but forgot that it's not in `install.sh`.)

maxim-belkin:
Re alternative 1: yes, this works.
Re alternative 2: `lastpipe` appeared in Bash 4.2

PR #345, opened 2020-11-05
README: Make it more visible that Linux and macOS share an installer

issyl0:
- I came here to get the command for the uninstall script for something, and noticed that the Linux and WSL instructions were oddly separate.  Also, the `Linuxbrew` docs site page links back to the homepage for the main install command, which is circular.
- Instead, add a note for the pre-requisite packages for Linux.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 ```
 
-More installation information and options at https://docs.brew.sh/Installation.html.
+More installation information and options: https://docs.brew.sh/Installation.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,16 +1,14 @@
 # Homebrew (un)installer
 
-## Install Homebrew
+## Install Homebrew (on macOS or Linux)
 
 ```bash
 ```
 
 More installation information and options at https://docs.brew.sh/Installation.html.
```
EricFromCanada, line 9:
```suggestion
More installation information and options: https://docs.brew.sh/Installation.
```

MikeMcQuaid:
makes sense, thanks @issyl0!

PR #347, opened 2020-11-12
install.sh: check processor.

MikeMcQuaid:
Refuse to install unless on Intel processors. If we're on macOS ARM then output a custom message letting people know they can use Rosetta.

**This needs to be merged and in a tagged release before Big Sur is released so please prioritise testing and reviewing this ASAP CC @claui @Bo98 @fxcoudert @mistydemeo**

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -322,7 +322,7 @@ if [[ "$UNAME_MACHINE" == "arm64" ]]; then
   abort "$(cat <<EOABORT
 Homebrew is not (yet) supported on ARM processors!
 Rerun the Homebrew installer under Rosetta 2.
-If you really know what you are doing and are prepared for a very broken experience you can read use another installation option for installing on ARM:
+If you really know what you are doing and are prepared for a very broken experience you can use another installation option for installing on ARM:
   ${tty_underline}https://docs.brew.sh/Installation${tty_reset}
 EOABORT
 )"
--- a/install.sh
+++ b/install.sh
@@ -317,6 +317,19 @@ EOABORT
 )"
 fi
 
+UNAME_MACHINE="$(uname -m)"
+if [[ "$UNAME_MACHINE" == "arm64" ]]; then
+  abort "$(cat <<EOABORT
+Homebrew is not (yet) supported on ARM processors!
+Rerun the Homebrew installer under Rosetta 2.
+If you really know what you are doing and are prepared for a very broken experience you can read use another installation option for installing on ARM:
+  ${tty_underline}https://docs.brew.sh/Installation${tty_reset}
+EOABORT
+)"
+elif [[ "$UNAME_MACHINE" != "x86_64" ]]; then
+  abort "Homebrew is only supported on Intel processors!"
+fi
+
 if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
   if version_lt "$macos_version" "10.7"; then
     abort "$(cat <<EOABORT


Bo98:
Thoughts on allowing it if `HOMEBREW_DEVELOPER` is set? Native ARM is ok if people know what they are doing, as shown by those with a DTK.

MikeMcQuaid:
- We don't even support Big Sur yet so it's a definite no until that is the case (and we have Intel CI setup)
- We will need additional logic to use the new ARM prefix

install.sh
```diff
@@ -317,6 +317,17 @@ EOABORT
 )"
 fi
 
+UNAME_MACHINE="$(uname -m)"
+if [[ "$UNAME_MACHINE" == "arm64" ]]; then
+  abort "$(cat <<EOABORT
+Homebrew is not (yet) supported on ARM processors!
```
claui, line 323:
```suggestion
Homebrew is not (yet) supported on ARM processors.
```

install.sh
```diff
@@ -317,6 +317,17 @@ EOABORT
...
+Rerun the Homebrew installer under Rosetta 2.
+EOABORT
+)"
+elif [[ "$UNAME_MACHINE" != "x86_64" ]]; then
+  abort "Homebrew is only supported on Intel processors!"
+fi
```
Bo98, line 329: Is this the case on Linux as well? ARM is allowed by `brew` on Linux.

install.sh
```diff
@@ -317,6 +317,17 @@ EOABORT
 )"
 fi
 
+UNAME_MACHINE="$(uname -m)"
+if [[ "$UNAME_MACHINE" == "arm64" ]]; then
+  abort "$(cat <<EOABORT
+Homebrew is not (yet) supported on ARM processors!
+Rerun the Homebrew installer under Rosetta 2.
```
claui, line 324:
```suggestion
To rerun the Homebrew installer under Rosetta 2,
select your Terminal app in Finder and click Get Info.
Make sure that "Open using Rosetta" is checked.
Then quit and re-open your Terminal.
```

install.sh
```diff
@@ -317,6 +317,17 @@ EOABORT
...
+EOABORT
+)"
+elif [[ "$UNAME_MACHINE" != "x86_64" ]]; then
+  abort "Homebrew is only supported on Intel processors!"
```
claui, line 328:
```suggestion
  abort "Homebrew is only supported on Intel processors."
```

claui (changes requested):
Providing basic instructions for enabling Rosetta 2 would help, I think.

install.sh
```diff
@@ -317,6 +317,19 @@ EOABORT
 )"
 fi
 
+UNAME_MACHINE="$(uname -m)"
+if [[ "$UNAME_MACHINE" == "arm64" ]]; then
+  abort "$(cat <<EOABORT
+Homebrew is not (yet) supported on ARM processors!
+Rerun the Homebrew installer under Rosetta 2.
+If you really know what you are doing and are prepared for a very broken experience you can read use another installation option for installing on ARM:
```
claui, line 325:
```suggestion
If you really know what you are doing and are prepared for a very broken experience you can use another installation option for installing on ARM:
```

install.sh
MikeMcQuaid: It's not supported on Linux also, correct.

install.sh
MikeMcQuaid: I stylistically prefer `!` on hard-failing errors.

install.sh
MikeMcQuaid: I stylistically prefer `!` on hard-failing errors.

install.sh
MikeMcQuaid: @claui Does this force your Terminal to always open in Rosetta? I'm wondering if there are any nicer ways than this as it seems like this may not be what people want, I dunno. If not: no worries.

claui: Yes, it forces the Terminal to always open in Rosetta.

It may not always be what people want but I’d still prefer to recommend it than recommending to prepend `arch -x86_64`.
The latter may be more convenient but:
- it’s extremely easy to forget during daily use, and
- imho it’s also kind of awkward to describe.

claui:
Applying https://github.com/Homebrew/brew/pull/9102:

```
$ cd "$(mktemp -d)" && curl -LO https://patch-diff.githubusercontent.com/raw/Homebrew/brew/pull/9102.patch && git -C /usr/local/Homebrew am "$(pwd)/9102.patch"
Applying: keg: add codesigning
```

Unsetting `HOMEBREW_SKIP_OR_LATER_BOTTLES`:

```
$ unset HOMEBREW_SKIP_OR_LATER_BOTTLES
```

Installing a formula from bottle:

```
$ arch -x86_64 brew install nettle
==> Downloading https://homebrew.bintray.com/bottles/gmp-6.2.0.catalina.bottle.tar.gz
==> Downloading from https://d29vzk4ow07wi7.cloudfront.net/2e6acd6e62d1b8ef0800061e113aea30a63f56b32b99c010234c0420fd6d3ecf?response-con
######################################################################## 100.0%
==> Downloading https://homebrew.bintray.com/bottles/nettle-3.6.catalina.bottle.tar.gz
==> Downloading from https://d29vzk4ow07wi7.cloudfront.net/7ac7677ba653dbef81dd83ed8cde3dfcb7b464d04442886c396179932f4f9faa?response-con
######################################################################## 100.0%
==> Installing dependencies for nettle: gmp
==> Installing nettle dependency: gmp
==> Pouring gmp-6.2.0.catalina.bottle.tar.gz
🍺  /usr/local/Cellar/gmp/6.2.0: 20 files, 3.2MB
==> Installing nettle
==> Pouring nettle-3.6.catalina.bottle.tar.gz
```

Testing:

```
$ arch -x86_64 brew test nettle
==> Testing nettle
==> /usr/bin/clang test.c -L/usr/local/Cellar/nettle/3.6/lib -lnettle -o test
==> ./test
```

Works just fine.

claui (approved):

install.sh
```diff
@@ -317,6 +317,19 @@ EOABORT
 )"
 fi
 
+UNAME_MACHINE="$(uname -m)"
```
iMichka, line 320: Can we exclude the whole block when HOMEBREW_ON_LINUX is 1? Just to make sure this Mac-specific code is not run there?

install.sh
MikeMcQuaid: What if Rosetta is not yet installed (as it sounds like it isn't/won't be by default)?



> It may not always be what people want but I’d still prefer to recommend it than recommending to prepend `arch -x86_64`.

Doing this and then having something like a `rosettabrew` alias that runs `arch -x86_64 brew` feels like it'd be nice.

iMichka: > What if Rosetta is not yet installed

Maybe add a link that tells people how to install Rosetta upfront. Like:

```
If Rosetta is not installed, please follow the instructions at <url>
```

install.sh
MikeMcQuaid: @iMichka sorry the page autorefresh didn't show this until after I merged 😬. This should be there still on ARM on Linux because it's an unsupported configuration.

install.sh
MikeMcQuaid: Leaving as-is for now but will happily 👀 any PR to improve the text. Thanks for the help and testing @claui ❤️

install.sh
iMichka: Yes it is, but it will tell people to use Rosetta on ARM linux I guess?

install.sh
claui: > Leaving as-is for now but will happily 👀 any PR to improve the text.

@MikeMcQuaid Fyi, https://github.com/Homebrew/brew/pull/9126 includes a few suggestions. I plan to apply those here once they’re merged there.

PR #348, opened 2020-11-12
install: recommend Rosetta only on macOS

iMichka:


sjackman (approved):

MikeMcQuaid:
Thanks for cleaning up my mess @iMichka 😆

PR #349, opened 2020-11-15
install: Return rosetta advice on arm64 macOS

roolebo:
Current version of installer prints a terse error on DTK:

  Homebrew is only supported on Intel processors!

Fixes: 8f22b74e637 ("install: recommend Rosetta only on macOS")

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -329,7 +329,7 @@ EOABORT
 )"
 fi
 
-if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && [[ "$UNAME_MACHINE" != "x86_64" ]]; then
+if [[ "$UNAME_MACHINE" != "x86_64" ]]; then
   abort "Homebrew is only supported on Intel processors!"
 fi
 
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -329,7 +329,7 @@ EOABORT
 )"
 fi
 
-if [[ "$UNAME_MACHINE" != "x86_64" ]]; then
+if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && [[ "$UNAME_MACHINE" != "x86_64" ]]; then
```
MikeMcQuaid, line 332:
```suggestion
if [[ "$UNAME_MACHINE" != "x86_64" ]]; then
```

This should be on all platforms (in case of e.g. running on an old PPC Mac).

MikeMcQuaid:
Thanks @roolebo!

install.sh
roolebo: I totally forgot that PPC Macs exist :) Thanks for catching the issue.

PR #350, opened 2020-11-19
add env for unattended install

trollkarlen:
HOMEBREW_PREFIX env for default location of homebrew on linux

Signed-off-by: Robert Marklund <robbelibobban@gmail.com>

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -286,7 +286,7 @@ if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
  have_sudo_access
 else
   if [[ -n "${CI-}" ]] || [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] || [[ -w "/home/linuxbrew" ]] || [[ -w "/home" ]]; then
-    HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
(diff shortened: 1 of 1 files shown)


trollkarlen:
I want to move way from my linuxbrew fork for unattended install and be able to set the prefix for homebrew.
It started to complain that i should use homebrew instead of linuxbrew git:s, so I guess this will stop working in the future.

//T

install.sh
```diff
@@ -286,7 +286,7 @@ if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
  have_sudo_access
 else
   if [[ -n "${CI-}" ]] || [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] || [[ -w "/home/linuxbrew" ]] || [[ -w "/home" ]]; then
-    HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
+    HOMEBREW_PREFIX="${HOMEBREW_PREFIX:-$HOMEBREW_PREFIX_DEFAULT}"
```
trollkarlen, line 289: Added this so we can force even if we have global install location

MikeMcQuaid:
We do not want this to be in the official installer, sorry. It installs to the default prefix because that's all we support.

PR #354, opened 2020-11-25
Don't add -A to sudo when you don't use sudo.

hwalinga:
If you don't use `sudo`, you don't have to add -A to the arguments, even if SUDO_ASKPASS is set. 

This happens for example when you try to force a home user install with `export SUDO_ASKPASS=false`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -117,10 +117,10 @@ execute() {
 
 execute_sudo() {
   local -a args=("$@")
-  if [[ -n "${SUDO_ASKPASS-}" ]]; then
-    args=("-A" "${args[@]}")
-  fi
   if have_sudo_access; then
+    if [[ -n "${SUDO_ASKPASS-}" ]]; then
+      args=("-A" "${args[@]}")
+    fi
     ohai "/usr/bin/sudo" "${args[@]}"
     execute "/usr/bin/sudo" "${args[@]}"
   else


andreineculau:
You were ~fast~ ~faster~ fastest!

hwalinga:
It is slightly unrelated, it should only help with setting the `export SUDO_ASKPASS=false` trick, but a side effect of that is that it forces home folder installation. And apparently given https://github.com/Homebrew/install/pull/341#issuecomment-733915458 forcing home folder installation might have other effects.

install.sh
```diff
@@ -117,10 +117,10 @@ execute() {
 
 execute_sudo() {
   local -a args=("$@")
-  if [[ -n "${SUDO_ASKPASS-}" ]]; then
-    args=("-A" "${args[@]}")
-  fi
   if have_sudo_access; then
+    if [[ -n "${SUDO_ASKPASS-}" ]]; then
+      args=("-A" "${args[@]}")
```
maxim-belkin, line 122: "-A" "$SUDO_ASKPASS"

andreineculau:
@hwalinga right. Now I see that this script is not running with strict bash i.e. `set -euo pipefail`. I'm testing this patch.

install.sh
andreineculau: ?

andreineculau:
I can confirm that we have a go. This is a fix for #341 
https://github.com/rokmoln/support-firecloud/runs/1455732858?check_suite_focus=true

andreineculau:
I talked too soon. I have another build that is building master (i.e. without this patch), and it works. The only difference from before when it was failing is that I enabled tracing `set -x` 🤔
https://github.com/rokmoln/support-firecloud/runs/1455713990?check_suite_focus=true

install.sh
hwalinga: No, -A is not flag that takes an option. Just give -A to sudo and will use what is in that variable to get a password. 

From the man page:

```
     -A, --askpass
                 Normally, if sudo requires a password, it will read it from
                 the user's terminal.  If the -A (askpass) option is speci‐
                 fied, a (possibly graphical) helper program is executed to
                 read the user's password and output the password to the stan‐
                 dard output.  If the SUDO_ASKPASS environment variable is
                 set, it specifies the path to the helper program.  Otherwise,
                 if sudo.conf(5) contains a line specifying the askpass pro‐
                 gram, that value will be used.  For example:

                     # Path to askpass helper program
                     Path askpass /usr/X11R6/bin/ssh-askpass

                 If no askpass program is available, sudo will exit with an
                 error.
```

maxim-belkin: Oh, yes, you're correct! I misread the man page. oops!

maxim-belkin (approved):
This resolves https://github.com/Homebrew/install/pull/341#issuecomment-733916265

maxim-belkin:
I'm pinging maintainers with write access to this repo

install.sh
```diff
@@ -117,10 +117,10 @@ execute() {
...
+    fi
     ohai "/usr/bin/sudo" "${args[@]}"
     execute "/usr/bin/sudo" "${args[@]}"
   else
```
sjackman, line 126:
```ruby
execute_sudo() {
  if have_sudo_access; then
    if [[ -n "${SUDO_ASKPASS-}" ]]; then
      ohai "/usr/bin/sudo" -A "$@"
      execute "/usr/bin/sudo" -A "$@"
    else
      ohai "/usr/bin/sudo" "$@"
      execute "/usr/bin/sudo" "$@"
    fi
  else
```
Stylistic suggestion. I haven't tested this code.

Thanks for the quick fix, Hielke!

sjackman (changes requested):

install.sh
sjackman: I see it was this PR was merged while I was typing up this comment! Feel free to open a PR with this stylistic change if you prefer it, but also feel free to leave it as is.

hwalinga: Well, we could just get rid of the whole -A in args together. Same `have_sudo_access`. 

But this already got merged, let's save this for another time.

PR #355, opened 2020-11-25
Set a NONINTERACTIVE variable to prevent prompting for password.

hwalinga:
Performs a few checks to see if we are running non-interactively (these checks can be extended if we know more checks, maybe some specific with CI? I don't have so much experience with CI). If this is the case, we don't prompt for a password.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -u
 
+# Check if script is run non-interactively (e.g. CI)
+# If it is run non-interactively we should not prompt for passwords.
+if [[ ! -t 0 ]] || [[ ! -t 1 ]] || [[ $SHLVL == 1 ]] || ! tty -s; then
+  NONINTERACTIVE=1
+fi
+
 # First check if the OS is Linux.
 if [[ "$(uname)" = "Linux" ]]; then
   HOMEBREW_ON_LINUX=1
@@ -61,6 +67,9 @@ tty_bold="$(tty_mkbold 39)"
 tty_reset="$(tty_escape 0)"
 
 have_sudo_access() {
+  if [[ ! -z "${NONINTERACTIVE-}" ]]; then
+    return 0
+  fi
   local -a args
   if [[ -n "${SUDO_ASKPASS-}" ]]; then
     args=("-A")
--- a/install.sh
+++ b/install.sh
@@ -67,7 +67,7 @@ tty_bold="$(tty_mkbold 39)"
 tty_reset="$(tty_escape 0)"
 
 have_sudo_access() {
-  if [[ ! -z "${NONINTERACTIVE-}" ]]; then
+  if [[ -n "${NONINTERACTIVE-}" ]]; then
     return 0
   fi
   local -a args


install.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -u
 
+# Check if script is run non-interactively (e.g. CI)
+# If it is run non-interactively we should not prompt for passwords.
+if [[ ! -t 0 ]] || [[ ! -t 1 ]] || [[ $SHLVL == 1 ]] || ! tty -s; then
+  NONINTERACTIVE=1
+fi
```
andreineculau, line 8: I believe you can detect a non-interactive session simpler via `[[ $- != *i* ]]`

install.sh
```diff
@@ -61,6 +67,9 @@ tty_bold="$(tty_mkbold 39)"
 tty_reset="$(tty_escape 0)"
 
 have_sudo_access() {
+  if [[ ! -z "${NONINTERACTIVE-}" ]]; then
```
andreineculau, line 70: maybe `! -z` -> `-n`?

install.sh
hwalinga: Yes, $- will only contain `i` if you ask for `$-` in an interactive shell. But if you invoke a shell script from the terminal (and so still have interactive usage over it) this variable does not have `i` anymore. And the install script is not sourced, but is invoked as a shell script on a terminal. 

Just create a script and put `echo $-` in there and you while see when you `bash file.sh`. (Or execute `bash <(echo 'echo $-')`)

install.sh
hwalinga: Yes, shellcheck already caught me on this!

install.sh
andreineculau: I came back here to take it back, but again you were faster :)

By coincidence I saw that in other parts this repo tests for interactivity simple with `-t 0`. https://github.com/Homebrew/install/blob/master/uninstall.sh#L287 . I prefer consistency in either direction, but that's just me.

MikeMcQuaid:
This shouldn't need to be set manually. We should detect this properly without users needing to set a variable.

MikeMcQuaid:
Passing on this in favour of a full revert in https://github.com/Homebrew/install/pull/356, sorry. Please create a new PR incorporating all these changes and we'll consider it once the CI is setup properly.

hwalinga:
>This shouldn't need to be set manually. We should detect this properly without users needing to set a variable.

Well, it isn't. If the install is non-interactive, it will be automatically detected and the NONINTERACTIVE variable is set accordingly. 

Of course, you can set it manually if you want to. 

But let's reconsider this when the CI is setup properly.

MikeMcQuaid:
> Well, it isn't. If the install is non-interactive, it will be automatically detected and the NONINTERACTIVE variable is set accordingly.

Yes, I see that on reread, sorry.



> But let's reconsider this when the CI is setup properly.

Agreed. Thinking about it more I think the issue here was adding the special case for macOS. There's nothing special about the macOS `sudo` configuration that means it will/can never apply on Linux.

hwalinga:
I think the mistake we made was thinking that MacOS was special enough that such a problem would only occur in MacOS. But it indeed wasn't really `sudo` on MacOS, but rather how /etc/sudoers is configured on the CI version of MacOS.

MikeMcQuaid:
> I think the mistake we made was thinking that MacOS was special enough that such a problem would only occur in MacOS. But it indeed wasn't really `sudo` on MacOS, but rather how /etc/sudoers is configured on the CI version of MacOS.

Agreed 💯 👍🏻

MikeMcQuaid:
https://github.com/Homebrew/install/pull/357 has been merged so I think the revert of https://github.com/Homebrew/install/pull/356 can happen whenever and the additional CI cases should catch at least one of the regressions as well as not special-casing macOS.

PR #353, opened 2020-11-25
test under WSL on Github Actions

andreineculau:
To test behaviour mentioned in https://github.com/Homebrew/install/pull/341#issuecomment-733786001
where the installer gets stuck, probably waiting for user input.

I duplicated the steps, otherwise each step needs conditionals to detect WSL and pipe commands to it (default `bash` on Github Actions' Windows agent is actually Git/MSYS2 `bash`).

Status: WIP

PR #357, opened 2020-11-26
workflows/tests: improve CI.

MikeMcQuaid:
- ~setup and test under WSL~ extracted to https://github.com/Homebrew/install/pull/360
- Use `/bin/bash` everywhere to ensure we're using the system Bash
- Run the install script more like how we do on the homepage
- Split up more commands so their output is more obvious

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -1,55 +1,62 @@
 name: GitHub Actions CI
 on:
   push:
-    branches: master
+    branches:
+      - master
   pull_request:
 jobs:
   tests:
     strategy:
       matrix:
         os:
-        - macOS-latest
-        - ubuntu-latest
+          - macOS-latest
+          - ubuntu-latest
     runs-on: ${{matrix.os}}
     steps:
-    - name: Set up Git repository
-      uses: actions/checkout@main
+      - name: Set up Git repository
+        uses: actions/checkout@main
 
-    - name: Cleanup macOS
-      if: runner.os == 'macOS'
-      run: |
-        sudo rm -rf /Applications/Xcode.app \
-                    /Library/Developer/CommandLineTools
-        sudo xcode-select --reset
+      - name: Cleanup macOS
+        if: runner.os == 'macOS'
+        run: |
+          sudo rm -rf /Applications/Xcode.app \
+                      /Library/Developer/CommandLineTools
+          sudo xcode-select --reset
 
-    - name: Set up Homebrew PATH
-      run: |
-        if [ "$RUNNER_OS" = "macOS" ]; then
-          echo "/usr/local/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
-        else
-          echo "/home/linuxbrew/.linuxbrew/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
-        fi
+      - name: Set up Homebrew PATH
+        run: |
+          if [ "${RUNNER_OS}" = "macOS" ]; then
+            echo "/usr/local/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
+          else
+            echo "/home/linuxbrew/.linuxbrew/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
+          fi
 
-    - name: Test uninstalling GitHub Actions Homebrew
-      run: |
-        if which brew &>/dev/null; then
-          bash uninstall.sh -d >/dev/null
-          bash uninstall.sh -f >/dev/null
-        fi
+      - name: Uninstall GitHub Actions Homebrew
+        run: |
+          if which brew &>/dev/null; then
+            /bin/bash uninstall.sh -d >/dev/null
+            /bin/bash uninstall.sh -f >/dev/null
+          fi
 
-    - name: Test install.sh
-      run: |
-        bash install.sh
-        brew install ack
-        bash uninstall.sh -f >/dev/null
-        bash install.sh
+      - run: /bin/bash -c "$(cat install.sh)"
 
-    - name: Install Shellcheck
-      run: brew install shellcheck
+      - run: brew install ack
 
-    - run: shellcheck *.sh
+      - run: /bin/bash uninstall.sh -f >/dev/null
 
-    - name: Test uninstalling newly installed Homebrew
-      run: |
-        bash uninstall.sh -d >/dev/null
-        bash uninstall.sh -f >/dev/null
+      - run: /bin/bash -c "$(cat install.sh)"
+
+      - name: Uninstall and reinstall with sudo NOPASSWD
+        if: runner.os == 'linux'
+        run: |
+          echo "$USER ALL=(ALL) NOPASSWD:ALL" | sudo tee "/etc/sudoers.d/$USER"
+          /bin/bash uninstall.sh -f >/dev/null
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -30,26 +31,31 @@ jobs:
           echo "/home/linuxbrew/.linuxbrew/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
         fi
 
-    - name: Test uninstalling GitHub Actions Homebrew
+    - name: Set up WSL
+      if: runner.os == 'windows'
+      uses: Vampire/setup-wsl@v1
```
reitermarkus, line 36: Given this is third-party, this should be pinned to a checksum.

MikeMcQuaid: @reitermarkus Done. Is that why CI isn't even starting?

andreineculau: @MikeMcQuaid `vampire/setup-wsl@bb4708da2512aac2a4c2947effc1061ecc0a9aef is not allowed to be used in Homebrew/install. Actions in this workflow must be: created by GitHub, within a repository owned by Homebrew or match the following: Homebrew/*, n1hility/cancel-previous-runs@*, peter-evans/create-or-updat...`

see https://github.blog/2020-10-05-fine-tune-access-to-external-actions/

reitermarkus: Yes, needs to be allowed first.

andreineculau:
in its current state, you will set up WSL *and not use it* it seems.
because the default shell on Windows (Github Actions) is `powershell` and then calling `/bin/bash` will fail, and even reverting to calling `bash` will call git/msys2 `bash` https://github.com/actions/virtual-environments/pull/1648 rather than the WSL `bash`. Time will tell.

even if `bash` will end up running inside WSL, `brew install` will fail, so I guess if the above holds, and you want not to duplicate the commands as I did in #353 , then `brew install ack` should be written as `bash -c "brew install ack"`. basically all the steps need `bash -c`

if this works, I'm going to steal the setup. dry ftw

.github/workflows/tests.yml
MikeMcQuaid: > @MikeMcQuaid `vampire/setup-wsl@bb4708da2512aac2a4c2947effc1061ecc0a9aef is not allowed to be used in Homebrew/install. Actions in this workflow must be: created by GitHub, within a repository owned by Homebrew or match the following: Homebrew/*, n1hility/cancel-previous-runs@*, peter-evans/create-or-updat...`

Thanks! Where are you seeing that message?

MikeMcQuaid:
> in its current state, you will set up WSL _and not use it_ it seems.

Yup, I'm somewhat expecting that and will fix it up when I see the exact error.

.github/workflows/tests.yml
andreineculau: https://github.com/Homebrew/install/actions/runs/384968078

andreineculau:
:( I see the `setup-wsl` action fails. Maybe the author @Vampire can pitch in to help.
I myself have not used it before, but I thought it has it's place in such a simple setup here.

Alternatively, you can have a look at how I setup WSL using shell commands here https://github.com/rokmoln/support-firecloud/blob/master/.github/workflows/main.yml#L115-L131 . Note that I'm creating a `github` user for WSL. I'm not 100% sure but I think you can set it as default with `ubuntu config --default-user github`, so that your idea about "just call bash cross-platform" still stands.

MikeMcQuaid:
@andreineculau Thanks for the link! Don't worry about it, I'll get there.

andreineculau:
Ah, now I see it's actually our commands trying to creating users and failing, not the action itself. Then I guess my snippet will be useful after all :+1:

andreineculau:
so I tried some things on my own, and I'm quite confident in saying that it is not possible to set a default user with the WSL version (v1) that Github Actions ships on windows-latest, even after writing directly to the registry https://github.com/microsoft/WSL/issues/3974#issuecomment-522921145 or restarting lxssmanager https://askubuntu.com/questions/816732/how-to-change-default-user-in-wsl-ubuntu-bash-on-windows-10

what I ended up with is creating a `wsl-bash` script which checks for `wsl`, and runs `bash` via `wsl` with a specific user, or else it proxies to the env's `bash`

https://github.com/rokmoln/support-firecloud/blob/124959b/bin/wsl-bash

🤞

Vampire:
Did I get it right that you don't need my help anymore or are there questions left regarding my action?
Btw. I don't move tags or they wouldn't be tags, so you could also use `@v1.0.1` instead of an explicit SHA if you like. :-)

reitermarkus:
> I don't move tags or they wouldn't be tags, so you could also use @v1.0.1 instead of an explicit SHA if you like.

It's not about versioning, it's about security. A commit SHA will never change, while a bad actor can change a tag to something else at any time.

Vampire:
Well, if you cannot even trust me that I don't move a tag, you shouldn't trust me at all and don't use my action. :-D

But honestly, I've had a quick look at what you try.
Have a look at PR #359 :-)

andreineculau:
@Vampire  it's not about trusting you, but reading the technical possibilities and deciding where to feel safe. It is possible for accounts to be hacked and it is possible to git push --force a tag.

Vampire:
@MikeMcQuaid that's why I made it in two steps in my PR. ;-)
If you set the shell to use the `runner` user and then try to use that shell to create the runner user it cannot work.
You either have to split it like I have shown in my PR, or configure the `shell` for the "Set up WSL users" step to use `wsl-bash {0}`, so that the `root` user is used for that run step.

Actually alternatively you can wait 1-2 days until I finish an update I'm knitting for you, then you even don't have to create the user manually if you don't want to ;-) https://github.com/Vampire/setup-wsl/compare/master...f9a6d15#diff-1e26d92e3b733c2c40cc73393d45d7f65f0429033c3096707cd3facdc5ca84e9

MikeMcQuaid:
Have extracted WSL stuff to https://github.com/Homebrew/install/pull/360. Don't think it needs to block this and sounds like @Vampire has some nice stuff in the pipeline for us. Thanks for the help ❤️

MikeMcQuaid:
I also think WSL will be nice-to-have but I'm increasingly unconvinced that the WSL issue is WSL specific and not just related to specific `sudo` settings that can be replicated on Linux.

MikeMcQuaid:
> Actually alternatively you can wait 1-2 days until I finish an update I'm knitting for you, then you even don't have to create the user manually if you don't want to ;-) [Vampire/setup-wsl@master...e314734diff-1e26d92e3b733c2c40cc73393d45d7f65f0429033c3096707cd3facdc5ca84e9](https://github.com/Vampire/setup-wsl/compare/master...e314734diff-1e26d92e3b733c2c40cc73393d45d7f65f0429033c3096707cd3facdc5ca84e9)

@Vampire Yes, this sounds and would be great.

Vampire:
Well, you can just use it almost like you had it in 1ae58b9d9347dff6f328b17853de9079535a942d, you just missed to use `shell: wsl-bash {0}` instead of `shell: wsl-bash` and then later replace with the new version and less boiler-plate.

MikeMcQuaid:
@Vampire I've pushed that to https://github.com/Homebrew/install/pull/360, let's continue discussion over there.

PR #358, opened 2020-11-26
execute_sudo: Fix when SUDO_ASKPASS without sudo

sjackman:
Fix a bug when `SUDO_ASKPASS` is defined but `have_sudo_access` is false.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -117,10 +117,10 @@ execute() {
 
 execute_sudo() {
   local -a args=("$@")
-  if [[ -n "${SUDO_ASKPASS-}" ]]; then
-    args=("-A" "${args[@]}")
-  fi
   if have_sudo_access; then
+    if [[ -n "${SUDO_ASKPASS-}" ]]; then
+      args=("-A" "${args[@]}")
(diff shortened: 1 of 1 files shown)


sjackman:
cc @hwalinga @andreineculau

MikeMcQuaid (changes requested):
Would love to see https://github.com/Homebrew/install/pull/357 land before we make any other changes here.

andreineculau (approved):

hwalinga:
You don't want to simplify with ?

```sh
if ..
   sudo -A args
else 
   sudo args
fi
```

sjackman:
> You don't want to simplify with ?

I wrote that up first, and then decided that I like this better. 🤷‍♂️

sjackman:
Sure. Blocked on PR https://github.com/Homebrew/install/pull/357

MikeMcQuaid:
https://github.com/Homebrew/install/pull/357 has been merged so I think this can land whenever!

MikeMcQuaid (approved):

PR #359, opened 2020-11-27
Fix default WSL user

Vampire:


MikeMcQuaid:
Thanks for this! Will consider this when https://github.com/Homebrew/install/pull/360 is ready.

PR #360, opened 2020-11-27
workflows/tests: add Windows 10 WSL support.

MikeMcQuaid:
Perhaps worth waiting until https://github.com/Vampire/setup-wsl/compare/master...e314734diff-1e26d92e3b733c2c40cc73393d45d7f65f0429033c3096707cd3facdc5ca84e9

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -44,21 +44,17 @@ jobs:
 
       - name: Install WSL
         if: runner.os == 'windows'
-        # https://github.com/Vampire/setup-wsl/releases/tag/v1.0.1
-        uses: Vampire/setup-wsl@bb4708da2512aac2a4c2947effc1061ecc0a9aef
+        # https://github.com/Vampire/setup-wsl/releases/tag/v1.1.0
+        uses: Vampire/setup-wsl@c8afb39d908a55f133accd98a27e160edc255810
         with:
           distribution: Ubuntu-16.04
-          wsl-shell-command: 'bash -c "sudo -u runner bash --noprofile --norc -euo pipefail "\'
-
-      - name: Set up WSL users
-        if: runner.os == 'windows'
-        shell: wsl-bash {0}
-        run: |
-          adduser runner --disabled-password --gecos ""
+          wsl-shell-user: runner
 
       - name: Set up Homebrew PATH
+        env:
+          WSLENV: GITHUB_PATH/p
         run: |
-          if [ "${RUNNER_OS}" = "macOS" ]; then
+          if [ "${{ runner.os }}" = "macOS" ]; then
             echo "/usr/local/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
           else
             echo "/home/linuxbrew/.linuxbrew/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -30,7 +30,6 @@ jobs:
         run: |
           git config --global core.autocrlf false
           git config --global core.eol lf
-          echo "/home/linuxbrew/.linuxbrew/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
 
       - name: Set up Git repository
         uses: actions/checkout@main
@@ -49,10 +48,10 @@ jobs:
         with:
           distribution: Ubuntu-16.04
           wsl-shell-user: runner
+          additional-packages: build-essential
 
       - name: Set up Homebrew PATH
-        env:
-          WSLENV: GITHUB_PATH/p
+        if: runner.os != 'windows'
         run: |
           if [ "${{ runner.os }}" = "macOS" ]; then
             echo "/usr/local/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
@@ -67,7 +66,18 @@ jobs:
             /bin/bash uninstall.sh -f >/dev/null
           fi
 
+      - name: Set up WSL environment
+        if: runner.os == 'windows'
+        shell: "wsl-bash -u root {0}"
+        run: |
+          echo "runner ALL=(ALL) NOPASSWD:ALL" | tee /etc/sudoers.d/runner
+          chmod 644 /etc/sudoers.d/runner
+          echo -e "#!/bin/bash\nexec /home/linuxbrew/.linuxbrew/bin/brew \"\$@\"" | tee /usr/local/bin/brew
+          chmod 755 /usr/local/bin/brew
+
       - run: /bin/bash -c "$(cat install.sh)"
+
+      - run: brew config
 
       - run: brew install ack
 
@@ -85,6 +95,7 @@ jobs:
       - run: brew install shellcheck
 
       - run: shellcheck *.sh
+        if: runner.os != 'windows'
 
       - run: /bin/bash uninstall.sh -d >/dev/null
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -11,8 +11,27 @@ jobs:
         os:
           - macOS-latest
           - ubuntu-latest
+          - windows-latest
+        include:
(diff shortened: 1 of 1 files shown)


Vampire:
Ah, forget what I said about just missing `{0}`, I'm stupid.
To do it now you need to do it like I showed in my PR.

MikeMcQuaid:
@Vampire I think I'll wait until there's a new release for as much of this as possible 🙇🏻. I'll adjust the PR here once that's the case (or you can open your own PR here and try to get it working).

Vampire:
Done :-)

.github/workflows/tests.yml
```diff
@@ -38,6 +67,10 @@ jobs:
             /bin/bash uninstall.sh -f >/dev/null
           fi
 
+      - name: Set up WSL sudo
+        if: runner.os == 'windows'
+        run: echo "$USER ALL=(ALL) NOPASSWD:ALL" | sudo tee "/etc/sudoers.d/$USER"
```
Vampire, line 72: You are trying to use `sudo` to allow you doing `sudo` ;-)

MikeMcQuaid: lol yup, doh.

.github/workflows/tests.yml
```diff
@@ -38,6 +67,11 @@ jobs:
             /bin/bash uninstall.sh -f >/dev/null
           fi
 
+      - name: Set up WSL sudo
+        if: runner.os == 'windows'
+        shell: "wsl-bash -u root {0}"
+        run: echo "runner ALL=(ALL) NOPASSWD:ALL" | tee "/etc/sudoers.d/runner"
```
MikeMcQuaid, line 73:
```suggestion
        run: |
          echo "runner ALL=(ALL) NOPASSWD:ALL" | tee "/etc/sudoers.d/runner"
          chmod 644 /etc/sudoers.d/runner
```

.github/workflows/tests.yml
```diff
@@ -38,6 +67,13 @@ jobs:
             /bin/bash uninstall.sh -f >/dev/null
           fi
 
+      - name: Set up WSL sudo
+        if: runner.os == 'windows'
+        shell: "wsl-bash -u root {0}"
+        run: |
+          echo "runner ALL=(ALL) NOPASSWD:ALL" | tee "/etc/sudoers.d/runner"
+          chmod 644 /etc/sudoers.d/runner
```
MikeMcQuaid, line 75:
```suggestion
          chmod 644 /etc/sudoers.d/runner
          echo "/home/linuxbrew/.linuxbrew/bin:/usr/bin:/bin" >> ${GITHUB_PATH}
```

MikeMcQuaid:
@Vampire Any ideas/suggestions on how to set the PATH for within WSL?

Vampire:
Within WSL you have a normal Linux distribution, so I'd say whatever is appropriate for that Linux distribution you chose.
The important point is, that on the basis of the default shell commands on GitHub action: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions#using-a-specific-shell
The default `wsl-shell-command` for my action is `bash --noprofile --norc -euo pipefail '%wslScriptFile%'` which means that neither Bash profile nor Bash RC scfripts are executed.
I would have said use `/etc/environment`, but unfortunately that seems to be broken according to https://github.com/microsoft/WSL/issues/1405.
So what you can do is to use a custom `wsl-shell-command` that does not suppress profile and / or rc, so you can use `/etc/profile`, `/etc/profile.d/`, `/etc/bash.bashrc`, `~/bash_profile` or `~/.bashrc`.

.github/workflows/tests.yml
```diff
@@ -38,8 +65,18 @@ jobs:
             /bin/bash uninstall.sh -f >/dev/null
           fi
 
+      - name: Set up WSL environment
+        if: runner.os == 'windows'
+        shell: "wsl-bash -u root {0}"
+        run: |
+          echo "runner ALL=(ALL) NOPASSWD:ALL" | tee "/etc/sudoers.d/runner"
+          chmod 644 /etc/sudoers.d/runner
+          ln -sf /home/linuxbrew/.linuxbrew/bin/brew /usr/local/bin/brew
```
Vampire, line 74: If you don't want to use a custom shell command, you might instead use a script instead of symlink as `/usr/local/bin/brew` that first sets the `PATH` and then calls `brew`.

Vampire: Oh, btw, your try with removing `--profile` did not work because you are not starting a login shell.
I reread part of the bash man page.
rc-files are only executed if a shell is interactive.
profile-files are only executed if a shell is a login shell.
So actually `--noprofile` and `--norc` are more documentary than effective.
So you could write your path manipulation to the profile file and use a custom `wsl-shell-command` without `--noprofile` and instead with `--login` or `-l` which makes the shell a login shell.

Vampire: Or you write your path manipulation to one of the rc-files and use a custom shell command without `--norc` and instead with `-i` which makes it interactive.

.github/workflows/tests.yml
```diff
@@ -38,8 +65,18 @@ jobs:
             /bin/bash uninstall.sh -f >/dev/null
           fi
 
+      - name: Set up WSL environment
+        if: runner.os == 'windows'
+        shell: "wsl-bash -u root {0}"
+        run: |
+          echo "runner ALL=(ALL) NOPASSWD:ALL" | tee "/etc/sudoers.d/runner"
+          chmod 644 /etc/sudoers.d/runner
+          echo "#!/bin/bash\nexec /home/linuxbrew/.linuxbrew/bin/brew \"\$@\"" | tee "/etc/sudoers.d/runner"
```
Vampire, line 74: copy & paste error ;-)

Vampire: And it will probably not work without setting the `PATH`

.github/workflows/tests.yml
```diff
@@ -38,8 +65,19 @@ jobs:
             /bin/bash uninstall.sh -f >/dev/null
           fi
 
+      - name: Set up WSL environment
+        if: runner.os == 'windows'
+        shell: "wsl-bash -u root {0}"
+        run: |
+          echo "runner ALL=(ALL) NOPASSWD:ALL" | tee /etc/sudoers.d/runner
+          chmod 644 /etc/sudoers.d/runner
+          echo "#!/bin/bash\nexec /home/linuxbrew/.linuxbrew/bin/brew \"\$@\"" | tee /usr/local/bin/brew
```
Vampire, line 74:
```suggestion
          echo -e "#!/bin/bash\nexport PATH=\"/home/linuxbrew/.linuxbrew/bin:\$PATH\"\nexec /home/linuxbrew/.linuxbrew/bin/brew \"\$@\"" | tee /usr/local/bin/brew
```

.github/workflows/tests.yml
```diff
@@ -38,8 +65,20 @@ jobs:
...
 
+      - name: Set up WSL environment
+        if: runner.os == 'windows'
+        shell: "wsl-bash -u root {0}"
+        run: |
+          echo "runner ALL=(ALL) NOPASSWD:ALL" | tee /etc/sudoers.d/runner
+          chmod 644 /etc/sudoers.d/runner
+          echo -e "#!/bin/bash\nexec /home/linuxbrew/.linuxbrew/bin/brew \"\$@\"" | tee /usr/local/bin/brew
+          chmod 755 /usr/local/bin/brew
+          apt-get install build-essential -y
```
Vampire, line 76: You can let the action do that right away if you like wiht `additional-packages` input. :-)

Vampire: And otherwise you either need to do `update: true` for the action or manually do `apt-get update` before the install. :-)

MikeMcQuaid:
Thanks for the help @Vampire!

Vampire:
You're welcome, now you just need a real Windows port :-D

PR #361, opened 2020-11-27
Use sudo -v instead of sudo -l but preserve possibility to do non-interactive installs

hwalinga:
This bring back in the changes of #341 but without the special check of MacOS. 

Instead check for non-interactive usage, which then in turn forces an install without having going to prompt for the password. (As done in #355.)

---

Additionally, if SUDO_ASKPASS is set in a non-interactive install, still try to make use of SUDO_ASKPASS, because

 1. You can provide the password with a helper program, if you have actually still *need* to provide the password, but want to do that non-interactively.

 2. But you can also make use of the `export SUDO_ASKPASS=false` trick to work and force a home folder installation.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -3,7 +3,7 @@ set -u
 
 # Check if script is run non-interactively (e.g. CI)
 # If it is run non-interactively we should not prompt for passwords.
-if [[ ! -t 0 || ! -t 1 || $SHLVL == 1 ]] || ! tty -s; then
+if [[ ! -t 0 || -n "${CI-}" ]]; then
   NONINTERACTIVE=1
 fi
 
@@ -67,13 +67,11 @@ tty_bold="$(tty_mkbold 39)"
 tty_reset="$(tty_escape 0)"
 
 have_sudo_access() {
-  if [[ -n "${NONINTERACTIVE-}" && -z "${SUDO_ASKPASS-}" ]]; then
-    return 0
-  fi
-
   local -a args
   if [[ -n "${SUDO_ASKPASS-}" ]]; then
     args=("-A")
+  elif [[ -n "${NONINTERACTIVE-}" ]]; then
+    args=("-n")
   fi
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
--- a/install.sh
+++ b/install.sh
@@ -67,7 +67,7 @@ tty_bold="$(tty_mkbold 39)"
 tty_reset="$(tty_escape 0)"
 
 have_sudo_access() {
-  if [[ -n "${NONINTERACTIVE-}" ]]; then
+  if [[ -n "${NONINTERACTIVE-}" && -z "${SUDO_ASKPASS-}" ]]; then
     return 0
   fi
 
--- a/install.sh
+++ b/install.sh
@@ -298,7 +298,10 @@ fi
 if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
  have_sudo_access
 else
-  if [[ -n "${CI-}" ]] || [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] || [[ -w "/home/linuxbrew" ]] || [[ -w "/home" ]]; then
+  if [[ -n "${NONINTERACTIVE-}" ]] ||
+     [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] ||
+     [[ -w "/home/linuxbrew" ]] ||
+     [[ -w "/home" ]]; then
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
   else
     trap exit SIGINT
@@ -479,7 +482,7 @@ if should_install_command_line_tools; then
   ohai "The Xcode Command Line Tools will be installed."
 fi
 
-if [[ -t 0 && -z "${CI-}" ]]; then
+if [[ -z "${NONINTERACTIVE-}" ]]; then
   wait_for_user
 fi
 
--- a/install.sh
+++ b/install.sh
@@ -3,7 +3,7 @@ set -u
 
 # Check if script is run non-interactively (e.g. CI)
 # If it is run non-interactively we should not prompt for passwords.
-if [[ ! -t 0 ]] || [[ ! -t 1 ]] || [[ $SHLVL == 1 ]] || ! tty -s; then
+if [[ ! -t 0 || ! -t 1 || $SHLVL == 1 ]] || ! tty -s; then
   NONINTERACTIVE=1
 fi
 
--- a/install.sh
+++ b/install.sh
@@ -67,6 +67,9 @@ tty_bold="$(tty_mkbold 39)"
 tty_reset="$(tty_escape 0)"
 
 have_sudo_access() {
+  if [[ -n "${NONINTERACTIVE-}" ]]; then
+    return 0
+  fi
   local -a args
   if [[ -n "${SUDO_ASKPASS-}" ]]; then
     args=("-A")
--- a/install.sh
+++ b/install.sh
@@ -67,9 +67,6 @@ tty_bold="$(tty_mkbold 39)"
 tty_reset="$(tty_escape 0)"
 
 have_sudo_access() {
-  # if [[ -n "${NONINTERACTIVE-}" ]]; then
-  #   return 0
-  # fi
   local -a args
   if [[ -n "${SUDO_ASKPASS-}" ]]; then
     args=("-A")
@@ -79,13 +76,14 @@ have_sudo_access() {
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
     if [[ -n "${args[*]-}" ]]; then
-        echo "HELLO RUN COMMAND with args"
-        echo THIS: "${args[@]}"
-      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir
-      echo "HAVE PRINTED ERROR"
-      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
+      SUDO="/usr/bin/sudo ${args[*]}"
     else
-      /usr/bin/sudo -v && /usr/bin/sudo -l mkdir &>/dev/null
+      SUDO="/usr/bin/sudo"
+    fi
+    if [[ -n "${NONINTERACTIVE-}" ]]; then
+      ${SUDO} -l mkdir &>/dev/null
+  else
+      ${SUDO} -v && ${SUDO} -l mkdir &>/dev/null
     fi
     HAVE_SUDO_ACCESS="$?"
   fi
--- a/install.sh
+++ b/install.sh
@@ -82,7 +82,7 @@ have_sudo_access() {
     fi
     if [[ -n "${NONINTERACTIVE-}" ]]; then
       ${SUDO} -l mkdir &>/dev/null
-  else
+    else
       ${SUDO} -v && ${SUDO} -l mkdir &>/dev/null
     fi
     HAVE_SUDO_ACCESS="$?"
--- a/install.sh
+++ b/install.sh
@@ -78,9 +78,9 @@ have_sudo_access() {
 
   if [[ -z "${HAVE_SUDO_ACCESS-}" ]]; then
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
+      /usr/bin/sudo "${args[@]}" -v && /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
     else
-      /usr/bin/sudo -l mkdir &>/dev/null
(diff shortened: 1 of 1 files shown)


hwalinga:
It worked!

Now let's hope all other configurations around the world don't break this magic non-interactive check I made. 

About that check. I check a few things (All basically just check if there is a connection between the program and a terminal, except SHLVL which checks nestedness of shell sessions, and if you install on a terminal you always have a nested shell). However `uninstall.sh` makes a similar check but only checks for `-t 0`. So, don't know if we just want to check only that, or just think that maybe the other checks might catch something unexpectedly while `-t 0` doesn't?

@sjackman @MikeMcQuaid You might be the best to carry on with further reviewing this.

gromgit:
I think `! -t 0` is sufficient for this purpose, since the underlying question being asked is "can `sudo` or any other program expect a response from a terminal prompt?" `-t 0` implies that either a human or a pty-wielding program can provide that input.

sjackman:
I haven't yet tested it, but it looks good to me in principle. Thanks for keeping at this issue, Hielke.

install.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -u
 
+# Check if script is run non-interactively (e.g. CI)
+# If it is run non-interactively we should not prompt for passwords.
+if [[ ! -t 0 || ! -t 1 || $SHLVL == 1 ]] || ! tty -s; then
+  NONINTERACTIVE=1
+fi
```
MikeMcQuaid, line 8: Can you try to remove this check and instead we adjust the `sudo` logic accordingly so it does not block?

MikeMcQuaid (changes requested):

install.sh
hwalinga: If you have any ideas?

Like I said before, there isn't really any way I can think off.

Also, it just boils down to, either you install it on your own computer in an interactive terminal and want to be prompted, or you install it non-interactively and don't want to be prompted. This is the clearest way to distinguish that. 

You could investigate the /etc/sudoers file and see what it informs you, but I am afraid there will non-interactive installs (that don't need to prompted) that have very similar /etc/sudoers files than interactive installs (that need to be prompted), so what better check than just check, "Have terminal" ? prompt : don't prompt.

So, what exactly do you don't like about the NONINTERACTIVE variable? We could perhaps change the if NONINTERACTIVE: return 0 block in have_sudo_access to args += "-n" (non-interactive), that might be a bit cleaner and so reusing the sudo logic, so it just makes the same checks, but just doesn't prompt? (I figured I could just do an early return, because it seems that on CI you just want the sudo install anyway.) We could even just bring all this logic to the have_sudo_access with if [[ ! -t 0 ]]; then args += "-n", but having a variable has the upside that you can set it yourself manually before running the install script. (If you are really sure you don't want to be prompted anyways, even if you are on a terminal.)

MikeMcQuaid: > So, what exactly do you don't like about the NONINTERACTIVE variable?

That it's wasn't necessary before the rest of the changes in here and that I'm not convinced it will be comprehensive enough to work consistently.



> return 0 block in have_sudo_access to args += "-n" (non-interactive), that might be a bit cleaner and so reusing the sudo logic, so it just makes the same checks, but just doesn't prompt?

Yes, that makes sense to me.

hwalinga: >That it's wasn't necessary before the rest of the changes in here and that I'm not convinced it will be comprehensive enough to work consistently.

The change that this PR primarily brought is more interactivity. So, it only makes sense that if we are making the script more interactive, more fine grained control on when or when not this interactivity must be activated.

----

I think the -t 0 is  already **very** strict and comprehensive if I understand its implementation correctly. It makes the check if the file descriptor (FD 0 is stdin) is connected to a **terminal**, in which the **terminal** used to be an actual hardware device talking to the TTY kernel module. Nowadays hardware **terminals** are rarely in use, and anyone uses a **virtual terminal**, xterm, urxvt, iTerm2 (but also tmux and sshd). But these **terminal emulators** should still **emulate** this hardware behavior to the kernel to make it be a terminal (and have terminal features). Technically it is thus possible for any piece of software to **emulate** being a **terminal** even though there was no human interacting with it, however it would in that case not make any sense for that piece of software to **emulate** being a **terminal**. So, that is why -t 0 is in my opinion a very comprehensive check to know if there is indeed a human there to see the prompt and act upon it.  

In fact the check -t 0 is already in use in the install script! `if [[ -t 0 && -z "${CI-}" ]]; then wait_for_user; fi`, and also in the uninstall script. `if [[ -t 0 && -z $opt_force && -z $opt_dry_run ]]; then read -rp "Are you sure you want to uninstall Homebrew? This will remove your installed packages! [y/N] ";  [[ $REPLY == [yY]* ]] || abort; fi`. 

----

Which brings me to the realization that there already exists a variable that can track this, `CI`. Didn't know that already. It is also not set automatically, so can only be set externally. I wonder if this is documented somewhere. 

So, we might be using CI instead of  NONINTERACTIVE (and than still set it automatically). However, if the variable CI is set it also unconditionally installs in `/home/linuxbrew/.linuxbrew` (because there is check that does that) without any way to install it somewhere else, which might not be desirable. So, we either have two variables, or remove that last check. (I have now kept both variables and added that if CI is set, NONINTERACTIVE is set as well.)

MikeMcQuaid: > So, we might be using CI instead of NONINTERACTIVE (and than still set it automatically). However, if the variable CI is set it also unconditionally installs in `/home/linuxbrew/.linuxbrew` (because there is check that does that) without any way to install it somewhere else, which might not be desirable. So, we either have two variables, or remove that last check. (I have now kept both variables and added that if CI is set, NONINTERACTIVE is set as well.)

This approach makes sense to me given the explanation above, thanks 👍🏻.

The macOS CI is now failing, though.

MikeMcQuaid:
This looks good! I want to hold off merging until tomorrow so I can be around to debug and fix any issues that crop up. Nice work!

hwalinga:
Yes, so as I said I added, (1) check on -t 0 and CI variable to set NONINTERACTIVE and (2) integrated the logic of sudo with NONINTERACTIVE, if NONINTERACTIVE we still check sudo mkdir (which is the same as before this PR). Basically this check is again the same as in the original PR, but instead of checking if on MacOS, we check non-interactive instead (so more general).

install.sh
```diff
@@ -64,13 +70,20 @@ have_sudo_access() {
...
     if [[ -n "${args[*]-}" ]]; then
-      /usr/bin/sudo "${args[@]}" -l mkdir &>/dev/null
+      SUDO="/usr/bin/sudo ${args[*]}"
     else
-      /usr/bin/sudo -l mkdir &>/dev/null
+      SUDO="/usr/bin/sudo"
+    fi
+    if [[ -n "${NONINTERACTIVE-}" ]]; then
+      ${SUDO} -l mkdir &>/dev/null
+  else
```
sjackman, line 85: Indentation looks wrong here or something else is messed up.

sjackman:
```console
$ sudo -v </dev/null >/dev/null 2>/dev/null
[sudo] password for linuxbrew:
$ sudo -v <&- >&- 2>&-
[sudo] password for linuxbrew:
$ tty
/dev/pts/0
```
Here's a piece of Unix design that I never fully understood. I can close all the standard file descriptors or redirect them to `/dev/null`, and `sudo` still prompts for an interactive password. I believe it's opening the associated teletype device `/dev/pts/0` associated with my interactive shell and reading from that. How do you tell `sudo` not to do that, by normal Unix-y mechanisms, without having to specify `sudo -n`?

Where I'm going with this, is I'd expect `./install.sh </dev/stdin` not to prompt for interactive input, yet without some extra magic, it does.

install.sh
```diff
@@ -1,6 +1,12 @@
 #!/bin/bash
 set -u
 
+# Check if script is run non-interactively (e.g. CI)
+# If it is run non-interactively we should not prompt for passwords.
+if [[ ! -t 0 || -n "${CI-}" ]]; then
```
sjackman, line 6: Is there any CI system where `CI` is defined, but `test -t 0` exits successfully? If so, I already hate that CI system.

gromgit:
> I believe it's opening the associated teletype device `/dev/pts/0` associated with my interactive shell and reading from that.

It's `/dev/tty`, according to the source code and `ls -l /proc/<sudo_pid>/fd`.

> How do you tell sudo not to do that, by normal Unix-y mechanisms, without having to specify `sudo -n`?

I don't think that's possible. According to the [1.8.28 release notes](https://github.com/sudo-project/sudo/releases/tag/SUDO_1_8_28):
> Sudo no longer refuses to prompt for a password when it cannot determine the user's terminal as long as it can open `/dev/tty`. This allows sudo to function on systems where `/proc` is unavailable, such as when running in a chroot environment.

sjackman:
I learned two things today! Thank you, Adrian, for doing this research! Clearly `sudo rm -f /dev/tty` is our only reasonable path forward. But alas we need `sudo` to do that. Foiled!

install.sh
sjackman: To put that more succinctly, is `if [[ ! -t 0 ]]; then` sufficient?

install.sh
```diff
@@ -64,13 +70,20 @@ have_sudo_access() {
   local -a args
   if [[ -n "${SUDO_ASKPASS-}" ]]; then
     args=("-A")
+  elif [[ -n "${NONINTERACTIVE-}" ]]; then
+    args=("-n")
```
sjackman, line 74: I like this hack!

install.sh
MikeMcQuaid, line 85:
```suggestion
    else
```

MikeMcQuaid (approved):

install.sh
hwalinga: It most likely is sufficient to only check for `! -t 0`, but later in the code there is a check like this:

```
if [[ -t 0 && -z "${CI-}" ]]; then
  wait_for_user
fi
```

So, I thought, I'll be consistent...

MikeMcQuaid: @sjackman @hwalinga Given that I've pushed 6f37ca9 to use `NONINTERACTIVE` in more places.

install.sh
```diff
@@ -298,7 +298,10 @@ fi
 if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
  have_sudo_access
 else
-  if [[ -n "${CI-}" ]] || [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] || [[ -w "/home/linuxbrew" ]] || [[ -w "/home" ]]; then
+  if [[ -n "${NONINTERACTIVE-}" ]] ||
+     [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] ||
+     [[ -w "/home/linuxbrew" ]] ||
+     [[ -w "/home" ]]; then
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
```
hwalinga, line 305: Using NONINTERACTIVE here makes it now impossible to install linuxbrew non-interactively in the home folder. Is that really what we want?

With this change, if you are non-interactively installing linuxbrew and don't have elevated permissions, the script will fail. Before this change, it would just install in the home folder with no problem. (You can even force that behavior with `export SUDO_ASKPASS=false`. 

If you just want to get rid of the CI variable, I would just completely remove this check. It doesn't makes sense to force install in the `/home/linuxbrew` if run NONINTERACTIVE, the script should decide that itself. 

I think the reason the check used to be here is that it was very likely that the script would block on the else block after this. So, if on CI, you didn't want to run that piece of code. But now we are more clever in the else block and won't block anymore. But since people weren't bothered setting the CI variable (undocumented?), still a lot of CI pipelines broke when we made the else block more interactive.

MikeMcQuaid: > Using NONINTERACTIVE here makes it now impossible to install linuxbrew non-interactively in the home folder. Is that really what we want?

Yes, we should non-interactively always install in the default, supported location if the user isn't being shown the `recommended` message.

> But since people weren't bothered setting the CI variable (undocumented?), still a lot of CI pipelines broke when we made the else block more interactive.

Most CI pipelines do set this variable, though?

hwalinga: >Yes, we should non-interactively always install in the default, supported location if the user isn't being shown the recommended message.

That is fine in that case!

>Most CI pipelines do set this variable, though?

I didn't knew that. Anyway, if you dig into the comments of #341 apparently not all do. For example, this person found that it installed in the home folder, so he did not had the CI variable set in that case. https://github.com/Homebrew/install/pull/341#issuecomment-733915458

hwalinga:
>How do you tell sudo not to do that, by normal Unix-y mechanisms, without having to specify sudo -n?

An interesting question. As it appears `sudo` indeed uses /dev/tty which I guess it gets assigned by the parent process, and even if you use things like nohup and disown, it still is going to block/ask for the password. `su` however doesn't do this.

MikeMcQuaid:
Thanks again @hwalinga!

sjackman:
Thank you, Hielke! 🏆

sjackman:
I like that `bash install.sh </dev/null` and `yes | bash install.sh` now work for a non-interactive installation.

PR #362, opened 2020-11-29
Fix default WSL user

Vampire:


MikeMcQuaid:
> **GitHub Actions CI / tests (windows-latest) (pull_request) ** Cancelled after 360m — tests (windows-latest)

😭

Vampire:
As far as this PR is concerned this doesn't matter, it is already after the WSL installation and user setup and user using.
That it then hangs later when asking for input is not my cup of tea, I don't know your project or what it should do there. ;-)
It seems what you call waits for interactive input and after the max time 6h GHA just cancels the workflow:
```
Run /bin/bash -c "$(cat install.sh)"
  /bin/bash -c "$(cat install.sh)"
  shell: D:\a\_temp\wsl-shell-wrapper\wsl-bash.BAT {0}
==> Select the Homebrew installation directory
- Enter your password to install to /home/linuxbrew/.linuxbrew (recommended)
- Press Control-D to install to /home/runner/.linuxbrew
- Press Control-C to cancel installation
Terminate batch job (Y/N)? 
^C
```

MikeMcQuaid:
> That it then hangs later when asking for input is not my cup of tea, I don't know your project or what it should do there.

Thanks 👍🏻

PR #365, opened 2020-12-03
fixed root user pre-flight check to check $EUID

HariSekhon:
The install script currently aborts if you run it as a user via su or similar from root because it checks `$UID` instead of `$EUID` which is more accurate of the currently executing user.

This patch attempts to use the correct `$EUID` and falls back to `$UID` (this is run in bash which should always have `$EUID`).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -321,7 +321,7 @@ else
   HOMEBREW_REPOSITORY="${HOMEBREW_PREFIX}/Homebrew"
 fi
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @HariSekhon!

HariSekhon:
You're welcome, love the project, keep up the good work! :)

PR #366, opened 2020-12-03
Fix noninteractive prefix selection

calavera:
Check if there is sudo access to pick HOMEBREW_PREFIX, otherwise, the script tries to create the directories in the homebrew user home regardless of whether you have sudo access or not.

This fixes an issue introduced by https://github.com/Homebrew/install/commit/6f37ca94af073c2971efbed8aa293322aa171f26#diff-043df5bdbf6639d7a77e1d44c5226fd7371e5259a1e4df3a0dd5d64c30dca44f.

Before that commit, in noninteractive mode, the install script would show the directory selection options but it'd pick the current user's home directory when you don't have sudo access. After that commit, it always tries to create /home/linuxbrew regardless of whether you have sudo access or not.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -298,11 +298,12 @@ fi
 if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
  have_sudo_access
 else
-  if [[ -n "${NONINTERACTIVE-}" ]] ||
-     [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] ||
-     [[ -w "/home/linuxbrew" ]] ||
-     [[ -w "/home" ]]; then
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
We intentionally want to only install into the default prefix location for a non-interactive installation because that's the only location we recommend or support.

MikeMcQuaid:
We'd accept a PR to error-out in a better way in that respect but we do not wish to install something that immediately fails `brew doctor` without being reinstalled elsewhere.

calavera:
Ok. I don't really understand the rationale because I lack a lot of context, so I'll keep using my modified script for the time being.

PR #368, opened 2020-12-05
Only fech brew master branch during installation

lucianosousa:
There is no need to fetch all branches during the installation process, as we can see on line 599, only the master is being used.


MikeMcQuaid:
I don't think it does any harm here to fetch non-master branches as `brew update` will do so anyway.

PR #371, opened 2020-12-11
install.sh: preload homebrew/core before brew update

EricFromCanada:
If the installer is run on a system that has outdated `curl` or `git` as determined by `brew.sh`, `cmd/update.sh` will attempt to install them before `homebrew/core` is tapped with `git_init_if_necessary()`. This change has the installer pull down the tap first, allowing the automatic install of either to (potentially) succeed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -54,7 +54,7 @@ else
   GROUP="$(id -gn)"
   TOUCH="/bin/touch"
 fi
-BREW_REPO="https://github.com/Homebrew/brew"
+HOMEBREW_BREW_GIT_REMOTE="https://github.com/Homebrew/brew"
 
 # TODO: bump version when new macOS is released or announced
 MACOS_NEWEST_UNSUPPORTED="12.0"
@@ -176,10 +176,6 @@ major_minor() {
   echo "${1%%.*}.$(x="${1#*.}"; echo "${x%%.*}")"
 }
 
-if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
-  macos_version="$(major_minor "$(/usr/bin/sw_vers -productVersion)")"
-fi
-
 version_gt() {
   [[ "${1%.*}" -gt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -gt "${2#*.}" ]]
 }
@@ -232,7 +228,7 @@ file_not_grpowned() {
 }
 
 # Please sync with 'test_ruby()' in 'Library/Homebrew/utils/ruby.sh' from Homebrew/brew repository.
-test_ruby () {
+test_ruby() {
   if [[ ! -x $1 ]]
   then
     return 1
@@ -307,10 +303,10 @@ EOABORT
 fi
 
 if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
- have_sudo_access
+  have_sudo_access
 else
   if [[ -n "${NONINTERACTIVE-}" ]] ||
-     [[ -w "$HOMEBREW_PREFIX_DEFAULT" ]] ||
+     [[ -w "${HOMEBREW_PREFIX_DEFAULT}" ]] ||
      [[ -w "/home/linuxbrew" ]] ||
      [[ -w "/home" ]]; then
     HOMEBREW_PREFIX="$HOMEBREW_PREFIX_DEFAULT"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -600,6 +600,12 @@ ohai "Downloading and installing Homebrew..."
 
   execute "ln" "-sf" "${HOMEBREW_REPOSITORY}/bin/brew" "${HOMEBREW_PREFIX}/bin/brew"
 
+  if [[ ! -d "${HOMEBREW_CORE}" ]] && [[ -x "/usr/bin/curl" ]] && [[ -x "/usr/bin/tar" ]]; then
+    execute "/bin/mkdir" "-p" "${HOMEBREW_CORE}"
+    execute "/usr/bin/curl" "-fsSL" "${HOMEBREW_CORE_GIT_REMOTE}/tarball/master" | \
+      execute "/usr/bin/tar" "xm" "-C" "${HOMEBREW_CORE}" "--strip" "1"
+  fi
```
MikeMcQuaid, line 607: I wonder if these should allow any of these tools in the `PATH` to be used?

Also, given we use `git` immediately above: why not clone it here instead of using `curl`?

EricFromCanada: Seems to work, but it does end up showing the donation message twice. That could be handled in another PR, though.
```
...
 * [new tag]             2.6.0                                             -> 2.6.0
 * [new tag]             2.6.1                                             -> 2.6.1
HEAD is now at a6800dc85 Merge pull request #9514 from Homebrew/sync-triage-config
==> Downloading https://homebrew.bintray.com/bottles-portable-ruby/portable-ruby-2.6.3_2.yosemite.bottle.tar.gz
######################################################################## 100.0%
==> Pouring portable-ruby-2.6.3_2.yosemite.bottle.tar.gz
==> Homebrew is run entirely by unpaid volunteers. Please consider donating:
  https://github.com/Homebrew/brew#donations
Updating Homebrew...
==> Tapping homebrew/core
Cloning into '/usr/local/Homebrew/Library/Taps/homebrew/homebrew-core'...
remote: Enumerating objects: 40, done.
remote: Counting objects: 100% (40/40), done.
remote: Compressing objects: 100% (28/28), done.
remote: Total 855006 (delta 20), reused 26 (delta 12), pack-reused 854966
Receiving objects: 100% (855006/855006), 340.10 MiB | 9.98 MiB/s, done.
Resolving deltas: 100% (576403/576403), done.
Checking out files: 100% (5650/5650), done.
Tapped 2 commands and 5374 formulae (5,674 files, 373.1MB).
Already up-to-date.
==> Installation successful!

==> Homebrew has enabled anonymous aggregate formulae and cask analytics.
Read the analytics documentation (and how to opt-out) here:
  https://docs.brew.sh/Analytics
No analytics data has been sent yet (or will be during this `install` run).

==> Homebrew is run entirely by unpaid volunteers. Please consider donating:
  https://github.com/Homebrew/brew#donations

==> Next steps:
- Run `brew help` to get started
- Further documentation: 
    https://docs.brew.sh
```

EricFromCanada: I think the advantage of the original method below is that it avoids running any Ruby code and the `vendor-install` step.
```
  if [[ ! -d "${HOMEBREW_CORE}" ]] && [[ -x "/usr/bin/curl" ]] && [[ -x "/usr/bin/tar" ]]; then
    execute "/bin/mkdir" "-p" "${HOMEBREW_CORE}"
    execute "/usr/bin/curl" "-fsSL" "${HOMEBREW_CORE_GIT_REMOTE}/tarball/master" | \
      execute "/usr/bin/tar" "xm" "-C" "${HOMEBREW_CORE}" "--strip" "1"
  fi
```
Or did you mean the installer should run the `git` commands directly as it does for `Homebrew/brew`?

MikeMcQuaid: > Seems to work, but it does end up showing the donation message twice.

If it's already been shown you can set the donation `gitconfig` as has been done with the analytics.

> Or did you mean the installer should run the `git` commands directly as it does for `Homebrew/brew`?

Yeh, this. What downsides are there here that I'm missing?

EricFromCanada:
Reworked things to clone the tapping of `homebrew/core` in the same manner as `git_init_if_necessary()` in `update.sh`. To fix the duplicate donation message, I added a `--quiet` flag to `brew update`, which is currently a no-op, but with Homebrew/brew#10021 it'll suppress the message in `update-report.rb`.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @EricFromCanada!

PR #509, opened 2020-12-25
Fix: Xcode Command Line Tools checking

toco8:
Command Line Tools installation is executed even though Xcode 12 is installed.
Xcode 12 includes Command Line Tools.
It's installed in `/Applications/Xcode.app/Contents/Developer`.
So I added IF statement to check whether Command Line Tools for Xcode is exists.
It occurred in macOS Catalina.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -185,7 +185,8 @@ should_install_command_line_tools() {
   fi
 
   if version_gt "$macos_version" "10.13"; then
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
These are not the same as the non-Xcode CLT and some formulae need these specifically so this is intentional. Sorry!

PR #510, opened 2020-12-26
uninstall: change short dry-run flag to `-n`

carlocab:
`-n` is more commonly recognised as the short form of the `--dry-run`
flag. This is the case, for example, with `git` and `brew`.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -170,7 +170,7 @@ Usage: $0 [options]
                      Skips removal of HOMEBREW_CACHE and HOMEBREW_LOGS.
     -f, --force      Uninstall without prompting.
     -q, --quiet      Suppress all output.
-    -d, --dry-run    Simulate uninstall but don't remove anything.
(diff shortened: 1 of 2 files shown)


uninstall.sh
```diff
@@ -183,7 +183,7 @@ while [[ $# -gt 0 ]]; do
     --skip-cache-and-logs) opt_skip_cache_and_logs=1;;
     -f|--force) opt_force=1;;
     -q|--quiet) opt_quiet=1;;
-    -d|--dry-run) opt_dry_run=1;;
+    -n|--dry-run) opt_dry_run=1;;
```
MikeMcQuaid, line 186:
```suggestion
    -n|-d|--dry-run) opt_dry_run=1;;
```

For backwards compatibility. We don't want to break existing uses of this script.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @carlocab!

PR #511, opened 2020-12-28
install.sh: early exit on unsupported OS

gromgit:
This should prevent a recurrence of https://github.com/Homebrew/discussions/discussions/363#discussioncomment-246193


MikeMcQuaid (commented):
Thanks for the PR! Could this be an `if`, `elsif`, `else`? The `: ;;` in macOS isn't super readable 😅

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @gromgit!

PR #513, opened 2020-12-31
install.sh: show shellenv instructions on Apple M1

carlocab:
Fixes https://github.com/Homebrew/brew/issues/10191.

Ok, this seems to work. At the end of the install, it should show
```
==> Next steps:
- Run `brew help` to get started
- Further documentation:
    https://docs.brew.sh
- Add Homebrew to your PATH in /Users/carlocab/.zprofile:
    echo 'eval $(/opt/homebrew/bin/brew shellenv)' >> /Users/carlocab/.zprofile
    eval $(/opt/homebrew/bin/brew shellenv)
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -656,7 +656,7 @@ echo "- Run \`brew help\` to get started"
 echo "- Further documentation: "
 echo "    ${tty_underline}https://docs.brew.sh${tty_reset}"
 
-if [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
+if [[ "$UNAME_MACHINE" == "arm64" ]] || [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
   case "$SHELL" in
     */bash*)
       if [[ -r "$HOME/.bash_profile" ]]; then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -672,7 +672,18 @@ if [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
       shell_profile="$HOME/.profile"
       ;;
   esac
+fi
+
+# HOMEBREW_PREFIX is not in PATH on ARM macOS, so remind users to set up their shell
+if [[ -z "${HOMEBREW_ON_LINUX-}" ]] && [[ "$UNAME_MACHINE" == "arm64" ]]; then
+  cat <<EOS
+- Add Homebrew to your ${tty_bold}PATH${tty_reset} in ${tty_underline}${shell_profile}${tty_reset}:
```
MikeMcQuaid, line 680: Maybe worth checking it's not already in the `PATH`?

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks @carlocab! One suggestion but happy to merge without if you disagree.

install.sh
carlocab: No, I think there's no need. 

Only users who already understand how to use the `PATH` variable are bothered by duplicate `PATH` entries, even if there's no harm to them. Being informed about `brew shellenv` is helpful even for users who know to set their `PATH` too, as it reminds them to set `MANPATH` (and maybe the slightly less useful `INFOPATH`).

richiksc:
Is it even necessary to manually set `MANPATH`? I did not know about `brew shellenv` and instead I manually added `/opt/homebrew/bin` to my `PATH` in `.zshrc`. Apparently `man` on macOS automatically searches through `PATH` and maps it to associated `man` directories to build a`MANPATH` automatically.  Some even recommend against manually setting `MANPATH` on macOS as that overrides the automatic behavior.

Sources:
1. https://scriptingosx.com/2017/05/where-paths-come-from/
2. `man` man page

MikeMcQuaid:
Thanks @carlocab!

carlocab:
@richiksc Hm, you're right, I guess it's not necessary. However, it doesn't hurt, as `brew shellenv` appends `:` to `MANPATH`, which tells `man` to search in its standard paths when it can't find the appropriate manpage specified explicitly in `MANPATH`.

lukegavahan:
Hi @carlocab, I just wanted to let you know that upon installing Homebrew on my Apple M1 Mac, I did not see any shellenv instructions.

PR #517, opened 2021-01-05
Added to check and add Homebrew in PATH with consent for M1 Macs

jiteshpubreja:
The Script will ask user if they want to add Homebrew to the path automatically or not.
If not, they can copy the commands manually and go on as usual.

In case of Homebrew already exists in the path (due to previous installation or something), This will do nothing.

and moved the "Next steps" after PATH Instructions as pointed in #516


MikeMcQuaid:
I'd rather not add a prompt here but I'd open to the other changes.

PR #521, opened 2021-01-07
install: fix next steps messaging

carlocab:
- make sure `shell_profile` is defined before use
- keep message order consistent between macOS and Linux

Fixes https://github.com/Homebrew/discussions/discussions/464.


MikeMcQuaid:
Thanks @carlocab!

PR #526, opened 2021-01-22
add functionality to lock installation to specific git refs

andreineculau:
rationalé: determinism.

I'm installing homebrew/linuxbrew in the CI, and right after installing brew, I do reset to specific git refs of brew/core tap. But that's not enough when installing with brewed ruby/git/curl, which would pick up formulas/bottles as per origin/master, which in turn means that the end state, after running `brew install` commands, is some packages are newer than they should be.

Some documentation would be nice, but maybe not. Up to you. Let me know.


MikeMcQuaid:
I'd suggest this be something you build yourself, we don't wish to support people using old revisions, sorry.

PR #528, opened 2021-01-27
install: widen chown scope

gromgit:
Everything under `$HOMEBREW_REPOSITORY` and `$HOMEBREW_CACHE` is owned by Homebrew, so any different ownerships therein should be fixed at install time.

Addresses https://github.com/Homebrew/discussions/discussions/140#discussioncomment-313198


MikeMcQuaid (approved):

MikeMcQuaid:
Makes sense, thanks again @gromgit!

PR #530, opened 2021-02-06
install.sh: tell the user that we're going to run sudo

woodruffw:
This is a pretty minor change: someone I work with mentioned being surprised that the Homebrew install script immediately prompted them for a password (without telling them which password) with no other information, so this just tells the user that we're testing for `sudo` and that they *may* get prompted.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -85,8 +85,6 @@ tty_bold="$(tty_mkbold 39)"
 tty_reset="$(tty_escape 0)"
 
 have_sudo_access() {
-  echo "We're testing for sudo access. You may be prompted for your password."
-
   local -a args
   if [[ -n "${SUDO_ASKPASS-}" ]]; then
     args=("-A")
@@ -305,6 +303,8 @@ You must install cURL before installing Homebrew. See:
 EOABORT
 )"
 fi
+
+echo "We're testing for sudo access. You may be prompted for your password."
 
 if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
   have_sudo_access
--- a/install.sh
+++ b/install.sh
@@ -304,6 +304,7 @@ EOABORT
 )"
 fi
 
+# shellcheck disable=SC2016
 ohai 'Checking for `sudo` access (which may request your password).'
 
 if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
This was on Linux presumably at https://github.com/Homebrew/install/blob/master/install.sh#L308

I think it'd be worth outputting this in a more specific place; as-is it's going to be output every time access is changed regardless of whether the password is prompted for or not.

woodruffw:
> This was on Linux presumably at https://github.com/Homebrew/install/blob/master/install.sh#L308

Actually, it was macOS! That line tests `HOMEBREW_ON_LINUX` for emptiness (if I'm reading it right), so it should trigger on macOS. This might also apply to Linux, but I haven't checked.

> I think it'd be worth outputting this in a more specific place; as-is it's going to be output every time access is changed regardless of whether the password is prompted for or not.

Yep, I agree. I'll place it somewhere else to avoid spitting it out a whole bunch.

install.sh
```diff
@@ -304,6 +304,8 @@ EOABORT
 )"
 fi
 
+echo "We're testing for sudo access. You may be prompted for your password."
```
MikeMcQuaid, line 307:
```suggestion
ohai 'Checking for `sudo` access (which may request your password).'
```

MikeMcQuaid, line 307:
```suggestion
ohai 'Checking for `sudo` access (which may request your password).'
```

MikeMcQuaid (approved):
Thanks @woodruffw!

woodruffw:
No problem, thanks for the turnaround 🙂

install.sh
```diff
@@ -304,6 +304,8 @@ EOABORT
 )"
 fi
 
+ohai 'Checking for `sudo` access (which may request your password).'
```
MikeMcQuaid, line 307:
```suggestion
# shellcheck disable=SC2016
ohai 'Checking for `sudo` access (which may request your password).'
```

MikeMcQuaid: (would commit to your branch but I don't think I have rights)

woodruffw:
I _think_ this is good to go; the CI failure looks like an API timeout on GitHub Actions rather than a test failure.

MikeMcQuaid:
Thanks @woodruffw!

PR #534, opened 2021-02-26
Read Homebrew remote URLs from environment variables and prompt for non-default values

XuehaiPan:
The current install script hardcodes `HOMEBREW_{BREW,CORE}_GIT_REMOTE`. Users can only get Homebrew from GitHub during the first installation. Currently the size of `homebrew-core` repo is up to  371.09 MB. Sometimes it can be really slow during the installation process.

Only after the installation is complete, the user can change the remote of the repositories.
```bash
# Users get Homebrew from GitHub

# Users can change the remote after long time for `git clone`
export HOMEBREW_BREW_GIT_REMOTE="https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/brew.git"
export HOMEBREW_CORE_GIT_REMOTE="https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/homebrew-core.git"
brew update
```

Currently, I hack the install script when getting Homebrew. I hope it can be officially supported.
```bash
# Before
export HOMEBREW_BREW_GIT_REMOTE="https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/brew.git"
export HOMEBREW_CORE_GIT_REMOTE="https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/homebrew-core.git"
/bin/bash -c "$(
    curl -fsSL https://github.com/Homebrew/install/raw/master/install.sh |
    sed -E 's#^(\s*)(HOMEBREW_(BREW|CORE)_GIT_REMOTE)=(.*)$#\1\2="${\2:-\4}"#'
)"

# After
export HOMEBREW_BREW_GIT_REMOTE="https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/brew.git"
export HOMEBREW_CORE_GIT_REMOTE="https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/homebrew-core.git"
```

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -83,8 +83,8 @@ jobs:
 
       - run: /bin/bash uninstall.sh -f >/dev/null
 
-      - name: Install Homebrew with arguments
-        # Use Git Protocol for default remotes
+      - name: Install Homebrew with non-default remotes
+        # Use the default remotes but with Git Protocol
         run: |
           HOMEBREW_BREW_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/brew"
           if [ "${{ runner.os }}" = "macOS" ]; then
@@ -92,11 +92,15 @@ jobs:
           else
             HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/linuxbrew-core"
           fi
-          /bin/bash -c "$(cat install.sh)" \
-            --brew-git-remote "${HOMEBREW_BREW_DEFAULT_GIT_REMOTE/%https/git}" \
-            --core-git-remote "${HOMEBREW_CORE_DEFAULT_GIT_REMOTE/%https/git}"
+          export HOMEBREW_BREW_GIT_REMOTE="${HOMEBREW_BREW_DEFAULT_GIT_REMOTE/%https/git}"
+          export HOMEBREW_CORE_GIT_REMOTE="${HOMEBREW_CORE_DEFAULT_GIT_REMOTE/%https/git}"
+          /bin/bash -c "$(cat install.sh)"
 
-      - run: /bin/bash uninstall.sh -f >/dev/null
+      - run: brew config
+
+      - run: |
+          /bin/bash uninstall.sh -f >/dev/null
+          unset HOMEBREW_{BREW,CORE}{,_DEFAULT}_GIT_REMOTE
 
       - run: /bin/bash -c "$(cat install.sh)"
 
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -83,6 +83,21 @@ jobs:
 
       - run: /bin/bash uninstall.sh -f >/dev/null
 
+      - name: Install Homebrew with arguments
+        # Use Git Protocol for default remotes
+        run: |
+          HOMEBREW_BREW_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/brew"
+          if [ "${{ runner.os }}" = "macOS" ]; then
+            HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/homebrew-core"
+          else
+            HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/linuxbrew-core"
+          fi
+          /bin/bash -c "$(cat install.sh)" \
+            --brew-git-remote "${HOMEBREW_BREW_DEFAULT_GIT_REMOTE/%https/git}" \
+            --core-git-remote "${HOMEBREW_CORE_DEFAULT_GIT_REMOTE/%https/git}"
+
+      - run: /bin/bash uninstall.sh -f >/dev/null
+
       - run: /bin/bash -c "$(cat install.sh)"
 
       - name: Uninstall and reinstall with sudo NOPASSWD
--- a/README.md
+++ b/README.md
@@ -10,6 +10,21 @@ More installation information and options: https://docs.brew.sh/Installation.
 
 If running Linux or WSL, [there are some pre-requisite packages to install](https://docs.brew.sh/Homebrew-on-Linux#requirements).
 
+You can set `HOMEBREW_BREW_GIT_REMOTE` and/or `HOMEBREW_CORE_GIT_REMOTE` in your shell environment to use custom Git mirrors to speed up brew update and brew tap.
+
+```bash
+export HOMEBREW_BREW_GIT_REMOTE="..."  # put your mirror URL of Homebrew/brew Git remote here
+export HOMEBREW_CORE_GIT_REMOTE="..."  # put your mirror URL of Homebrew/core Git remote here
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+```
+
+| Variable                 | Default                                                                                                |
+| ------------------------ | ------------------------------------------------------------------------------------------------------ |
+| HOMEBREW_BREW_GIT_REMOTE | macOS / Linux: https://github.com/Homebrew/brew                                                        |
+| HOMEBREW_CORE_GIT_REMOTE | macOS: https://github.com/Homebrew/homebrew-core<br/>Linux: https://github.com/Homebrew/linuxbrew-core |
+
+The default Git remote will be used if the corresponding environment variable is unset.
+
 ## Uninstall Homebrew
 
 ```bash
--- a/README.md
+++ b/README.md
@@ -10,18 +10,13 @@ More installation information and options: https://docs.brew.sh/Installation.
 
 If running Linux or WSL, [there are some pre-requisite packages to install](https://docs.brew.sh/Homebrew-on-Linux#requirements).
 
-You can set `HOMEBREW_BREW_GIT_REMOTE` and/or `HOMEBREW_CORE_GIT_REMOTE` in your shell environment to use custom Git mirrors to speed up brew update and brew tap.
+You can set `HOMEBREW_BREW_GIT_REMOTE` and/or `HOMEBREW_CORE_GIT_REMOTE` in your shell environment to use geolocalized Git mirrors to speed up Homebrew's installation with this script and, after installation, `brew update`.
 
 ```bash
-export HOMEBREW_BREW_GIT_REMOTE="..."  # put your mirror URL of Homebrew/brew Git remote here
-export HOMEBREW_CORE_GIT_REMOTE="..."  # put your mirror URL of Homebrew/core Git remote here
+export HOMEBREW_BREW_GIT_REMOTE="..."  # put your Git mirror of Homebrew/brew here
+export HOMEBREW_CORE_GIT_REMOTE="..."  # put your Git mirror of Homebrew/homebrew-core here
 /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
 ```
-
-| Variable                 | Default                                                                                                |
-| ------------------------ | ------------------------------------------------------------------------------------------------------ |
-| HOMEBREW_BREW_GIT_REMOTE | macOS / Linux: https://github.com/Homebrew/brew                                                        |
-| HOMEBREW_CORE_GIT_REMOTE | macOS: https://github.com/Homebrew/homebrew-core<br/>Linux: https://github.com/Homebrew/linuxbrew-core |
 
 The default Git remote will be used if the corresponding environment variable is unset.
 
--- a/install.sh
+++ b/install.sh
@@ -36,7 +36,7 @@ if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
     HOMEBREW_REPOSITORY="${HOMEBREW_PREFIX}/Homebrew"
   fi
   HOMEBREW_CACHE="${HOME}/Library/Caches/Homebrew"
-  HOMEBREW_CORE_GIT_REMOTE="https://github.com/Homebrew/homebrew-core"
+  HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/homebrew-core"
 
   STAT="stat -f"
   CHOWN="/usr/sbin/chown"
@@ -50,7 +50,7 @@ else
   # and ~/.linuxbrew (which is unsupported) if run interactively.
   HOMEBREW_PREFIX_DEFAULT="/home/linuxbrew/.linuxbrew"
   HOMEBREW_CACHE="${HOME}/.cache/Homebrew"
-  HOMEBREW_CORE_GIT_REMOTE="https://github.com/Homebrew/linuxbrew-core"
+  HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/linuxbrew-core"
 
   STAT="stat --printf"
   CHOWN="/bin/chown"
@@ -58,7 +58,10 @@ else
   GROUP="$(id -gn)"
   TOUCH="/bin/touch"
 fi
-HOMEBREW_BREW_GIT_REMOTE="https://github.com/Homebrew/brew"
+HOMEBREW_BREW_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/brew"
+
+HOMEBREW_BREW_GIT_REMOTE="${HOMEBREW_BREW_GIT_REMOTE:-"${HOMEBREW_BREW_DEFAULT_GIT_REMOTE}"}"
+HOMEBREW_CORE_GIT_REMOTE="${HOMEBREW_CORE_GIT_REMOTE:-"${HOMEBREW_CORE_DEFAULT_GIT_REMOTE}"}"
 
 # TODO: bump version when new macOS is released or announced
 MACOS_NEWEST_UNSUPPORTED="12.0"
--- a/install.sh
+++ b/install.sh
@@ -316,8 +316,8 @@ while [[ "$#" -gt 0 ]]; do
   esac
 done
 # Use the default remotes if not specified.
-HOMEBREW_BREW_GIT_REMOTE="${HOMEBREW_BREW_GIT_REMOTE:-"${HOMEBREW_BREW_DEFAULT_GIT_REMOTE}"}"
-HOMEBREW_CORE_GIT_REMOTE="${HOMEBREW_CORE_GIT_REMOTE:-"${HOMEBREW_CORE_DEFAULT_GIT_REMOTE}"}"
+export HOMEBREW_BREW_GIT_REMOTE="${HOMEBREW_BREW_GIT_REMOTE:-"${HOMEBREW_BREW_DEFAULT_GIT_REMOTE}"}"
+export HOMEBREW_CORE_GIT_REMOTE="${HOMEBREW_CORE_GIT_REMOTE:-"${HOMEBREW_CORE_DEFAULT_GIT_REMOTE}"}"
 
 if [[ -n "${HOMEBREW_ON_LINUX-}" ]] && no_usable_ruby && outdated_glibc
 then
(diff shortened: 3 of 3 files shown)


XuehaiPan:
As a reminder, I think such changes may cause security issues. For example, set `HOMEBREW_{BREW,CORE}_GIT_REMOTE` to repos that irrelevant homebrew. And a `sudo` passward is requested during installation. But in the meanwhile, `brew update` command also do not verify `HOMEBREW_{BREW,CORE}_GIT_REMOTE` and arbitrarily changes the remotes for homebrew.

MikeMcQuaid:
> As a reminder, I think such changes may cause security issues.

I agree. I think this is fine once Homebrew is installed from the official repositories but I don't like the idea of environment variables silently changing the installation location. If this was documented and required e.g. command-line flags or something that would be preferable to me.

XuehaiPan:
Parse Homebrew repository remote URLs from command line arguments if presented.

```bash
# Before
export HOMEBREW_BREW_GIT_REMOTE="https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/brew.git"
export HOMEBREW_CORE_GIT_REMOTE="https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/homebrew-core.git"
/bin/bash -c "$(
    curl -fsSL https://github.com/Homebrew/install/raw/master/install.sh |
    sed -E 's#^(\s*)(HOMEBREW_(BREW|CORE)_GIT_REMOTE)=(.*)$#\1\2="${\2:-\4}"#'
)"

# After
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)" \
    --brew-git-remote https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/brew.git \
    --core-git-remote https://mirrors.tuna.tsinghua.edu.cn/git/homebrew/homebrew-core.git
```

install.sh
```diff
@@ -50,15 +50,61 @@ else
...
   TOUCH="/bin/touch"
 fi
-HOMEBREW_BREW_GIT_REMOTE="https://github.com/Homebrew/brew"
+HOMEBREW_BREW_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/brew"
+
+# Parse remote URLs of Homebrew repositories from command line arguments if presented.
+parse_remote_urls_from_arguments() {
+  verify_git_remote_url() {
+    # Empty string or valid Git remote URL
+    [[ "$1" =~ ^(((file:(//)?)|((((git|ssh|http(s)?):(//)?)?)(([_a-zA-Z0-9]+@)?)([_a-zA-Z0-9.]+)((:[0-9]+)?)(:?)))([_a-zA-Z0-9./\-]+)(\.git)(/)?)?$ ]]
```
MikeMcQuaid, line 67: I admire the thoroughness here but I don't think we need to validate this.

install.sh
```diff
@@ -50,15 +50,61 @@ else
...
+    esac
+  done
+  if ! verify_git_remote_url "${HOMEBREW_BREW_GIT_REMOTE}"; then
+    echo "Invalid remote URL: HOMEBREW_BREW_GIT_REMOTE=${HOMEBREW_BREW_GIT_REMOTE}" >&2; exit 1
+  fi
+  if ! verify_git_remote_url "${HOMEBREW_CORE_GIT_REMOTE}"; then
+    echo "Invalid remote URL: HOMEBREW_CORE_GIT_REMOTE=${HOMEBREW_CORE_GIT_REMOTE}" >&2; exit 1
+  fi
+  set -u
+}
```
MikeMcQuaid, line 95: I wonder if we could consider using just the first and second argument for these remotes unconditionally? CC @Homebrew/maintainers for any thoughts on all this.

install.sh
```diff
@@ -606,6 +652,10 @@ ohai "Downloading and installing Homebrew..."
 (
   cd "${HOMEBREW_REPOSITORY}" >/dev/null || return
 
+  if [[ "$HOMEBREW_BREW_DEFAULT_GIT_REMOTE" != "$HOMEBREW_BREW_GIT_REMOTE" ]]; then
+    echo "HOMEBREW_BREW_GIT_REMOTE set: using $HOMEBREW_BREW_GIT_REMOTE for Homebrew/brew Git remote URL."
```
MikeMcQuaid, line 656: 👏🏻 for printing this. Another option to consider would be we prompt if this is non-default and otherwise use environment variables?

MikeMcQuaid (commented):
Great work here, sorry to be jumping around with approaches!

install.sh
XuehaiPan: This will only prompt the user that he or she is using a non-default URL, nothing more.

The variables from environment has been cleared at line 98:
```bash
unset HOMEBREW_{BREW,CORE}_GIT_REMOTE  # unset variables from environment
```

`HOMEBREW_{BREW,CORE}_GIT_REMOTE` is non-default, only when the user passes command line options to the script.

install.sh
XuehaiPan: The user needs to pass two URLs as unconditional arguments. Even if he or she only wants to mirror one of the repo (`brew` or `brew-core`), not both of them.

sjackman:
Thanks for this work, @XuehaiPan. Check out the issue https://github.com/Homebrew/install/issues/523 `Speeding up install times` if you haven't yet seen it.

install.sh
MikeMcQuaid: Yeh, my thinking is we could use environment variables (your original idea) but stop the installation and prompt if they are set to non-default values? Thoughts?

install.sh
```diff
@@ -262,6 +262,60 @@ outdated_glibc() {
...
+        warn "Unrecognized option: '${opt}'"; usage 1;;
+    esac
+  done
+}
+
+unset HOMEBREW_{BREW,CORE}_GIT_REMOTE  # unset variables from environment
+if [[ "$(basename -- "$0")" != "bash" && "$(basename -- "$0")" != "install.sh" ]]; then
+  # The script is invoked by `bash -c "$(cat install.sh)" [--brew-git-remote URL] [--core-git-remote URL]`
+  # Unshift the first argument back to the argument list
+  parse_remote_urls_from_arguments "$0" "$@"
```
maxim-belkin, line 311: you also could `set -- "$0" "$@"` here instead and then use `parse_remote_urls_from_arguments "$@"` universally (outside of if/else).

install.sh
```diff
@@ -262,6 +262,60 @@ outdated_glibc() {
...
+  exit "${1:-0}"
+}
+
+# Parse remote URLs of Homebrew repositories from command line arguments if presented.
+parse_remote_urls_from_arguments() {
+  while [[ "$#" -gt 0 ]]; do
+    opt="$1"
+    shift  # expose the next argument
+    case "$opt" in
+      --) break 2 ;;
```
maxim-belkin, line 290: Two questions:
1. In what situations do we need this? This script doesn't pass control to any other script.
2. `break` exits `for`, `while`, and `until` loops. Why do you break two levels here?

install.sh
```diff
@@ -262,6 +262,59 @@ outdated_glibc() {
...
+  exit "${1:-0}"
+}
+
+# Parse remote URLs of Homebrew repositories from command line arguments if presented.
+unset HOMEBREW_{BREW,CORE}_GIT_REMOTE  # unset variables from environment at first
+# The value of the script basename in different situations:
+#   - "bash":       invoked by `bash -c` with no arguments.
+#   - "install.sh": invoked by bash directly.
+#   - otherwise:    invoked by `bash -c` with at least one argument.
+if [[ "$(basename -- "$0")" != "bash" && "$(basename -- "$0")" != "install.sh" ]]; then
```
XuehaiPan, line 290: `bash -c 'basename -- "$0"'` returns `bash`, but `sh -c 'basename -- "$0"'` returns `sh` and `zsh` for `zsh`, etc. I haven't found an approach to write universal codes for all POSIX-compliant shells. I assume users always use `bash`.

install.sh
XuehaiPan: I prefer to make everything automated once the sudo privilege is checked. The installation process will be interrupted for non-default remotes if we prompt and wait. There will be three different locations that interrupt the process and wait for user input. One for sudo privilege check at the beginning, and two for remote URL confirmation in the intermediate.

install.sh
carlocab: The shebang is `#!/bin/bash`, plus the code snippet for installing `brew` tells users to run the script using `/bin/bash`, so I think that's a safe assumption.

sjackman: `install.sh` requires specifically BASH and disallows other shells.

install.sh
```diff
@@ -262,6 +262,59 @@ outdated_glibc() {
...
+  # The script is invoked by `bash -c "$(cat install.sh)" ...` with at least one argument.
+  # Unshift the first argument back to the argument list
+  set -- "$0" "$@"
+fi
+while [[ "$#" -gt 0 ]]; do
+  opt="$1"
+  shift  # expose the next argument
+  case "$opt" in
+    --) break ;;
```
maxim-belkin, line 299: I think we don't need this `--` tbh because this script never passes control to other scripts or programs.

install.sh
XuehaiPan: I find a way to determine which shell we are using with `ps -p $$ -ucomm=''` (tested on macOS Catalina and Ubuntu 16.04/20.04). And I add an assertion for BASH here:
```bash
SHELL_NAME="$(ps -p $$ -oucomm='' | head -n 1 | tr -d ' ')"
if [[ -z "${BASH_VERSION:-}" || "${SHELL_NAME}" != "bash" ]]; then
  warn "BASH is required to interpret this script."; usage 1
fi
```
Without this assertion, invoke `install.sh` with `zsh` or `sh` do not get any error, but stuck at waiting for user input here:
> Press RETURN to continue or any other key to abort

(`sh` is symlinked to `bash` on macOS, but on Ubuntu they are different shell.)

XuehaiPan: I think `-n "${BASH_VERSION:-}"` is enough. We don't need the shell name at all.

.github/workflows/tests.yml
```diff
@@ -83,7 +83,18 @@ jobs:
 
       - run: /bin/bash uninstall.sh -f >/dev/null
 
-      - run: /bin/bash -c "$(cat install.sh)"
```
carlocab, line 86: I think we still need a test that successfully runs the install script without arguments.

XuehaiPan: There already is a test at line 78 for no argmuents. And the new test only replaces the protocol of git remotes.

install.sh
```diff
@@ -1,17 +1,22 @@
...
 set -u
 
+abort() {
+  printf "%s\n" "$@"
+  exit 1
+}
+
+# BASH is required
+if [ -z "${BASH_VERSION:-}" ]; then
+  abort "BASH is required to interpret this script."
```
MikeMcQuaid, line 11:
```suggestion
if [ -z "${BASH_VERSION:-}" ]; then
  abort "Bash is required to interpret this script."
```

install.sh
MikeMcQuaid: > I prefer to make everything automated once the sudo privilege is checked.

Given it's already not automated: we could make it a single prompt. I guess you'll need those variables set for Homebrew/brew anyway so as long as we notify and prompt it seems preferable to having to add e.g. argument parsing into this script.

XuehaiPan: I put the prompts for Git remotes before `wait_for_user` now. We can read the remote URLs from environment variables, but it needs three commands to install:
```bash
export HOMEBREW_BREW_GIT_REMOTE="..."
export HOMEBREW_CORE_GIT_REMOTE="..."
bash -c "$(cat install.sh)"
```
It only need one command to install if the URLs are parsed from arguments:
```bash
bash -c "$(cat install.sh)" -b "..." -c "..."
```
So I think _parse the remotes from arguments_ is better.

On my devices, I am using non-default remotes for brew's repos, but I do not set `HOMEBREW_{BREW,CORE}_GIT_REMOTE` in my shell environment. `brew update` set the remote of local repos if `HOMEBREW_{BREW,CORE}_GIT_REMOTE` is set to non-default value. However, `brew update` does not change the remote of local repos when those env-vars are not set, even if `git -C "$(brew --repo)" remote get-url origin` reports a non-default URL. In other words, users only need to set these env-vars at the begining of installation. Once the installation is complete, these variables can be unset. Users do not need them for daily use.

install.sh
```diff
@@ -262,6 +266,59 @@ outdated_glibc() {
...
+  -c, --core-git-remote=URL   Use this URL as the Homebrew/homebrew-core git remote.
+                              If no argument is given, the default remote is used.
+                              Default: ${tty_underline}${HOMEBREW_CORE_DEFAULT_GIT_REMOTE}${tty_reset}.
+  -h, --help                  Show this help message and exit.
+EOS
+  exit "${1:-0}"
+}
+
+# Parse remote URLs of Homebrew repositories from command line arguments if presented.
+unset HOMEBREW_{BREW,CORE}_GIT_REMOTE  # unset variables from environment at first
```
XuehaiPan, line 289: We can keep these variables from environments if remove the `unset` statement here. The priority order for `HOMEBREW_{BREW,CORE}_GIT_REMOTE` is:
1. from arguments if specified
2. from environment if set
3. the default value

But I think it could be difficult to write the usage clear enough for users.

install.sh
MikeMcQuaid: > It only need one command to install if the URLs are parsed from arguments:

You'll need to set those variables after installation, though, or `brew doctor` will complain, which isn't something we want to be possible with something from the official installation script.

XuehaiPan: Oh I see. I have never ran `brew doctor` before, so I didn't know that. So users are recommanded to set `HOMEBREW_{BREW,CORE}_GIT_REMOTE` if they are using non-default ones, is that correct? I revert to the original approach, _read Homebrew remote URLs from environment variables and prompt for non-default values_.

install.sh
```diff
@@ -500,6 +516,21 @@ if should_install_command_line_tools; then
...
 
+non_default_repos=""
+additional_shellenv_command=()
+if [[ "$HOMEBREW_BREW_DEFAULT_GIT_REMOTE" != "$HOMEBREW_BREW_GIT_REMOTE" ]]; then
+  ohai "HOMEBREW_BREW_GIT_REMOTE is set to a non-default URL:"
+  echo "${tty_underline}${HOMEBREW_BREW_GIT_REMOTE}${tty_reset} will be used for Homebrew/brew Git remote."
+  non_default_repos="Homebrew/brew"
+  additional_shellenv_command+=("export HOMEBREW_BREW_GIT_REMOTE=\"$HOMEBREW_BREW_GIT_REMOTE\"")
+fi
+if [[ "$HOMEBREW_CORE_DEFAULT_GIT_REMOTE" != "$HOMEBREW_CORE_GIT_REMOTE" ]]; then
```
MikeMcQuaid, line 527:
```suggestion
fi

if [[ "$HOMEBREW_CORE_DEFAULT_GIT_REMOTE" != "$HOMEBREW_CORE_GIT_REMOTE" ]]; then
```

install.sh
```diff
@@ -500,6 +516,21 @@ if should_install_command_line_tools; then
   ohai "The Xcode Command Line Tools will be installed."
 fi
 
+non_default_repos=""
+additional_shellenv_command=()
+if [[ "$HOMEBREW_BREW_DEFAULT_GIT_REMOTE" != "$HOMEBREW_BREW_GIT_REMOTE" ]]; then
+  ohai "HOMEBREW_BREW_GIT_REMOTE is set to a non-default URL:"
+  echo "${tty_underline}${HOMEBREW_BREW_GIT_REMOTE}${tty_reset} will be used for Homebrew/brew Git remote."
```
MikeMcQuaid, line 523: Feels like a shell bell might be nice here too: 

Could split this into a function and use it:
https://github.com/Homebrew/install/blob/e2448b098d5519e605b04d4c81622b075f33117b/install.sh#L656-L659

install.sh
```diff
@@ -500,6 +516,21 @@ if should_install_command_line_tools; then
...
+  non_default_repos="Homebrew/brew"
+  additional_shellenv_command+=("export HOMEBREW_BREW_GIT_REMOTE=\"$HOMEBREW_BREW_GIT_REMOTE\"")
+fi
+if [[ "$HOMEBREW_CORE_DEFAULT_GIT_REMOTE" != "$HOMEBREW_CORE_GIT_REMOTE" ]]; then
+  ohai "HOMEBREW_CORE_GIT_REMOTE is set to a non-default URL:"
+  echo "${tty_underline}${HOMEBREW_CORE_GIT_REMOTE}${tty_reset} will be used for Homebrew/core Git remote."
```
MikeMcQuaid, line 529: Same here with shell bell

install.sh
```diff
@@ -682,29 +713,34 @@ EOS
...
   cat <<EOS
 - Add Homebrew to your ${tty_bold}PATH${tty_reset} in ${tty_underline}${shell_profile}${tty_reset}:
-    echo 'eval \$(${HOMEBREW_PREFIX}/bin/brew shellenv)' >> ${shell_profile}
-    eval \$(${HOMEBREW_PREFIX}/bin/brew shellenv)
+    echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"' >> ${shell_profile}
+    eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
 EOS
 fi
+if [[ -n "${non_default_repos}" ]]; then
+  if [[ "${#additional_shellenv_command[@]}" -gt 1 ]]; then remotes="remotes"; else remotes="remote"; fi
```
MikeMcQuaid, line 739:
```suggestion
  if [[ "${#additional_shellenv_command[@]}" -gt 1 ]]; then
    remotes="remotes"
  else
    remotes="remote"
  fi
```

MikeMcQuaid (commented):
This is looking great. Final request would be to document these in the `README` for this repository and/or in https://github.com/Homebrew/brew/edit/master/docs/Installation.md.

README.md
```diff
@@ -10,6 +10,21 @@ More installation information and options: https://docs.brew.sh/Installation.
 
 If running Linux or WSL, [there are some pre-requisite packages to install](https://docs.brew.sh/Homebrew-on-Linux#requirements).
 
+You can set `HOMEBREW_BREW_GIT_REMOTE` and/or `HOMEBREW_CORE_GIT_REMOTE` in your shell environment to use custom Git mirrors to speed up brew update and brew tap.
```
MikeMcQuaid, line 13:
```suggestion
You can set `HOMEBREW_BREW_GIT_REMOTE` and/or `HOMEBREW_CORE_GIT_REMOTE` in your shell environment to use geolocalised Git mirrors to speed up Homebrew's installation with this script and, after installation, `brew update`.
```

README.md
```diff
@@ -10,6 +10,21 @@ More installation information and options: https://docs.brew.sh/Installation.
...
+
+```bash
+export HOMEBREW_BREW_GIT_REMOTE="..."  # put your mirror URL of Homebrew/brew Git remote here
+export HOMEBREW_CORE_GIT_REMOTE="..."  # put your mirror URL of Homebrew/core Git remote here
```
MikeMcQuaid, line 17:
```suggestion
export HOMEBREW_BREW_GIT_REMOTE="..."  # put your Git mirror of Homebrew/brew here
export HOMEBREW_CORE_GIT_REMOTE="..."  # put your Git mirror of Homebrew/homebrew-core here
```

README.md
```diff
@@ -10,6 +10,21 @@ More installation information and options: https://docs.brew.sh/Installation.
...
+/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install.sh)"
+```
+
+| Variable                 | Default                                                                                                |
+| ------------------------ | ------------------------------------------------------------------------------------------------------ |
+| HOMEBREW_BREW_GIT_REMOTE | macOS / Linux: https://github.com/Homebrew/brew                                                        |
+| HOMEBREW_CORE_GIT_REMOTE | macOS: https://github.com/Homebrew/homebrew-core<br/>Linux: https://github.com/Homebrew/linuxbrew-core |
```
MikeMcQuaid, line 24:
```suggestion
```

I think this bit can be skipped.

MikeMcQuaid (commented):
Great work! Final suggested changes and then I think this can 🚢

MikeMcQuaid (approved):

MikeMcQuaid:
Incredible work here @XuehaiPan, thanks so much!

PR #537, opened 2021-03-23
Replace references to master with HEAD

carlocab:
We use `HEAD` in the install instructions on brew.sh, so we may as well
do the same here.

There are still some references to `master` in `install.sh`, but I don't
think those can be changed.


Rylan12 (approved):
🎉

MikeMcQuaid:
Nice work @carlocab!

PR #542, opened 2021-05-10
install.sh: set sticky bit on some directories

Bo98:
Seems like a safer thing to do if anything shares prefix with us.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -576,6 +576,13 @@ if [[ "${#mkdirs[@]}" -gt 0 ]]; then
   execute_sudo "$CHGRP" "$GROUP" "${mkdirs[@]}"
 fi
 
+directories=(bin etc lib sbin)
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -576,6 +576,11 @@ if [[ "${#mkdirs[@]}" -gt 0 ]]; then
   execute_sudo "$CHGRP" "$GROUP" "${mkdirs[@]}"
 fi
 
+directories=(bin etc lib sbin)
+for dir in "${directories[@]}"; do
+  execute_sudo "/bin/chmod" "+t" "${HOMEBREW_PREFIX}/${dir}"
```
MikeMcQuaid, line 581: Can you add a wee comment explaining what the sticky bit does and why we're setting it?

MikeMcQuaid (dismissed):

PR #550, opened 2021-06-25
install: fix directory permissions (fixes #549)

### LINKED ISSUE #549: Wrong file permissions for zsh completion directory /usr/local/share/zsh/site-functions
The install script creates the `/usr/local/share/zsh` and `/usr/local/share/zsh/site-functions` directories with the wrong permission the first time it is executed. They are created as group writable, which prevents the completion scripts in those directories from being used.

Running the install script for a second time will fix the permissions.

I suspect that the command that creates the directories and the ones that ensure their correct permissions are not in the correct order.

The issue can be verified using compaudit on a fresh installation:
```
autoload -Uz compaudit
compaudit
```

XuehaiPan:
Several fixes:

1. On macOS, the octal format of `stat` for access rights is `%A`, but it is `%a` on Linux.
2. Do add rights for other users not in the file's groups (`o`). Command `chmod 755` will add read and execute permissions for other users not in the file's groups (`o`). For exmaple, if users set `umask 027` in their shell profile, command `mkdir newdir` will create a directory with mode 750 (rwxr-x---), but `chmod 755 newdir` will change it's mode to 755 (rwxr-xr-x).
3. Always remove write permission for groups (`g` and `o`) for zsh site-functions (fixes #549).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -43,6 +43,7 @@ if [[ -z "${HOMEBREW_ON_LINUX-}" ]]; then
   HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/homebrew-core"
 
   STAT="stat -f"
+  PERMISSION_FORMAT="%A"
   CHOWN="/usr/sbin/chown"
   CHGRP="/usr/bin/chgrp"
   GROUP="admin"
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Makes sense, thanks @XuehaiPan!

PR #567, opened 2021-08-09
install.sh: add conditional post-install message on Linux

maxim-belkin:
On Linux systems where cURL/Git (that satisfy Homebrew requirements) are not in /usr/bin, instruct users which HOMEBREW_* variables they should set to enable Homebrew to run.
Example of proposed output:
```
- Attention: Git/cURL that satisfy Homebrew requirements are not in /usr/bin:
    cURL: /sw/EasyBuild/software/cURL/7.59.0/bin/curl
    Git: /sw/EasyBuild/software/git/2.17.0/bin/git
  To instruct Homebrew to use them, set the following environment variables in your current shell
  and in /u/staff/mbelkin/.bash_profile:
    export HOMEBREW_DEVELOPER=1
    export HOMEBREW_CURL_PATH="/sw/EasyBuild/software/cURL/7.59.0/bin/curl"
    export HOMEBREW_GIT_PATH="/sw/EasyBuild/software/git/2.17.0/bin/git"
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -819,6 +819,21 @@ if [[ "$UNAME_MACHINE" == "arm64" ]] || [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
     eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
 EOS
 fi
+if [[ "${HOMEBREW_ON_LINUX-}" && ( -n "${HOMEBREW_CURL_PATH-}" || -n "${HOMEBREW_GIT_PATH-}" ) ]]
+then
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
If we used these automatically in the installer I think we should probably just use them automatically in Homebrew/brew now too (like we do with Ruby). Thoughts?

maxim-belkin:
> If we used these automatically in the installer I think we should probably just use them automatically in Homebrew/brew now too (like we do with Ruby). Thoughts?

Agreed. I'll work on this in homebrew/brew.

PR #572, opened 2021-09-01
install.sh: show prompt messages before asking password

XuehaiPan:
Function `have_sudo_access` executes before any prompt message showing up.

Screenshot


MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @XuehaiPan!

PR #573, opened 2021-09-01
install.sh: use relative path for symlinks

XuehaiPan:
Change absolute path to relative path for `brew` executable on Linux and ARM64 macOS. This modification allows user to move their brew installation elsewhere after then.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -742,7 +742,11 @@ ohai "Downloading and installing Homebrew..."
   execute "git" "reset" "--hard" "origin/master"
 
   if [[ "${HOMEBREW_REPOSITORY}" != "${HOMEBREW_PREFIX}" ]]; then
-    execute "ln" "-sf" "${HOMEBREW_REPOSITORY}/bin/brew" "${HOMEBREW_PREFIX}/bin/brew"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -742,7 +742,11 @@ ohai "Downloading and installing Homebrew..."
   execute "git" "reset" "--hard" "origin/master"
 
   if [[ "${HOMEBREW_REPOSITORY}" != "${HOMEBREW_PREFIX}" ]]; then
-    execute "ln" "-sf" "${HOMEBREW_REPOSITORY}/bin/brew" "${HOMEBREW_PREFIX}/bin/brew"
+    if [[ "${HOMEBREW_REPOSITORY}" == "${HOMEBREW_PREFIX}/Homebrew" ]]; then
+      execute "ln" "-sf" "../Homebrew/bin/brew" "${HOMEBREW_PREFIX}/bin/brew"
+    else  # absolute path as fallback option (we probably never reach here)
```
MikeMcQuaid, line 747: Let's error out on this case if we should never reach here.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @XuehaiPan!

PR #577, opened 2021-09-15
style: fix inconsistent code style for shell scripts

XuehaiPan:
These changes are fixed by `brew style --fix install.sh`. The new shell formater is introduced in Homebrew/brew#12044.

See Homebrew/brew#12044 for more details.

`brew style --fix install.sh` before all commits (d0833a151922ae830435cc682cc999bb1626b427@master):

<details>
<summary>Results...</summary>
<p>

```

In install.sh line 21:
if [[ "$OS" == "Linux" ]]; then
       ^-^ SC2250: Prefer putting braces around variable references even when not strictly required.

Did you mean: 
if [[ "${OS}" == "Linux" ]]; then


In install.sh line 23:
elif [[ "$OS" != "Darwin" ]]; then
         ^-^ SC2250: Prefer putting braces around variable references even when not strictly required.

Did you mean: 
elif [[ "${OS}" != "Darwin" ]]; then


In install.sh line 33:
  if [[ "$UNAME_MACHINE" == "arm64" ]]; then
         ^------------^ SC2250: Prefer putting braces around variable references even when not strictly required.

Did you mean: 
  if [[ "${UNAME_MACHINE}" == "arm64" ]]; then


In install.sh line 73:
if [[ "$HOMEBREW_BREW_GIT_REMOTE" == "${HOMEBREW_BREW_DEFAULT_GIT_REMOTE}.git" ]]; then
       ^-----------------------^ SC2250: Prefer putting braces around variable references even when not strictly required.

Did you mean: 
if [[ "${HOMEBREW_BREW_GIT_REMOTE}" == "${HOMEBREW_BREW_DEFAULT_GIT_REMOTE}.git" ]]; then


In install.sh line 76:
if [[ "$HOMEBREW_CORE_GIT_REMOTE" == "${HOMEBREW_CORE_DEFAULT_GIT_REMOTE}.git" ]]; then
       ^-----------------------^ SC2250: Prefer putting braces around variable references even when not strictly required.

Did you mean: 
if [[ "${HOMEBREW_CORE_GIT_REMOTE}" == "${HOMEBREW_CORE_DEFAULT_GIT_REMOTE}.git" ]]; then


In install.sh line 135:
  if [[ -z "${HOMEBREW_ON_LINUX-}" ]] && [[ "$HAVE_SUDO_ACCESS" -ne 0 ]]; then
                                             ^---------------^ SC2250: Prefer putting braces around variable references even when not strictly required.

Did you mean: 
[...]

PR #581, opened 2021-09-17
fix: correct overzealous quoting in install.sh

BMDan:
The `$STAT` variable was quoted in #577 , causing errors like:

```
/bin/bash: line 265: stat -f: command not found
/bin/bash: line 265: stat -f: command not found
/bin/bash: line 277: stat -f: command not found
/bin/bash: line 285: stat -f: command not found
/bin/bash: line 277: stat -f: command not found
/bin/bash: line 285: stat -f: command not found
```

Here, we refactor it into `$STAT_FLAG`, so that we can keep our quotes nice and safe.  Note also that it's `$STAT_FLAG` and not `$STAT_FLAGS` as a subtle reminder to only supply a single flag.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -48,7 +48,7 @@ then
   HOMEBREW_CACHE="${HOME}/Library/Caches/Homebrew"
   HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/homebrew-core"
 
-  STAT="stat -f"
+  STAT_FLAG="-f"
   PERMISSION_FORMAT="%A"
   CHOWN="/usr/sbin/chown"
   CHGRP="/usr/bin/chgrp"
@@ -63,7 +63,7 @@ else
   HOMEBREW_CACHE="${HOME}/.cache/Homebrew"
   HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/linuxbrew-core"
 
-  STAT="stat --printf"
+  STAT_FLAG="--printf"
   PERMISSION_FORMAT="%a"
   CHOWN="/bin/chown"
   CHGRP="/bin/chgrp"
@@ -263,7 +263,7 @@ should_install_command_line_tools() {
 }
 
 get_permission() {
-  "${STAT}" "${PERMISSION_FORMAT}" "$1"
+  stat "${STAT_FLAG}" "${PERMISSION_FORMAT}" "$1"
 }
 
 user_only_chmod() {
@@ -275,7 +275,7 @@ exists_but_not_writable() {
 }
 
 get_owner() {
-  "${STAT}" "%u" "$1"
+  stat "${STAT_FLAG}" "%u" "$1"
 }
 
 file_not_owned() {
@@ -283,7 +283,7 @@ file_not_owned() {
 }
 
 get_group() {
-  "${STAT}" "%g" "$1"
+  stat "${STAT_FLAG}" "%g" "$1"
 }
 
 file_not_grpowned() {


BMDan:
👀 @MikeMcQuaid @XuehaiPan

XuehaiPan:
Thanks @BMDan! I think its OK to just remove the quotes and add a comment above. Or refactor all these commands to arrays like `“CMD[@]”`

https://github.com/Homebrew/install/blob/9eca6c27ff87a93c3d9dd61c32f5102f7e754c4f/install.sh#L141-L147

install.sh
```diff
@@ -48,7 +48,7 @@ then
   HOMEBREW_CACHE="${HOME}/Library/Caches/Homebrew"
   HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/homebrew-core"
 
-  STAT="stat -f"
+  STAT_FLAG="-f"
   PERMISSION_FORMAT="%A"
   CHOWN="/usr/sbin/chown"
   CHGRP="/usr/bin/chgrp"
```
XuehaiPan, line 54:
```suggestion
  STAT=("stat" "-f")
  PERMISSION_FORMAT="%A"
  CHOWN=("/usr/sbin/chown")
  CHGRP=("/usr/bin/chgrp")
```

Similarly, change `TOUCH` to an array below.

Then replace all `${STAT}` to `${STAT[@]}` in this file. Similar changes for `CHOWN`, `CHGRP`, and `TOUCH`.

install.sh
```diff
@@ -63,7 +63,7 @@ else
   HOMEBREW_CACHE="${HOME}/.cache/Homebrew"
   HOMEBREW_CORE_DEFAULT_GIT_REMOTE="https://github.com/Homebrew/linuxbrew-core"
 
-  STAT="stat --printf"
+  STAT_FLAG="--printf"
   PERMISSION_FORMAT="%a"
   CHOWN="/bin/chown"
```
XuehaiPan, line 68: Change to arrays mentioned above.

install.sh
```diff
@@ -263,7 +263,7 @@ should_install_command_line_tools() {
 }
 
 get_permission() {
-  "${STAT}" "${PERMISSION_FORMAT}" "$1"
+  stat "${STAT_FLAG}" "${PERMISSION_FORMAT}" "$1"
```
XuehaiPan, line 266:
```suggestion
  "${STAT[@]}" "${PERMISSION_FORMAT}" "$1"
```

XuehaiPan:
Refactor the followings:

https://github.com/Homebrew/install/blob/9eca6c27ff87a93c3d9dd61c32f5102f7e754c4f/install.sh#L133-L149

to:

```bash
  if [[ -z "${HAVE_SUDO_ACCESS-}" ]]
  then
    local -a SUDO=("/usr/bin/sudo")
    if [[ -n "${args[*]-}" ]]
    then
      SUDO+=("${args[@]}")
    fi
    if [[ -n "${NONINTERACTIVE-}" ]]
    then
      "${SUDO[@]}" -l mkdir &>/dev/null
    else
      "${SUDO[@]}" -v && "${SUDO[@]}" -l mkdir &>/dev/null
    fi
    HAVE_SUDO_ACCESS="$?"
  fi
```

XuehaiPan:
We could merge this PR to make the installer work first, and then open a new PR to refactor the above changes (command strings as arrays).

cc @MikeMcQuaid @BMDan.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks for the quick fix @BMDan!

@XuehaiPan Yes, good idea on merging and then refactoring 👍🏻

PR #592, opened 2021-11-07
install.sh: wording maintenance

EricFromCanada:
- drop reference to IRC and match wording from diagnostic.rb
- various grammar fixes
- some formatting fixes for consistency
- ensure documentation link appears last with a blank line following

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -11,8 +11,8 @@ abort() {
   exit 1
 }
 
-# Fail fast with concise message when not using bash
-# Single brackets is needed here for POSIX compatibility
+# Fail fast with a concise message when not using bash
+# Single brackets are needed here for POSIX compatibility
 # shellcheck disable=SC2292
 if [ -z "${BASH_VERSION:-}" ]
 then
@@ -299,7 +299,7 @@ file_not_grpowned() {
   [[ " $(id -G "${USER}") " != *" $(get_group "$1") "* ]]
 }
 
-# Please sync with 'test_ruby()' in 'Library/Homebrew/utils/ruby.sh' from Homebrew/brew repository.
+# Please sync with 'test_ruby()' in 'Library/Homebrew/utils/ruby.sh' from the Homebrew/brew repository.
 test_ruby() {
   if [[ ! -x "$1" ]]
   then
@@ -334,7 +334,7 @@ test_git() {
   version_ge "$(major_minor "${git_version_output##* }")" "$(major_minor "${REQUIRED_GIT_VERSION}")"
 }
 
-# Search given executable in PATH (remove dependency for `which` command)
+# Search for the given executable in PATH (avoids a dependency on the `which` command)
 which() {
   # Alias to Bash built-in command `type -P`
   type -P "$@"
@@ -378,7 +378,7 @@ Homebrew requires Ruby ${REQUIRED_RUBY_VERSION} which was not found on your syst
 Homebrew portable Ruby requires Glibc version ${REQUIRED_GLIBC_VERSION} or newer,
 and your Glibc version is too old.
 See ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#requirements${tty_reset}
-Install Ruby ${REQUIRED_RUBY_VERSION} and add its location to your PATH.
+Please install Ruby ${REQUIRED_RUBY_VERSION} and add its location to your PATH.
 EOFABORT
   )"
 fi
@@ -416,7 +416,7 @@ then
   then
     abort "$(
       cat <<-EOABORT
-Git that is available on your system does not satisfy Homebrew requirements.
+The version of Git that was found does not satisfy requirements for Homebrew.
 Please install Git ${REQUIRED_GIT_VERSION} or newer and add it to your PATH.
 EOABORT
     )"
@@ -442,7 +442,7 @@ then
   then
     abort "$(
       cat <<-EOABORT
-cURL that is available on your system does not satisfy Homebrew requirements.
+The version of cURL that was found does not satisfy requirements for Homebrew.
 Please install cURL ${REQUIRED_CURL_VERSION} or newer and add it to your PATH.
 EOABORT
     )"
@@ -484,7 +484,7 @@ else
     trap exit SIGINT
     if ! /usr/bin/sudo -n -v &>/dev/null
     then
-      ohai "Select the Homebrew installation directory"
+      ohai "Select a Homebrew installation directory:"
       echo "- ${tty_bold}Enter your password${tty_reset} to install to ${tty_underline}${HOMEBREW_PREFIX_DEFAULT}${tty_reset} (${tty_bold}recommended${tty_reset})"
       echo "- ${tty_bold}Press Control-D${tty_reset} to install to ${tty_underline}${HOME}/.linuxbrew${tty_reset}"
       echo "- ${tty_bold}Press Control-C${tty_reset} to cancel installation"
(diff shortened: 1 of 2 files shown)


dawidd6 (approved):

RandomDSdevel:
> - various grammar fixes

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;I was actually thinking about making some changes similar, but not entirely identical, to the yours here at some point maybe–soon-ish.  Doing that crossed my mind while I was poking around in the same areas of the code your commits touch as part of a minor attempt to stave a tiny amount of the development purgatory which [Portage-Brew](https://github.com/portage-brew/brew)'s been stuck in for ages off.  I wasn't sure how many of those changes I'd end up sending upstream, however; though I've yet to formally document it, I heavily suspect our prose style's going to be a at least a bit different.  I don't think that would conflict too much here, though.  Anyway, would you mind if I gave a few suggestions regarding your changes?

install.sh
```diff
@@ -712,16 +712,16 @@ additional_shellenv_commands=()
 if [[ "${HOMEBREW_BREW_DEFAULT_GIT_REMOTE}" != "${HOMEBREW_BREW_GIT_REMOTE}" ]]
 then
   ohai "HOMEBREW_BREW_GIT_REMOTE is set to a non-default URL:"
-  echo "${tty_underline}${HOMEBREW_BREW_GIT_REMOTE}${tty_reset} will be used for Homebrew/brew Git remote."
```
XuehaiPan, line 715: The phrase was taken from:

https://github.com/Homebrew/brew/blob/aaea979a428a004a9a681e92deecc122320c5a48/Library/Homebrew/cmd/update.sh#L495-L497

install.sh
```diff
@@ -712,16 +712,16 @@ additional_shellenv_commands=()
...
+  echo "${tty_underline}${HOMEBREW_BREW_GIT_REMOTE}${tty_reset} will be used as the Homebrew/brew Git remote."
   non_default_repos="Homebrew/brew"
   additional_shellenv_commands+=("export HOMEBREW_BREW_GIT_REMOTE=\"${HOMEBREW_BREW_GIT_REMOTE}\"")
 fi
 
 if [[ "${HOMEBREW_CORE_DEFAULT_GIT_REMOTE}" != "${HOMEBREW_CORE_GIT_REMOTE}" ]]
 then
   ohai "HOMEBREW_CORE_GIT_REMOTE is set to a non-default URL:"
-  echo "${tty_underline}${HOMEBREW_CORE_GIT_REMOTE}${tty_reset} will be used for Homebrew/core Git remote."
-  non_default_repos="${non_default_repos:-}${non_default_repos:+ and }Homebrew/core"
```
XuehaiPan, line 724: Similiar for Homebrew/core:

https://github.com/Homebrew/brew/blob/aaea979a428a004a9a681e92deecc122320c5a48/Library/Homebrew/cmd/update.sh#L514-L516

https://github.com/Homebrew/brew/blob/5c44aca2e197fcaf0d1377257686e3eea7710405/Library/Homebrew/tap.rb#L788-L790

install.sh
```diff
@@ -573,11 +573,11 @@ EOABORT
...
 After installation, you will encounter build failures with some formulae.
 Please create pull requests instead of asking for help on Homebrew\'s GitHub,
-Twitter or IRC. You are responsible for resolving any issues you experience
-while you are running this ${what}.
+Twitter or any other official channels. You are responsible for resolving any
+issues you experience while you are running this ${what}.
 EOS
     )
-"
+" | tr -d "\\"
```
XuehaiPan, line 580: Just remove the backslash above.

```
Homebrew\'s GitHub --> Homebrew's GitHub
```

MikeMcQuaid (approved):
Thanks for this @EricFromCanada!

install.sh
EricFromCanada: I did that originally, but it turns out a bare single quote in that context triggers a strange bug further down the file.
```
./install.sh: line 965: syntax error near unexpected token `('
./install.sh: line 965: `    echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"' >> ${shell_profile}'
```

XuehaiPan: We can use `cat <<"EOS"` to avoid substitution for escape characters.

XuehaiPan: > We can use `cat <<"EOS"` to avoid substitution for escape characters.

Oh, I'm wrong. We need to expand ${what} in the message.

> but it turns out a bare single quote in that context triggers a strange bug further down the file.
>
> ```
> ./install.sh: line 965: syntax error near unexpected token `('
> ./install.sh: line 965: `    echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"' >> ${shell_profile}'
> ```

I cannot reproduce this with Bash 5 and the error message looks doesn't make sense. If the issue still arises (probably for other versions of Bash?), it is fine for me to use `tr -d`.

EricFromCanada: It affects bash 3.2, which is what macOS still ships with.

install.sh
EricFromCanada: Thanks, I've made a note to update those as well.

install.sh
EricFromCanada: These too.

install.sh
```diff
@@ -922,7 +925,7 @@ echo "$(
   cat <<EOS
 ${tty_bold}Read the analytics documentation (and how to opt-out) here:
   ${tty_underline}https://docs.brew.sh/Analytics${tty_reset}
-No analytics data has been sent yet (or will be during this \`install\` run).
+No analytics data has been sent yet (nor will be during this ${tty_bold}install${tty_reset} run).
```
RandomDSdevel, line 928: &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;'Nor' goes with a preceding 'neither,' as in 'neither …, nor ….'  Alternatively, give the dependent clause its own internal subject to avoid having to use that kind of construction:  
```suggestion
No analytics data has been sent yet (nor will it be during this ${tty_bold}install${tty_reset} run).
```

or

```suggestion
No analytics data has been sent yet (nor will any be during this ${tty_bold}install${tty_reset} run).
```

install.sh
```diff
@@ -978,10 +981,6 @@ then
   printf "    %s\n" "${additional_shellenv_commands[@]}"
 fi
 
-echo "- Run \`brew help\` to get started"
-echo "- Further documentation: "
-echo "    ${tty_underline}https://docs.brew.sh${tty_reset}"
-
 if [[ -n "${HOMEBREW_ON_LINUX-}" ]]
 then
   echo "- Install the Homebrew dependencies if you have sudo access:"
```
RandomDSdevel, line 986:
```suggestion
  echo "- Install Homebrew's dependencies if you have sudo access:"
```

RandomDSdevel (commented):
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Regarding leaving suggestions like I mentioned above in https://github.com/Homebrew/install/pull/592#issuecomment-962673718, I went ahead and left mine.  Take which of them might appear suitable for what you will.

MikeMcQuaid (dismissed):

Rylan12 (commented):
It would be great to set up a standard set of formatting rules for things like this (e.g. underline paths and urls, bold commands, etc.) that could apply across the organization. It's probably not automatically enforceable, but at least would serve as a guideline to reference.

The [Prose Style Guidelines](https://docs.brew.sh/Prose-Style-Guidelines) are probably a good place for this

MikeMcQuaid:
Thanks again @EricFromCanada!

PR #608, opened 2022-01-03
install.sh: correct detection of arm64 on Linux

andreineculau:
'uname -m' returns aarch64, not arm64

ref: https://github.com/Homebrew/install/pull/518
ref: https://github.com/Homebrew/install/commit/0d46eb17cff2649ac15a0b3296c782d75218863f

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -527,7 +527,7 @@ then
   fi
 else
   # On Linux, support only 64-bit Intel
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):

MikeMcQuaid:
Good catch, thanks @andreineculau!

PR #615, opened 2022-01-29
always remove command line tools placeholder

andreineculau:
I was looking for inspiration on how homebrew does a silent install of Command Line Tools,
and noticed that removing the placeholder is wrapped in the conditional.
In practice, I don't see how command line tools are not installed, and softwareupdate can come up empty handed, but who knows. I honestly don't know if/what consequences there are if the placeholder is left behind.


MikeMcQuaid (approved):
Makes sense, thanks!

PR #628, opened 2022-03-24
Respect `$HOMEBREW_INSTALL_FROM_API` in install script

kidonng:
The `$HOMEBREW_INSTALL_FROM_API` environment, which lets Homebrew use the homebrew/core and homebrew/cask without local git checkout, is [official supported since 3.3.0](https://brew.sh/2021/10/25/homebrew-3.3.0/).

This pull request makes the install script respect this environment variable, by skip cloning homebrew/core entirely during installation if `$HOMEBREW_INSTALL_FROM_API` is set.

This also addresses #523 with a different approach, by allowing a minimal installation possible with the official installation script instead of a hand-rolled one.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -759,6 +759,13 @@ then
   additional_shellenv_commands+=("export HOMEBREW_CORE_GIT_REMOTE=\"${HOMEBREW_CORE_GIT_REMOTE}\"")
 fi
 
+if [[ -n "${HOMEBREW_INSTALL_FROM_API-}" ]]
+then
+  ohai "HOMEBREW_INSTALL_FROM_API is set."
+  echo "Homebrew/homebrew-core will not be tapped during this ${tty_bold}install${tty_reset} run."
+  additional_shellenv_commands+=("export HOMEBREW_INSTALL_FROM_API=\"${HOMEBREW_INSTALL_FROM_API}\"")
+fi
+
 if [[ -z "${NONINTERACTIVE-}" ]]
 then
   ring_bell
@@ -916,7 +923,15 @@ ohai "Downloading and installing Homebrew..."
     fi
   fi
 
-  if [[ ! -d "${HOMEBREW_CORE}" ]]
+  if [[ -n "${HOMEBREW_INSTALL_FROM_API-}" ]]
+  then
+    ohai 'Skip tapping homebrew/core because `$HOMEBREW_INSTALL_FROM_API` is set.'
+    # Unset HOMEBREW_DEVELOPER since it is no longer needed and causes warnings during brew update below
+    if [[ -n "${HOMEBREW_ON_LINUX-}" && (-n "${HOMEBREW_CURL_PATH-}" || -n "${HOMEBREW_GIT_PATH-}") ]]
+    then
+      export -n HOMEBREW_DEVELOPER
+    fi
+  elif [[ ! -d "${HOMEBREW_CORE}" ]]
   then
     ohai "Tapping homebrew/core"
     (


install.sh
```diff
@@ -759,6 +759,13 @@ then
   additional_shellenv_commands+=("export HOMEBREW_CORE_GIT_REMOTE=\"${HOMEBREW_CORE_GIT_REMOTE}\"")
 fi
 
+if [[ -n "${HOMEBREW_INSTALL_FROM_API-}" ]]
+then
+  ohai "HOMEBREW_INSTALL_FROM_API is set."
+  echo "Homebrew/homebrew-core will not be tapped during this ${tty_bold}install${tty_reset} run."
+  additional_shellenv_commands+=("export HOMEBREW_INSTALL_FROM_API=\"${HOMEBREW_INSTALL_FROM_API}\"")
```
MikeMcQuaid, line 766: If it's already set: why does it need exported?

install.sh
```diff
@@ -916,7 +923,15 @@ ohai "Downloading and installing Homebrew..."
     fi
   fi
 
-  if [[ ! -d "${HOMEBREW_CORE}" ]]
+  if [[ -n "${HOMEBREW_INSTALL_FROM_API-}" ]]
+  then
+    ohai 'Skip tapping homebrew/core because `$HOMEBREW_INSTALL_FROM_API` is set.'
+    # Unset HOMEBREW_DEVELOPER since it is no longer needed and causes warnings during brew update below
```
MikeMcQuaid, line 929: What warnings? Was it set by this script or elsewhere?

MikeMcQuaid (commented):
Thanks for the PR, great work so far. Would be good to get this working together nicely.

install.sh
kidonng: Thanks for the reminder, this is copied from the code block just above and indeed not needed (it was for non default remotes).

install.sh
kidonng: `$HOMEBREW_DEVELOPER` may be set here:

https://github.com/Homebrew/install/blob/a64608bec0e11adfae1cea3b6b57fd4dcd3b1b69/install.sh#L482-L487

And the installation script would proceed to do a `brew update`, which will warn if both `$HOMEBREW_DEVELOPER` and `$HOMEBREW_INSTALL_FROM_API` are set:

https://github.com/Homebrew/install/blob/a64608bec0e11adfae1cea3b6b57fd4dcd3b1b69/install.sh#L938

```
...snip...
HEAD is now at b1577d347 Merge pull request #13038 from gromgit/github_api/fix_url
==> Skip tapping homebrew/core because `$HOMEBREW_INSTALL_FROM_API` is set.
Warning: Developers should not have HOMEBREW_INSTALL_FROM_API set!
Please unset HOMEBREW_INSTALL_FROM_API or turn developer mode off by running:
  brew developer off

Warning: Developers should not have HOMEBREW_INSTALL_FROM_API set!
Please unset HOMEBREW_INSTALL_FROM_API or turn developer mode off by running:
  brew developer off

Warning: Developers should not have HOMEBREW_INSTALL_FROM_API set!
Please unset HOMEBREW_INSTALL_FROM_API or turn developer mode off by running:
  brew developer off

==> Downloading https://ghcr.io/v2/homebrew/portable-ruby/portable-ruby/blobs/sha256:cd7fffb18ef9338baa670fc5e8fce99b0e3cc0f0fd7627bcbb56f3c8d54161d4
...snip...
```

MikeMcQuaid (dismissed):

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @kidonng!

Issue #634, opened 2022-04-12
Investigate .pkg and sharball installers


MikeMcQuaid:
These could be [generated using CPack](https://cmake.org/cmake/help/latest/manual/cpack-generators.7.html) and then generated on Homebrew/brew releases.

XuehaiPan:
How about `.sh` installer with embedded binaries. Since `brew` is a command-line tool and doesn't have a  GUI interface, it's natural for users to run the installer in Terminal rather than a GUI `.pkg` installer.

`conda` provides offline self-contained `.sh` installer. In the installer, it extract embedded binaries with command `dd`.

```bash
CONDA_EXEC="$PREFIX/conda.exe"
# 3-part dd from https://unix.stackexchange.com/a/121798/34459
# this is similar below with the tarball payload - see shar.py in constructor to see how
#    these values are computed.
{
    dd if="$THIS_PATH" bs=1 skip=28139                  count=4629                      2>/dev/null
    dd if="$THIS_PATH" bs=16384        skip=2                      count=978                   2>/dev/null
    dd if="$THIS_PATH" bs=1 skip=16056320                   count=307                     2>/dev/null
} > "$CONDA_EXEC"

chmod +x "$CONDA_EXEC"
```

See also [conda/constructor](https://github.com/conda/constructor).

MikeMcQuaid:
@XuehaiPan Would rather avoid including binaries here and instead let Homebrew bootstrap those itself. Would want to include essentially a tarball of just Homebrew/brew. homebrew-core isn't work handling as we're planning on using `HOMEBREW_INSTALL_FROM_API` to handle that case.

MikeMcQuaid:
Passing on this, sorry for the wasted work @XuehaiPan 😭

PR #636, opened 2022-04-12
install: sharball release support

XuehaiPan:
Add a new script to create a standalone sharball installer.

In the installer creator:

1. get the latest tag of Homebrew/brew from GItHub API
2. download tarball of Homebrew/brew
3. embed the tarball into `install.sh`

See also #634.

### DIFF
--- a/sharball-release.sh
+++ b/sharball-release.sh
@@ -1,0 +1,192 @@
+#!/bin/bash
+
+# We don't need return codes for "$(command)", only stdout is needed.
+# Allow `[[ -n "$(command)" ]]`, `func "$(command)"`, pipes, etc.
+# shellcheck disable=SC2312
+
+set -u
+
+abort() {
+  printf "%s\n" "$@"
+  exit 1
+}
+
+# Fail fast with a concise message when not using bash
+# Single brackets are needed here for POSIX compatibility
+# shellcheck disable=SC2292
+if [ -z "${BASH_VERSION:-}" ]
+then
+  abort "Bash is required to interpret this script."
+fi
+
+# string formatters
+if [[ -t 1 ]]
+then
+  tty_escape() { printf "\033[%sm" "$1"; }
+else
+  tty_escape() { :; }
+fi
+tty_mkbold() { tty_escape "1;$1"; }
+tty_blue="$(tty_mkbold 34)"
+tty_bold="$(tty_mkbold 39)"
+tty_reset="$(tty_escape 0)"
+
+shell_join() {
+  local arg
+  printf "%s" "$1"
+  shift
+  for arg in "$@"
+  do
+    printf " "
+    printf "%s" "${arg// /\ }"
+  done
+}
+
+ohai() {
+  printf "${tty_blue}==>${tty_bold} %s${tty_reset}\n" "$(shell_join "$@")"
+}
+
+line_range_of() {
+  # Usage: line_range_of file start end
+  head -n +"$3" "$1" | tail -n +"$2"
+}
+
+curl() {
+  command curl -fSL --retry "${HOMEBREW_CURL_RETRIES:-3}" "$@"
+}
+
+wait_for() {
+  local MESSAGE="$2" TIME="$1" t s="s"
+  echo "${MESSAGE}" >&2
+  echo -ne "\033[?25l" >&2
+  for ((t = TIME; t > 0; --t))
+  do
+    ((t > 1)) || s=""
+    echo -ne "Wait for ${t} second${s}.\033[K\r" >&2
+    sleep 1
+  done
+  echo -ne "\033[K\033[?25h" >&2
+}
+
+get_latest_version() {
+  # Usage: get_latest_version repo timeout
+  local REPO="$1" VERSION=""
+  local TIMEOUT="${2:-300}"
+  local TIME=0 INTERVAL=1 t
+
+  while true
+  do
+    ((TIME > 0)) && echo "Retrying..." >&2
+
+    VERSION="$(
+      curl --silent --connect-timeout 10 "https://api.github.com/repos/${REPO}/releases/latest" |
+        grep '"tag_name":' |
+        sed -E 's/^.*: *"([^"]+)",?$/\1/'
+    )"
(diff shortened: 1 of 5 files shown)


sharball-release.sh
```diff
@@ -0,0 +1,139 @@
...
+  head -n +"$3" "$1" | tail -n +"$2"
+}
+
+TMP_DIR="$(mktemp -d -t brew-install.XXXXXX)"
+BREW_DIR="${TMP_DIR}/brew"
+trap 'rm -rf "${TMP_DIR}"' EXIT
+
+HOMEBREW_BREW_GIT_REMOTE="https://github.com/Homebrew/brew.git"
+ohai "Creating fresh clone of Homebrew/brew from ${HOMEBREW_BREW_GIT_REMOTE}..."
+git clone https://github.com/Homebrew/brew.git "${BREW_DIR}"
```
MikeMcQuaid, line 59: What about downloading the release tarball instead of cloning?

MikeMcQuaid (commented):
Would probably rather this defers to another tool e.g. CPack?

sharball-release.sh
XuehaiPan: I'm considering including the `.git` directory, which the release tarball does not have. Besides, it's easier to get the latest tag with a complete clone.

MikeMcQuaid: I don't think we should be including a Git repository here.

XuehaiPan: In the current installer `install.sh`, it always requires a Internet connection for `git fetch`. I think the sharball would be a offine installer. If we don't include a Git repository, what's the difference to use the online installer `install.sh`.

MikeMcQuaid: Homebrew is usable without a Git repository. It's pretty atypical for software to bundle Git repositories. This would still allow you to download and install offline without including a Git repository.

MikeMcQuaid: Note the primary goal is the DMG and secondary is the sharball which is another reason for using something like CPack that can generate both.

sharball-release.sh
```diff
@@ -0,0 +1,165 @@
+#!/bin/bash
```
MikeMcQuaid, line 1: @XuehaiPan appreciate the work here but to save you effort: I'd really like to see how far we can go with an existing tool e.g. CPack or other rather than creating/maintaining a script like this by hand.

MikeMcQuaid (changes requested):

sharball-release.sh
XuehaiPan: @MikeMcQuaid I investigated CPack packaging tool last weekend. But I don't find a way to generate sharball packages using CPack.

[Cpack Generators](https://cmake.org/cmake/help/latest/manual/cpack-generators.7.html#id1):

- [CPack Archive Generator](https://cmake.org/cmake/help/latest/cpack_gen/archive.html)
- [CPack Bundle Generator](https://cmake.org/cmake/help/latest/cpack_gen/bundle.html)
- [CPack Cygwin Generator](https://cmake.org/cmake/help/latest/cpack_gen/cygwin.html)
- [CPack DEB Generator](https://cmake.org/cmake/help/latest/cpack_gen/deb.html)
- [CPack DragNDrop Generator](https://cmake.org/cmake/help/latest/cpack_gen/dmg.html)
- [CPack External Generator](https://cmake.org/cmake/help/latest/cpack_gen/external.html)
- [CPack FreeBSD Generator](https://cmake.org/cmake/help/latest/cpack_gen/freebsd.html)
- [CPack IFW Generator](https://cmake.org/cmake/help/latest/cpack_gen/ifw.html)
- [CPack NSIS Generator](https://cmake.org/cmake/help/latest/cpack_gen/nsis.html)
- [CPack NuGet Generator](https://cmake.org/cmake/help/latest/cpack_gen/nuget.html)
- [CPack PackageMaker Generator](https://cmake.org/cmake/help/latest/cpack_gen/packagemaker.html)
- [CPack productbuild Generator](https://cmake.org/cmake/help/latest/cpack_gen/productbuild.html)
- [CPack RPM Generator](https://cmake.org/cmake/help/latest/cpack_gen/rpm.html)
- [CPack WIX Generator](https://cmake.org/cmake/help/latest/cpack_gen/wix.html)

The most useful generators we are interested in are:

- [CPack Archive Generator](https://cmake.org/cmake/help/latest/cpack_gen/archive.html)
  - 7Z - 7zip - (.7z)
  - TBZ2 (.tar.bz2)
  - TGZ (.tar.gz)
  - TXZ (.tar.xz)
  - TZ (.tar.Z)
  - TZST (.tar.zst)
  - ZIP (.zip)

- [CPack DEB Generator](https://cmake.org/cmake/help/latest/cpack_gen/deb.html)
  - DEB (.deb) for Debian / Ubuntu

- [CPack DragNDrop Generator](https://cmake.org/cmake/help/latest/cpack_gen/dmg.html)
  - DragNDrop (.dmg) for macOS

- [CPack FreeBSD Generator](https://cmake.org/cmake/help/latest/cpack_gen/freebsd.html)
  - FreeBSD (.pkg) for macOS

- [CPack RPM Generator](https://cmake.org/cmake/help/latest/cpack_gen/rpm.html)
  - RPM (.rpm) for Red-Hat / CentOS

MikeMcQuaid (commented):
Revoking previous requested changes!

sharball-release.sh
MikeMcQuaid: My mistake, I'm sure this used to be supported but probably has been removed! 

Is there any other software that can be used to generate something like this?

Metolinam:
> Bağımsız bir sharball yükleyici oluşturmak için yeni bir komut dosyası ekleyin.
> 
> Yükleyici oluşturucuda:
> 
> 1. GItHub API'sinden en son Homebrew/brew etiketini alın
> 2. Homebrew/brew tarball'ı indirin
> 3. tarball'ı içine gömmek`install.sh`
> 
> Ayrıca [#634'e](https://github.com/Homebrew/install/issues/634) bakın .

XuehaiPan:sharball-installer

PR #644, opened 2022-04-27
It should be master and not HEAD

junedjabbar:
The url has HEAD which seems incorrect and 404 is coming. It should be master

### DIFF
--- a/install
+++ b/install
@@ -3,8 +3,8 @@
 STDERR.print <<EOS
 Warning: The Ruby Homebrew installer is now deprecated and has been rewritten in
 Bash. Please migrate to the following command:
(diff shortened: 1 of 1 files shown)


Bo98:
We're tracking this with GitHub as the change appears unexpected: #645.

> Strangely, when I used the correct link, I received this error:
> 
> ```
> error: Not a valid ref: refs/remotes/origin/master
> fatal: ambiguous argument 'refs/remotes/origin/master': unknown revision or path not in the working tree.
> Use '--' to separate paths from revisions, like this:
> 'git <command> [<revision>...] -- [<file>...]'
> ```

This though is typically indicative of a network issue. You can see https://github.com/orgs/Homebrew/discussions/666 for advice on that.

Bo98:
GitHub have rolled out a fix. Everything should be working now.

PR #657, opened 2022-05-15
unset `GIT_DIR` and `GIT_WORK_TREE`

branchvincent:
These environment variables allow overriding the `.git` and working directories for `git` commands. Before this change, a user with these set would cause our installer to clone the `Homebrew/brew` repo to `$GIT_WORK_TREE`

See https://git-scm.com/book/en/v2/Git-Internals-Environment-Variables

For a reproduction:

```sh
cd "$(mktemp -d)"
export GIT_DIR="$(pwd)/.git"
export GIT_WORK_TREE="$(pwd)"
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
ls -la
```

Note, I encountered this issue as a [`yadm`](https://github.com/TheLocehiliosan/yadm) user, which sets these to manage a dotfiles repo.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -187,7 +187,7 @@ REQUIRED_GIT_VERSION=2.7.0   # HOMEBREW_MINIMUM_GIT_VERSION in brew.sh in Homebr
 export HOMEBREW_NO_ANALYTICS_THIS_RUN=1
 export HOMEBREW_NO_ANALYTICS_MESSAGE_OUTPUT=1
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
I'm unsure about this. If you have these variables set you're going to break literally anything using `git`. Yeh, Homebrew/brew covers this with environment filtering but this feels like it's just gonna be a game of whack-a-mole to keep unsetting variables like this?

Issue #668, opened 2022-06-11
brew ignores HOMEBREW_* variables


ogerardin:
- [x] your problem was from running the official `install` or `uninstall` script?
- [X] after installation: ran `brew config` and `brew doctor` and included their output with your issue? If you couldn't install: provided your OS version with the output of your issue?

## What you were trying to do (and why)
I have installed Homebrew on a Synology NAS (I had to cheat a little bit with glibc check but that's not the point here).
As the home directories on DSM are in `/var/services/homes`, I have a symbolic link `/home -> /var/services/homes`, which itself resolves to `/volume2/homes` on my NAS.

## What happened (include command output)
Although Homebrew is effectively installed in `/home/linuxbrew/.linuxbrew`, it insists that it is installed in `/volume2/homes/linuxbrew/.linuxbrew`, which is bad because as you know it will compile stuff instead of using prebuilt binaries.

I thought overriding the environment variables set by `brew shellenv` in `~/.profile` would work, so instead of 
`eval "$(/home/linuxbrew/.linuxbrew/bin/brew shellenv)"`
I have explicitly defined all the variables that shellenv would set, replacing `/volume2/homes/linuxbrew` with  `/home/linuxbrew`:
```
export HOMEBREW_PREFIX="/home/linuxbrew/.linuxbrew";
export HOMEBREW_CELLAR="/home/linuxbrew/.linuxbrew/Cellar";
export HOMEBREW_REPOSITORY="/home/linuxbrew/.linuxbrew/Homebrew";
export PATH="/home/linuxbrew/.linuxbrew/bin:/home/linuxbrew/.linuxbrew/sbin${PATH+:$PATH}";
export MANPATH="/home/linuxbrew/.linuxbrew/share/man${MANPATH+:$MANPATH}:";
export INFOPATH="/home/linuxbrew/.linuxbrew/share/info:${INFOPATH:-}";
```
Unfortunately it doesn't change anything, as it seems the variables are ignored and brew always recomputes its real path...
So when I try to install gcc for example I get:
```
Warning: Building gcc from source as the bottle needs:
- HOMEBREW_CELLAR: /home/linuxbrew/.linuxbrew/Cellar (yours is /volume2/homes/linuxbrew/.linuxbrew/Cellar)
- HOMEBREW_PREFIX: /home/linuxbrew/.linuxbrew (yours is /volume2/homes/linuxbrew/.linuxbrew)
```
(I didn't let it finish)

[...]

PR #671, opened 2022-06-18
install.sh: warn for symlinked prefixes

XuehaiPan:
Check if the `HOMEBREW_PREFIX` contains symlinks.

1. If `HOMEBREW_PREFIX` or its parent directory is a symlink and exists, prompt a warning message.

Screenshot 1

2. If `HOMEBREW_PREFIX` or its parent directory is a dead symlink, warn and abort.

Screenshot 2

3. Do nothing otherwise.

See also:

- #668
- Homebrew/brew#13422

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -253,10 +253,12 @@ ring_bell() {
   fi
 }
 
+CONFIRM_MESSAGE="Press ${tty_bold}RETURN${tty_reset}/${tty_bold}ENTER${tty_reset} to continue or any other key to abort:"
+
 wait_for_user() {
   local c
   echo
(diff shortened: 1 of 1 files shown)


hkloudou:
@XuehaiPan Waiting for you merga new version,thanks

MikeMcQuaid (commented):
This seems like a lot of code for something we've only seen a single issue about? What problem does this solve to add this message?

MikeMcQuaid:
@XuehaiPan sorry, passing on this for now. Will reconsider if we see more users with these problems.

PR #692, opened 2022-09-07
install: Improve Docker detection

### LINKED ISSUE #621: Unable to install inside Docker container on macOS

I was trying to install Homebrew inside Docker container running on macOS host.

Install script aborts with message `Don't run this as root!`.

<details>
  <summary>Command output</summary>
  <pre>
    macOS 11.5.2, Docker Desktop 4.5.0 (74594)
    $ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
    Warning: Running in non-interactive mode because `$CI` is set.
    Don't run this as root!
  </pre>
</details>

Homebrew is successfully installed.

Install latest Docker Desktop for macOS, try installing Homebrew inside container.
This error happend due to `grep -E "azpl_job|actions_job|docker|garden|kubepods" -q /proc/1/cgroup` line inside install script. Seems like latest version of Docker uses cgroups v2.
```
# cat /proc/1/cgroup
0::/
```
This is expected behaviour, according to the [cgroups-v2 documentation](https://www.kernel.org/doc/Documentation/cgroup-v2.txt):
>If legacy cgroup is in use in the system, this file may contain multiple lines, one for each hierarchy. The entry for cgroup v2 is always in the format "0::$PATH"

EricFromCanada:
Fixes #621 with the changes made in https://github.com/Homebrew/actions/pull/305 and https://github.com/Homebrew/brew/pull/13814.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -297,6 +297,16 @@ version_lt() {
   [[ "${1%.*}" -lt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -lt "${2#*.}" ]]
 }
 
+check_run_command_as_root() {
+  [[ "${EUID:-${UID}}" == "0" ]] || return
+
+  # Allow Azure Pipelines/GitHub Actions/Docker/Concourse/Kubernetes to do everything as root (as it's normal there)
+  [[ -f /.dockerenv ]] && return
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Nice work, thanks @EricFromCanada!

PR #697, opened 2022-09-20
install: add newline to post install command

donvince:
When I installed homebrew, the command to update `~/.zprofile` outputted onto the same line as the previous command written to that file

image

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1018,7 +1018,7 @@ if [[ "$(which brew)" != "${HOMEBREW_PREFIX}/bin/brew" ]]
 then
   cat <<EOS
 - Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
@donvince that would only be if you don't have a trailing newline in this file. Almost all text editors add it so: passing on this for now, sorry.

donvince:
> @donvince that would only be if you don't have a trailing newline in this file. Almost all text editors add it so: passing on this for now, sorry.

Very kind of you to get back to me. 

I understand, it's reasonable to expect this file to have been previously terminated with a new line. For info, in this case, I hadn't at that stage used a text editor: I'm installing a new laptop and I presume the previous install left that file lacking a new line. I'll try and track down where that comes from too and arrange a fix there. I figured it worth treating here too mind as homebrew was ultimately what was affected (as in running `brew --version` in a new command window gave a not found) and this adds a tiny bit of defensiveness.

Issue #688, opened 2022-08-20
/usr/bin/install-info may not exist 


colindean:
- [X] your problem was from running the official `install` or `uninstall` script?
- [X] after installation: ran `brew config` and `brew doctor` and included their output with your issue? If you couldn't install: provided your OS version with the output of your issue?

## What you were trying to do (and why)

Uninstall homebrew as a part of migrating from `~/.linuxbrew` to `/home/linuxbrew` to avoid mass recompliation on a slow machine!

## What happened (include command output)

<details>
  <summary>Command output</summary>
  <pre>
$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/uninstall.sh)"
/bin/bash: line 237: /home/linuxbrew/.linuxbrew/bin/brew: No such file or directory
Warning: This script will remove:
/home/colin/.cache/Homebrew/
/home/colin/.linuxbrew/bin/brew -> /home/colin/.linuxbrew/Homebrew/bin/brew
/home/colin/.linuxbrew/Caskroom/
/home/colin/.linuxbrew/Cellar/
/home/colin/.linuxbrew/etc/bash_completion.d/brew -> /home/colin/.linuxbrew/Homebrew/completions/bash/brew
/home/colin/.linuxbrew/Homebrew/
/home/colin/.linuxbrew/Homebrew/bin/brew
/home/colin/.linuxbrew/Homebrew/CHANGELOG.md
/home/colin/.linuxbrew/Homebrew/completions/
/home/colin/.linuxbrew/Homebrew/CONTRIBUTING.md
/home/colin/.linuxbrew/Homebrew/Dockerfile
/home/colin/.linuxbrew/Homebrew/.dockerignore
/home/colin/.linuxbrew/Homebrew/docs/
/home/colin/.linuxbrew/Homebrew/.editorconfig
/home/colin/.linuxbrew/Homebrew/.github/
/home/colin/.linuxbrew/Homebrew/.gitignore
/home/colin/.linuxbrew/Homebrew/Library//
/home/colin/.linuxbrew/Homebrew/LICENSE.txt
/home/colin/.linuxbrew/Homebrew/manpages/
/home/colin/.linuxbrew/Homebrew/README.md
/home/colin/.linuxbrew/Homebrew/.shellcheckrc
/home/colin/.linuxbrew/Homebrew/.sublime/
/home/colin/.linuxbrew/Homebrew/.vale.ini
[...]

PR #699, opened 2022-09-27
Add a temporary patch for missing install-info on some systems

marcus-crane:
Closes: https://github.com/Homebrew/install/issues/688

CleanShot 2022-09-28 at 00 04 34@2x

I don't love this PR but it seems to do the trick as a workaround for the lack of `install-info` on some systems.

Since the uninstall script implicitly depends on `install-info` already, I opted to just make it an explicit requirement instead of anything fancy.

I've also changed the hardcoding of `/usr/bin/install-info` to instead rely on the user's `PATH` given that `texinfo` (which provides `install-info`) is keg-only and not reachable via the default Homebrew pathing.

Anyway, I've nuked my Brew installation successfully with this script so it should be good to go 🙂

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -25,6 +25,16 @@ then
   abort 'Bash must not run in POSIX mode. Please unset POSIXLY_CORRECT and try again.'
 fi
 
+if ! command -v install-info &> /dev/null
+then
+  cat <<EOS
+install-info is needed as part of the uninstall process.
+Please install the texinfo package (via Homebrew or your system package manager).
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (commented):
Sorry, I don't think this is a great approach. Would rather figure out how to avoid the need for `install-info` entirely.

marcus-crane:
No worries, I wouldn't intend for this to be a final result by any means but just a bandaid until something better is in place but happy to opt for something more polished off the bat. I'll see what I can come up with once I get all of my tooling installed again

PR #702, opened 2022-10-06
install: only support default prefix on Linux.

MikeMcQuaid:
The experience in a non-default experience is very poor. Let's not encourage it.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -112,15 +112,17 @@ OS="$(uname)"
 if [[ "${OS}" == "Linux" ]]
 then
   HOMEBREW_ON_LINUX=1
-elif [[ "${OS}" != "Darwin" ]]
+elif [[ "${OS}" == "Darwin" ]]
 then
+  HOMEBREW_ON_MACOS=1
+else
   abort "Homebrew is only supported on macOS and Linux."
 fi
 
 # Required installation paths. To install elsewhere (which is unsupported)
 # you can untar https://github.com/Homebrew/brew/tarball/master
 # anywhere you like.
-if [[ -z "${HOMEBREW_ON_LINUX-}" ]]
+if [[ -n "${HOMEBREW_ON_MACOS-}" ]]
 then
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -507,41 +509,15 @@ fi
...
-    fi
-    trap - SIGINT
-  fi
-  HOMEBREW_REPOSITORY="${HOMEBREW_PREFIX}/Homebrew"
+elif ! [[ -w "${HOMEBREW_PREFIX}" ]] &&
+     ! [[ -w "/home/linuxbrew" ]] &&
+     ! [[ -w "/home" ]] &&
+     ! have_sudo_access
+then
+  abort "Insufficient permissions to install Homebrew to \"${HOMEBREW_PREFIX}\"."
```
XuehaiPan, line 520: > Let's not encourage it.

Should we add a link <https://docs.brew.sh/Installation#alternative-installs> here to tell people how to install Homebrew without `sudo` privilege? One of the reasons the user using Homebrew is they don't have `sudo` privilege.

carlocab (dismissed):
I'd avoid pointing users to the alternative installs page for now, but consider adding it if we get multiple issues that could have been avoided if we point them there.

Alternatively, we should just link to our documentation on installation on any type of failure so help users work out what to do for themselves.

install.sh
XuehaiPan: > It'll be clearer that this is unsupported after Homebrew/brew#13971

I see, we only support default prefixes both on macOS and Linux now. If non-default prefixes are not officially supported, does that means we will not ship relocatable bottles in the future or the priority is lowered?

MikeMcQuaid: @XuehaiPan We've only ever supported default prefix/bottles on Linux, we've just been bad at wording it.

If we can get to the point where we have roughly >=90% bottles relocatable outside the default prefix: we'll reconsider what we support.

Our support is pretty much based on "on what configurations can we continue to provide bottles".

PR #703, opened 2022-10-08
install: force symlink creation

gromgit:
Several users have set `core.symlinks = false` at global level, likely in a WSL context. This breaks Aliases, among other things.

Also ensure consistency in `git config` invocations with `brew update*`.

Ref: https://github.com/Homebrew/homebrew-core/issues/112539


carlocab (approved):

MikeMcQuaid (approved):
Makes sense to me, thanks @gromgit!

If we get any more of these, may want to consider having a method that sets them all consistently between both repos.

PR #706, opened 2022-10-29
install.sh: Fix broken docs link for ARM Linux

issyl0:
- I found it a little misleading that we said "alternate installation methods" but the "ARM unsupported" docs page doesn't refer the user to those.

- I am tempted to improve this situation with an "if you know what you are doing and won't ask for help, you can install Homebrew on ARM Linux", and gate it also with `HOMEBREW_DEVELOPER` or something? Since in my experiments with Homebrew on ARM Linux I had to comment out line 555 ("`else` variously abort") to get past that message. Yet the docs say that while it may be _unsupported_ (fine), it is _possible_, which the installer currently doesn't respect.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -559,8 +559,7 @@ else
     abort "$(
       cat <<EOABORT
 Homebrew on Linux is not supported on ARM processors.
-You can try an alternate installation method instead:
-  ${tty_underline}https://docs.brew.sh/Homebrew-on-Linux#arm${tty_reset}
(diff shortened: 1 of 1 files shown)


Bo98:
> * I am tempted to improve this situation with an "if you know what you are doing and won't ask for help, you can install Homebrew on ARM Linux", and gate it also with `HOMEBREW_DEVELOPER` or something?

The installer has always been quite strict. For example, 10.11-10.15 works absolutely fine, albeit no bottles, but we also have a hard block there that I edit out when setting up CI.

Bo98 (approved):

SMillerDev (approved):

PR #709, opened 2022-11-10
Do not use sudo for GUI installation of Dev Tools

vitorgalvao:
`/usr/bin/xcode-select --install` does not require `sudo`, it works fine without it.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -864,7 +864,7 @@ fi
 if should_install_command_line_tools && test -t 0
 then
   ohai "Installing the Command Line Tools (expect a GUI popup):"
-  execute_sudo "/usr/bin/xcode-select" "--install"
+  execute "/usr/bin/xcode-select" "--install"
   echo "Press any key when the installation has completed."
   getc
   execute_sudo "/usr/bin/xcode-select" "--switch" "/Library/Developer/CommandLineTools"


MikeMcQuaid (approved):

install.sh
```diff
@@ -864,7 +864,7 @@ fi
 if should_install_command_line_tools && test -t 0
 then
   ohai "Installing the Command Line Tools (expect a GUI popup):"
-  execute_sudo "/usr/bin/xcode-select" "--install"
+  execute "/usr/bin/xcode-select" "--install"
   echo "Press any key when the installation has completed."
   getc
   execute_sudo "/usr/bin/xcode-select" "--switch" "/Library/Developer/CommandLineTools"
```
MikeMcQuaid, line 870: @vitorgalvao Given this requires/uses `sudo` immediately afterwards: is there any real advantage of not using `sudo` here?

vitorgalvao: > is there any real advantage of not using `sudo` here?

Installing the Command Line Tools takes a while. Especially if someone has lowered the timeout, that means entering the password one more time than necessary.

Though the motivation for the change was that we tend to avoid `sudo` unless absolutely necessary.

If (for example) we ever drop the `--switch` line (e.g. `xcode-select` changes behaviour and it’s no longer necessary), chances are we’d just assume the previous `--install` line was still correct in using `sudo` and we’d continue to let it be even if technically at that point it should get the change.

Granted, that last one is unlikely. But there’s no big advantage to having the `sudo` for the `--install` either: even if it doesn’t timeout, the user will have to return to the terminal anyway to “press any key” and keep going.

MikeMcQuaid (approved):
@vitorgalvao I'm sold!

PR #712, opened 2022-11-15
install.sh: fix "printf: - : invalid option" & format

dirname:
- Expect: 

Screenshot 2022-11-16 at 2 54 19 AM

- Actually:

Screenshot 2022-11-16 at 2 53 41 AM

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1019,9 +1019,9 @@ then
   then
     plural="s"
   fi
-  printf "- Run these commands in your terminal to add the non-default Git remote%s for %s:" "${plural}" "${non_default_repos}"
-  printf "    echo '# Set PATH, MANPATH, etc., for Homebrew.' >> %s" "${shell_profile}"
-  printf "    echo '%s' >> %s\n" "${additional_shellenv_commands[@]}" "${shell_profile}"
+  printf -- "- Run these commands in your terminal to add the non-default Git remote%s for %s:\n" "${plural}" "${non_default_repos}"
+  printf "    echo '# Set PATH, MANPATH, etc., for Homebrew.' >> %s\n" "${shell_profile}"
+  printf "    echo '%s' >> ${shell_profile}\n" "${additional_shellenv_commands[@]}"
   printf "    %s\n" "${additional_shellenv_commands[@]}"
 fi
 


install.sh
```diff
@@ -1019,9 +1019,9 @@ then
   then
     plural="s"
   fi
-  printf "- Run these commands in your terminal to add the non-default Git remote%s for %s:" "${plural}" "${non_default_repos}"
-  printf "    echo '# Set PATH, MANPATH, etc., for Homebrew.' >> %s" "${shell_profile}"
-  printf "    echo '%s' >> %s\n" "${additional_shellenv_commands[@]}" "${shell_profile}"
+  echo "- Run these commands in your terminal to add the non-default Git remote${plural} for ${non_default_repos}:"
```
SMillerDev, line 1022: Why replace it with an echo only for the first line?

dirname: I refer to the code of the last macOS version (Monterey):  https://github.com/Homebrew/install/commit/cc459e4c030b5a6a4aa42e0ffdd13e54cedbcddb

MikeMcQuaid: @dirname Sorry, I don't understand the justification here. What's your `brew config` output?

dirname: > @dirname Sorry, I don't understand the justification here. What's your `brew config` output?

I set `HOMEBREW_BREW_GIT_REMOTE`, `HOMEBREW_CORE_GIT_REMOTE` to use the mirror to install brew, and then an error occurred as described in the picture `/bin/bash: line 1021: printf: - : invalid option`. This may be caused by `/bin/bash -c`, so I refer to the code that the previous version can install.

https://github.com/Homebrew/install/blob/684032061383ab14862240aaff4273cde4e1847c/install.sh#L1022

Similar error: 
`/bin/bash -c 'printf "hello"'` will print `hello` normally, but `/bin/bash -c 'printf "-hello"'` will cause same error.

image

MikeMcQuaid: I'd suggest we use `printf -- "-hello"` instead which will fix this more simply and not require changing as much.

install.sh
```diff
@@ -1019,9 +1019,9 @@ then
   then
     plural="s"
   fi
-  printf "- Run these commands in your terminal to add the non-default Git remote%s for %s:" "${plural}" "${non_default_repos}"
-  printf "    echo '# Set PATH, MANPATH, etc., for Homebrew.' >> %s" "${shell_profile}"
-  printf "    echo '%s' >> %s\n" "${additional_shellenv_commands[@]}" "${shell_profile}"
+  printf -- "- Run these commands in your terminal to add the non-default Git remote%s for %s:\n" "${plural}" "${non_default_repos}"
+  printf "    echo '# Set PATH, MANPATH, etc., for Homebrew.' >> %s\n" "${shell_profile}"
+  printf "    echo '%s' >> ${shell_profile}\n" "${additional_shellenv_commands[@]}"
```
MikeMcQuaid, line 1024: Can you explain this change? I'm not sure I understand it.

install.sh
maxim-belkin: > I'd suggest we use `printf -- "-hello"`

This looks like the safest solution. Alternative ones are to use `echo` (as was suggested originally), replace `-` with `\055`, or pass the message as a parameter (`printf "%s\n" "- Run these"`)

I think it's a perfect time for a reminder that in Bash double quotes are used to prevent word splitting, not to indicate a string (as in other programming languages). So, things like `var=""` are meaningless and the proper way to write the same in Bash is `var=`.

install.sh
maxim-belkin: Addition of `\n` on line 1023 is 👍🏻.
Change on line 1024 seems to be unnecessary.

dirname: > Can you explain this change? I'm not sure I understand it.

When multiple environment variables are set, it looks like this:

snap

MikeMcQuaid: Ok, thanks.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @dirname!

Issue #724, opened 2023-01-10
Short flag doesn't support space after option


n8henrie:
# Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.

- [x] your problem was from running the official `install` or `uninstall` script?
- [x] after installation: ran `brew config` and `brew doctor` and included their output with your issue? If you couldn't install: provided your OS version with the output of your issue?

## What you were trying to do (and why)

Remove an old Intel (`/usr/local`) installation from an M1 that's working great from `/opt/homebrew`

## What happened (include command output)

``` console
$ wget -q https://raw.githubusercontent.com/Homebrew/install/HEAD/uninstall.sh
$ sha256sum uninstall.sh
83cbc1ce42adac22a53edcfa4cd11907c0da747d00ad556ec2ea6c84d331e602  uninstall.sh
$ bash uninstall.sh -n -p /usr/local
Warning: Unrecognized option: '/usr/local'
...
```

## What you expected to happen

It seems conventional to allow an optional space after a short flag (`--foo=bar` and `-f bar`); I expected `-p /usr/local/` to work the same as `--path=/usr/local/` or `-p/usr/local` (both of which work).

## Step-by-step reproduction instructions (by running `brew` commands)

As per above

I'd be happy to make a PR to this effect if it would be helpful; I think all that would be required would be an extra case statement for the short and long `-p`:

```
    case "${opt}" in
      -p)
        homebrew_prefix_candidates+=("$2")
        shift
        ;;
      --path)
        homebrew_prefix_candidates+=("$2")
        shift
        ;;
      # continue with -p*)
```

SMillerDev:
If you have a fix a pull request would be appreciated

PR #725, opened 2023-01-10
Support optional space after `-p` / `--path`

n8henrie:
Fixes https://github.com/Homebrew/install/issues/724

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -222,6 +222,14 @@ EOS
 while [[ $# -gt 0 ]]
 do
   case "$1" in
+    -p)
+      homebrew_prefix_candidates+=("$2")
+      shift
+      ;;
+    --path)
+      homebrew_prefix_candidates+=("$2")
+      shift
+      ;;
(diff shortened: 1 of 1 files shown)


MikeMcQuaid:
Thanks @n8henrie! Does `install.sh` already work with this format for any arguments (if there are any where it applies)? If not: could it be adjusted similarly?

n8henrie:
@MikeMcQuaid `install.sh` is a much more complicated script! Looks like in part due to handling a lot of different environments, architectures, and shells (I only use bash).

A quick run through, I don't see any relevant `case` statements so we can say it is not using an *identical* method of argument handling. In fact, I'm not sure I see any configurable options other than `INTERACTIVE` / `NONINTERACTIVE`, which look like they are set solely from the environment.

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @n8henrie!

PR #733, opened 2023-01-29
Use `/home/${USER}` instead of `${HOME}`

sdwolfz:
While trying to automate installing linuxbrew in an ansible playbook I noticed that all went well, apart from creating the `.cache` directory in the user's home.

This is because of multiple reasons:
1. To run in ansible I need it to `export NONINTERACTIVE=t`
2. To be able to execute it with sudo privileges through the ansible `become` functionality I need to already be root, and use sudo to switch back to the oriningal user.
3. It requires `-E` to preserve non-interactive
4. Because of this `-u` switches `${USER}` correctly but preserves `${HOME}` from `root`.

The workaround was to `export HOME=/home/{{ ansible_user_id }}` before execution, but honestly this should not be needed.

For reference, this is the ansible task that I wrote:

```yml
- name: Install homebrew
  ansible.builtin.shell: |
    set -eo pipefail

    export NONINTERACTIVE=1
    export HOME=/home/{{ ansible_user_id }}

    sudo -E -u {{ ansible_user_id }} `pwd`/install.sh
  args:
    chdir: '{{ homebrew_registered_repository_location.stdout }}'
  when: not homebrew_already_installed.stat.exists
  become: true
```

Unfortunately I could not find a better way to be able to automate this using ansible, reason why I'm proposing this change. If we replace the usage of `${HOME}` with `/home/${USER}` it should be identical to the current behaviour.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -112,9 +112,11 @@ OS="$(uname)"
 if [[ "${OS}" == "Linux" ]]
 then
   HOMEBREW_ON_LINUX=1
+  USER_HOME="/home/${USER}"
 elif [[ "${OS}" == "Darwin" ]]
 then
   HOMEBREW_ON_MACOS=1
+  USER_HOME="/Users/${USER}"
 else
   abort "Homebrew is only supported on macOS and Linux."
 fi
@@ -136,7 +138,7 @@ then
     HOMEBREW_PREFIX="/usr/local"
     HOMEBREW_REPOSITORY="${HOMEBREW_PREFIX}/Homebrew"
   fi
-  HOMEBREW_CACHE="${HOME}/Library/Caches/Homebrew"
+  HOMEBREW_CACHE="${USER_HOME}/Library/Caches/Homebrew"
 
   STAT_PRINTF=("stat" "-f")
   PERMISSION_FORMAT="%A"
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -112,9 +112,11 @@ OS="$(uname)"
 if [[ "${OS}" == "Linux" ]]
 then
   HOMEBREW_ON_LINUX=1
+  USER_HOME="/home/${USER}"
```
XuehaiPan, line 115: `$HOME` on Linux is not always `/home/$USER`

SMillerDev (changes requested):
That variable is supposed to contain your home directory. If it does not that's not up to Homebrew to fix.

dawidd6:
Instead of relying on the installation script, you can leverage instructions from https://docs.brew.sh/Installation#untar-anywhere-unsupported and automate in Ansible around that.

sdwolfz:
How about an alternative approach then? A way to force non-interactive mode, like from a command line argument? That way we could have at the beginning of the script:

```sh
# Process command line arguments
while [ $# -gt 0 ]
do
  key="$1"
  case "$key" in
    --non-interactive)
      NONINTERACTIVE=1
      shift
      ;;
    *)
      shift
      ;;
  esac
done
```

The ansible task can become:

```yml
- name: Install homebrew
  ansible.builtin.shell: |
    set -eo pipefail

    sudo -u {{ ansible_user_id }} `pwd`/install.sh --non-interactive
  args:
    chdir: '{{ homebrew_registered_repository_location.stdout }}'
  when: not homebrew_already_installed.stat.exists
  become: true
```

gromgit:
Does:
```
sudo -u {{ ansible_user_id }} NONINTERACTIVE=1 `pwd`/install.sh
```
not work?

sdwolfz:
For some reason it did not work initially when I tried it out (probably because I was using command instead of shell in ansible). Yes this works now.

Sorry for the trouble.

PR #742, opened 2023-02-19
Check if Homebrew path evaluation statement is already in shell profile

UnrealApex:
Check if Homebrew is already in a user's `$PATH`. If it already is, then do not provide instructions to add it.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -998,7 +998,7 @@ esac
 if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}" && ! type brew &>/dev/null
 then
   cat <<EOS
-- Homebrew was already found in your PATH, run this command to activate it in your current shell:
+- Run this command in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
     eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
 EOS
 elif ! grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
--- a/install.sh
+++ b/install.sh
@@ -996,7 +996,7 @@ esac
 
 cat <<EOS
 - Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
-    grep -qsF 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"' ${shell_profile} || (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
+    grep --quiet --no-messages --fixed-strings 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"' ${shell_profile} || (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
     eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
 EOS
 if [[ -n "${non_default_repos}" ]]
--- a/install.sh
+++ b/install.sh
@@ -994,22 +994,16 @@ case "${SHELL}" in
     ;;
 esac
 
-# show different instructions on adding Homebrew to path based on if the user has already done so
-if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}" && ! type brew &>/dev/null
+# don't show how to add Homebrew to path if the user has already done so
+if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
 then
-  cat <<EOS
-- Run this command in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
-    eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
-EOS
-elif ! grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
-then
+  :
+else
   cat <<EOS
 - Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
     (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
     eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
 EOS
-else
-  :
 fi
 
 if [[ -n "${non_default_repos}" ]]
--- a/install.sh
+++ b/install.sh
@@ -998,7 +998,7 @@ in_shell_profile=$(grep -s "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "
 
 # FIXME: figure out why this give a "command not found" error
 # show different instructions on adding Homebrew to path based on if the user has already done so
-if "${in_shell_profile}" && ! type brew &>/dev/null
+if [[ -n $in_shell_profile ]] && ! type brew &>/dev/null
 then
   cat <<EOS
 - Run this command in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
--- a/install.sh
+++ b/install.sh
@@ -1001,7 +1001,7 @@ cat <<EOS
 - Homebrew was already found in your PATH, run this command to activate it in your current shell:
     eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
 EOS
-elif ! grep -q "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
+elif ! grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
 then
 cat <<EOS
 - Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
--- a/install.sh
+++ b/install.sh
@@ -994,7 +994,6 @@ case "${SHELL}" in
     ;;
 esac
 
-# show different instructions on adding Homebrew to path based on if the user has already done so
 if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
 then
   if ! [[ -x "$(command -v brew)" ]]
--- a/install.sh
+++ b/install.sh
@@ -995,7 +995,7 @@ case "${SHELL}" in
 esac
 
 # show different instructions if the user already has Homebrew in their PATH
-if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}" && ! type brew >/dev/null 2>/dev/null
+if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}" && ! type brew &>/dev/null
 then
 cat <<EOS
 - Homebrew was already found in your PATH, run this command to activate it in your current shell:
--- a/install.sh
+++ b/install.sh
@@ -997,19 +997,19 @@ esac
 # show different instructions on adding Homebrew to path based on if the user has already done so
 if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}" && ! type brew &>/dev/null
 then
-cat <<EOS
+  cat <<EOS
 - Homebrew was already found in your PATH, run this command to activate it in your current shell:
     eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
 EOS
 elif ! grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
 then
-cat <<EOS
+  cat <<EOS
 - Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
     (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
     eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
 EOS
 else
-cat <<EOS
+  cat <<EOS
 - Homebrew is already set up. Your system is ready to brew.
 EOS
 fi
--- a/install.sh
+++ b/install.sh
@@ -994,25 +994,24 @@ case "${SHELL}" in
     ;;
 esac
 
-in_shell_profile=$(grep -s "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}")
-
-# FIXME: figure out why this give a "command not found" error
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -996,7 +996,7 @@ esac
 
 cat <<EOS
 - Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
-    (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
+    grep -qsF 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"' ${shell_profile} || (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
```
vitorgalvao, line 999: This makes the preceding line potentially incorrect/confusing, in that it’ll say “Run these **two** commands” while showing one. Also, command flags should be expanded in scripts so they’re easier to understand: `--quiet --no-messages --fixed-strings`.

UnrealApex: > This makes the preceding line potentially incorrect/confusing, in that it’ll say “Run these two commands” while showing one.

Could you clarify what you mean?

> Also, command flags should be expanded in scripts so they’re easier to understand: --quiet --no-messages --fixed-strings

Good pro-tip. I will use the full flag names.

vitorgalvao: > Could you clarify what you mean?

See the line before your code. I says:

```shell
```

**Two** commands, it says. However, what you change does is make it so one of the two commands does not display in a specific condition (already having the `eval` line). Which means that users who meet that condition will be told to run two commands but only see **one**.

That is a possible source of confusion, which leads to wasted time, which leads to frustration, which leads to the Dark Side. Er, I mean it leads to avoidable bug reports.

UnrealApex: Even if the `eval` line is present in the user's shell profile, the command would still display regardless of the boolean returned by the grep. This is because the two lines are not evaluated; they are just printed to stdout in using a `cat` heredoc.

gromgit: I think @vitorgalvao means you should instead:
1. _parameterize_  "these two commands" and the `brew shellenv` injection string, so that they can be replaced with "this command" and the empty string respectively if the `brew shellenv` line already exists, and
2. test for `brew shellenv` existence _before_ the `cat`, instead of leaving it to users to run the test themselves

The second step above is _critical_; it's to be expected that some users will type in commands manually instead of copy-n-pasting, so complex command lines like what you're proposing are not a good idea.

There's also no technical reason for the installer to avoid doing the test on the user's behalf. The probability of a `brew shellenv` magically appearing in a user's profile, between the time they see those commands and actually run them, is essentially zero.

install.sh
```diff
@@ -994,11 +994,21 @@ case "${SHELL}" in
     ;;
 esac
 
+# show different instructions if the user already has Homebrew in their PATH
+if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
+then
+cat <<EOS
+- Homebrew was already found in your PATH, run this command to activate it in your current shell:
```
MikeMcQuaid, line 1001: Should we check if it's actually already been run?

gromgit: I think that's an orthogonal concern, based on the various permutations of "I've added `shellenv` to my profile but my PATH's still wrong". If the problem to be solved is "reliably add `shellenv` to user's profile", methinks the necessary and sufficient test is "has it already been added?"

UnrealApex: Perhaps that might be the better approach for this. If Homebrew was already in their current shell environment, I would assume the user would probably be aware of it.

install.sh
```diff
@@ -995,18 +995,23 @@ case "${SHELL}" in
...
+elif ! grep -q "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
+then
 cat <<EOS
 - Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
     (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
     eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
 EOS
+else
+cat <<EOS
+- Homebrew is already set up. Your system is ready to brew.
```
MikeMcQuaid, line 1013: Is there any chance more steps will be output below? If so, we should probably not do this.

install.sh
```diff
@@ -995,18 +995,23 @@ case "${SHELL}" in
 esac
 
 # show different instructions if the user already has Homebrew in their PATH
-if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
+if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}" && ! type brew >/dev/null 2>/dev/null
```
MikeMcQuaid, line 998:
```suggestion
if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}" && ! type brew &>/dev/null
```

install.sh
UnrealApex: No, more steps will not display in that case.

MikeMcQuaid: What about the `Run these commands in your terminal`, `Install Homebrew's dependencies`, etc. below?

MikeMcQuaid (commented):
One more comment and CI is failing.

install.sh
UnrealApex: My bad for not seeing that. I will change the wording.

UnrealApex: > If so, we should probably not do this.

By not doing this, I infer you mean remove the prompt in the case Homebrew is set up. Correct?

MikeMcQuaid: Display nothing in this case, I think, sorry!

install.sh
```diff
@@ -994,11 +994,24 @@ case "${SHELL}" in
     ;;
 esac
 
-cat <<EOS
+# show different instructions on adding Homebrew to path based on if the user has already done so
+if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}" && ! type brew &>/dev/null
```
MikeMcQuaid, line 998:
```suggestion
if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}" && type brew &>/dev/null
```

As this says `brew` was found.

install.sh
```diff
@@ -994,11 +994,24 @@ case "${SHELL}" in
...
+then
+  cat <<EOS
+- Homebrew was already found in your PATH, run this command to activate it in your current shell:
+    eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
+EOS
+elif ! grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
```
MikeMcQuaid, line 1004: It'd be good not to run this same `grep` twice.

install.sh
UnrealApex: Removing the negation would make the condition evaluate as false since `brew` would not be found. I will change the prompt wording.

UnrealApex: I do not think this block is necessary because `brew` will already be in PATH. I will rewrite this logic

install.sh
```diff
@@ -994,11 +994,27 @@ case "${SHELL}" in
     ;;
 esac
 
-cat <<EOS
+in_shell_profile=$(grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}")
+
+# FIXME: figure out why this give a "command not found" error
```
MikeMcQuaid, line 999: Which command is not found?

install.sh
```diff
@@ -994,11 +994,27 @@ case "${SHELL}" in
...
+# show different instructions on adding Homebrew to path based on if the user has already done so
+if "${in_shell_profile}" && ! type brew &>/dev/null
+then
+  cat <<EOS
+- Run this command in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
+    eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
+EOS
+elif ! "${in_shell_profile}"
```
MikeMcQuaid, line 1007: If you use two `if`s here you can avoid checking the same condition multiple times, no?

install.sh
UnrealApex: The error occurs when using the `$in_shell_profile` variable in the if statements.

install.sh
UnrealApex: Yeah, I could use a nested if. I will do so.

vitorgalvao: By using the nested `if` you can again make it a direct command (`if grep`) instead of saving to a variable, which should handle the error you got above.

install.sh
```diff
@@ -994,11 +994,27 @@ case "${SHELL}" in
     ;;
 esac
 
-cat <<EOS
+in_shell_profile=$(grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}")
```
gromgit, line 997: `grep -qs` suppresses _all_ output, so `in_shell_profile` is always set to the empty string, which explains the "command not found" error you're seeing.

```suggestion
in_shell_profile=$(grep -s "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}")
```

install.sh
gromgit: That's because `grep -qs` always returns nothing.

install.sh
```diff
@@ -994,11 +994,27 @@ case "${SHELL}" in
...
+
+# FIXME: figure out why this give a "command not found" error
+# show different instructions on adding Homebrew to path based on if the user has already done so
+if "${in_shell_profile}" && ! type brew &>/dev/null
```
gromgit, line 1001:
```suggestion
if [[ -n $in_shell_profile ]] && ! type brew &>/dev/null
```
This requires the `grep` fix I mentioned above.

install.sh
vitorgalvao: Assigning to a variable and checking its value is in general less robust than directly checking for the condition (`if grep`) which returns *the exit code*, the true marker of a command’s success/failure.

With @MikeMcQuaid’s suggestion below of a nested `if`, the command can again be used directly instead of assigned to a variable.

install.sh
UnrealApex: I see. That would make more sense to do now since we are only comparing it once.

install.sh
```diff
@@ -994,11 +994,26 @@ case "${SHELL}" in
...
+# show different instructions on adding Homebrew to path based on if the user has already done so
+if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
+then
+  if ! type brew &>/dev/null
+  then
+    cat <<EOS
+    - Run this command in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
+      eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
+EOS
+  else
```
vitorgalvao, line 1006: Why the `else` not doing anything? Having an `else` is not mandatory.

UnrealApex: I know. I included it because it made the code look better than just having an if statement without the else. Shall I remove it?

gromgit (dismissed):

install.sh
```diff
@@ -994,11 +994,24 @@ case "${SHELL}" in
...
 esac
 
-cat <<EOS
-- Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
-    (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
-    eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
+# show different instructions on adding Homebrew to path based on if the user has already done so
+if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
+then
+  if ! type brew &>/dev/null
```
XuehaiPan, line 1000:
```suggestion
  if ! type -P brew &>/dev/null
```

add `-P` option for force `PATH` lookup to avoid looking for alias and functions.

install.sh
```diff
@@ -994,11 +994,24 @@ case "${SHELL}" in
...
-    (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
-    eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
+# show different instructions on adding Homebrew to path based on if the user has already done so
+if grep -qs "eval \"\$(${HOMEBREW_PREFIX}/bin/brew shellenv)\"" "${shell_profile}"
+then
+  if ! type -P brew &>/dev/null
+  then
+    cat <<EOS
+    - Run this command in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
+      eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
```
XuehaiPan, line 1004:
```suggestion
- Run this command in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
    eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
```

extra spaces.

install.sh
```diff
@@ -994,11 +994,24 @@ case "${SHELL}" in
...
 EOS
+  fi
+else
+  cat <<EOS
+  - Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
+      (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
+      eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
```
XuehaiPan, line 1011:
```suggestion
- Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
    (echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"') >> ${shell_profile}
    eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"
```

MikeMcQuaid (dismissed):
This now makes sense to me, thanks! Will get a few more reviews and then ship.

XuehaiPan (commented):
LGTM. One small suggestion is to change `type -P` to `command -v`. It is more POSIX-compliant and consistent with other conditions in this script.

```diff
- if ! type -P brew &>/dev/null
+ if [[ -x "$(command -v brew)" ]]
```

since we only support `bash`, so they are identical here.

vitorgalvao (dismissed):

install.sh
```diff
@@ -994,11 +994,24 @@ case "${SHELL}" in
     ;;
 esac
 
-cat <<EOS
+# show different instructions on adding Homebrew to path based on if the user has already done so
```
UnrealApex, line 997: Should I change the wording of this comment?

gromgit: Since the intent is (well, should be) clear from the code, I'm not sure it serves a useful purpose to begin with.

UnrealApex: Shall I remove it then?

gromgit: I would, but I'm not a maintainer. :grinning:

carlocab (approved):

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks again @UnrealApex! Sorry it took so long!

PR #743, opened 2023-02-20
add uninstall help text to README

gabroo:
add the output of `bash uninstall.sh --help` so you don't have to download it to see


gromgit:
Sorry, but I don't see the utility of this:
1. `-p, --path=PATH Sets Homebrew prefix. Defaults to /home/linuxbrew/.linuxbrew.`: the defaults on macOS Intel and macOS M1 are _very_ different.
2. All the flags can only be used with a local copy of the script, i.e. you'd have to download it anyway.

MikeMcQuaid:
Agreed, sorry, passing on this as it'll get out of sync very quickly.

PR #747, opened 2023-02-23
install.sh: ensure `master` is used for `git init`

ZhongRuoyu:
This is not the case when `init.defaultBranch` is set to something else, like `main`. And in that case, if `HOMEBREW_DEVELOPER` is also set, the `Homebrew/brew` repository and `homebrew/core` (if tapped) will stay on that branch instead of `master` (and won't be updated anymore since there will be no tracking information available).

Also, for consistency, replace the short option `-q` with `--quiet`, the long one.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -887,7 +887,7 @@ ohai "Downloading and installing Homebrew..."
   cd "${HOMEBREW_REPOSITORY}" >/dev/null || return
 
   # we do it in four steps to avoid merge errors when reinstalling
-  execute "git" "init" "-q"
+  execute "git" "init" "--quiet" "--initial-branch" "master"
 
   # "git remote add" will fail if the remote is defined in the global config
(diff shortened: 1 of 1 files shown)


carlocab (approved):

MikeMcQuaid:
Great catch, thanks @ZhongRuoyu!

MikeMcQuaid:
@ZhongRuoyu Sorry, had to revert in #749 due to reason in #748.

As @Bo98 pointed out: https://github.com/Homebrew/brew/blob/76b87c4a00ed636bcf59c104b85863637609b66d/Library/Homebrew/dev-cmd/tap-new.rb#L162 is probably a better approach here.

Similarly, may want to add CI for Ubuntu 18.04.

ZhongRuoyu:
> @ZhongRuoyu Sorry, had to revert in #749 due to reason in #748.

No worries; it was on me. Sorry for the trouble!

PR #748, opened 2023-02-23
use initial-branch option for only git versions that support it

An-GG:
Script fails on Ubuntu 20.04 LTS, and probably a lot more since a recent version of git (required for the git init --inital-branch option) is not listed as a installation dependency on [brew.sh](brew.sh)

Honestly suprised #747 merged without failing CI, suprised they only check for compatibility with latest and not oldest LTS, anyway

this just checks the option is available before using it.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -887,7 +887,13 @@ ohai "Downloading and installing Homebrew..."
   cd "${HOMEBREW_REPOSITORY}" >/dev/null || return
 
   # we do it in four steps to avoid merge errors when reinstalling
-  execute "git" "init" "--quiet" "--initial-branch" "master"
+
+  if grep -q initial-branch <(git init -h)
+  then
+    execute "git" "init" "--quiet"
+  else
+    execute "git" "init" "--quiet" "--initial-branch" "master"
+  fi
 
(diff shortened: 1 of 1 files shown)


Bo98:
Doing what we do in `brew` is probably the cleanest way to handle this: https://github.com/Homebrew/brew/blob/76b87c4a00ed636bcf59c104b85863637609b66d/Library/Homebrew/dev-cmd/tap-new.rb#L162

MikeMcQuaid:
Thanks for the PR! Just reverted the previous one instead.

ZhongRuoyu:
Sorry for the trouble, @An-GG! I should have considered that when I wrote #747.

An-GG:
> Thanks for the PR! Just reverted the previous one instead.

thanks for unbreaking the installer for 20.04 👍

also jesus i did not expect `brew style --fix` to end up installing linux-headers, style check info should be printed in the CI log tbh

Bo98:
> style check info should be printed in the CI log tbh

It does. The diff it's telling you to do is perhaps not obvious by eye (trailing whitespace).

PR #750, opened 2023-02-23
install.sh: override default branch config for `git init` invocations

ZhongRuoyu:
Same as #747, but with a better approach which should be compatible with older versions of `git`. This makes sure `git init` always creates a `master` branch.

The same is being done at
https://github.com/Homebrew/brew/blob/76b87c4a00ed636bcf59c104b85863637609b66d/Library/Homebrew/dev-cmd/tap-new.rb#L162.

---

On a side note, @MikeMcQuaid mentioned in https://github.com/Homebrew/install/pull/747#issuecomment-1441774532:

> Similarly, may want to add CI for Ubuntu 18.04.

GitHub Actions will [drop](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu1804-Readme.md) their Ubuntu 18.04 runner image this April. And, all runner images provided by GitHub Actions seem to have a very recent version of Git installed. Maybe a Docker image like `buildpack-deps:18.04` could be used? Or, start from scratch with `ubuntu:18.04`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -887,7 +887,7 @@ ohai "Downloading and installing Homebrew..."
   cd "${HOMEBREW_REPOSITORY}" >/dev/null || return
 
   # we do it in four steps to avoid merge errors when reinstalling
-  execute "git" "init" "-q"
+  execute "git" "-c" "init.defaultBranch=master" "init" "--quiet"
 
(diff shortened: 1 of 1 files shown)


MikeMcQuaid (approved):
Thanks again @ZhongRuoyu!

PR #751, opened 2023-02-28
install: use known-good Git

gromgit:
Since we're testing each `git` in `$PATH` for suitability, we might as well use the one we've approved. :)

Addresses https://github.com/orgs/Homebrew/discussions/4275.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -447,7 +447,8 @@ fi
 cd "/usr" || exit 1
 
 ####################################################################### script
-if ! command -v git >/dev/null
+USABLE_GIT="$(command -v git)"
+if [[ -z "${USABLE_GIT}" ]]
 then
   abort "$(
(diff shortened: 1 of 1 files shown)


SMillerDev (approved):

install.sh
```diff
@@ -447,7 +447,8 @@ fi
 cd "/usr" || exit 1
 
 ####################################################################### script
-if ! command -v git >/dev/null
+USABLE_GIT="$(command -v git >/dev/null)"
```
XuehaiPan, line 450:
```suggestion
USABLE_GIT="$(command -v git)"
```

gromgit: :man_facepalming:

MikeMcQuaid (approved):
Makes sense, thanks again @gromgit!

PR #756, opened 2023-03-06
Fix `./bin` being in `$PATH`

ChillerDragon:
Ensures the installer does not pick a local path to the git executable

Could happen if one has `export PATH='./bin:$PATH` in the shell rc file then it would set USABLE_GIT to be `./bin/git`
which later fails when another cd was executed

So `command -v git` migth spit out `./bin/git`
But `command -p -v git` prints `/usr/bin/git`

Follow up to https://github.com/Homebrew/install/pull/752

Worked fine with found and not found git on debian 11 and on macOS Monterey.
Would be nice if someone could verify it works fine on older and newer macOS before merging this.

```
chiller@debian11:~$ cd /usr/
chiller@debian11:/usr$ export PATH="./bin:$PATH"
chiller@debian11:/usr$ which -a git
./bin/git
/usr/bin/git
chiller@debian11:/usr$ command -v git
./bin/git
chiller@debian11:/usr$ command -V git
git is /usr/bin/git
chiller@debian11:/usr$ command -v git
./bin/git
chiller@debian11:/usr$ command -p -v git
/usr/bin/git
chiller@debian11:/usr$ command -p -v fakegit
chiller@debian11:/usr$ uname -v
#1 SMP Debian 5.10.158-2 (2022-12-13)
```

```
chiller@MacMonterey ~ % cd /usr
chiller@MacMonterey /usr % export PATH="./bin:$PATH"
chiller@MacMonterey /usr % which -a git
./bin/git
./bin/git
/usr/local/bin/git
/usr/bin/git
chiller@MacMonterey /usr % command -v git
./bin/git
chiller@MacMonterey /usr % command -p -v  git
/usr/bin/git
chiller@MacMonterey /usr % command -p -v fakegit
chiller@MacMonterey /usr % uname -v
Darwin Kernel Version 21.4.0: Mon Feb 21 20:34:37 PST 2022; root:xnu-8020.101.4~2/RELEASE_X86_64
```


ChillerDragon:
This is basically the same as

```
export PATH="$(getconf PATH)"
command -v git
```

So in my case that is setting PATH to be only `/bin:/usr/bin` on my debian machine and to `/usr/bin:/bin:/usr/sbin:/sbin` on mac. So depending on in which custom PATHS someone was expected to be found there this is no longer working. I guess that defeats the whole purpose of the whole USABLE_GIT variable so that fix is not desirable I just realized.

gromgit:
I think it'll be more robust to simply ignore (and optionally warn the user) any relative-pathed tools that `which -a` finds in `find_tool()`: https://github.com/Homebrew/install/pull/752#issuecomment-1457315767

ChillerDragon:
> I think it'll be more robust to simply ignore (and optionally warn the user about) any relative-path tools that `which -a` finds in `find_tool()`: [#752 (comment)](https://github.com/Homebrew/install/pull/752#issuecomment-1457315767)

Yea that seems smooth. And then use ``find_tool`` instead of ``command -v``

ChillerDragon:
Thats how it would look like on my system. When I uninstall git it looks like this:

```
$ ./install.sh 
You must install Git before installing Homebrew. See:
  https://docs.brew.sh/Installation
```

with ./bin in PATH it looks like this:

```
Warning: Ignoring ./bin/git - relative paths don't work
Warning: Ignoring ./bin/git - relative paths don't work
Warning: Ignoring ./bin/curl - relative paths don't work
==> This script will install:

```

I hope @SMillerDev enjoys that we educate every disaster setup user while keeping their install working :) 

Since we are printing a message in ``find_tool`` which uses stdout to "return" data I had to modify warn to print to stderr instead. I would assume that catching the result of a subshell does not contain the stderr output but since [this comment](https://github.com/Homebrew/install/pull/752#issuecomment-1452429283) I am not so sure about its portability anymore. Could someone verify that this code snippet keeps the variable ``x`` empty on all systems:

```
chiller@debian:~$ x="$(echo "foo" >&2)"
foo
chiller@debian:~$ echo "$x"

chiller@debian:~$ 
```

install.sh
```diff
@@ -405,7 +405,10 @@ find_tool() {
   local executable
   while read -r executable
   do
-    if "test_$1" "${executable}"
+    if [[ "${executable}" != /* ]]
+    then
+      warn "Ignoring ${executable} - relative paths don't work"
```
MikeMcQuaid, line 410:
```suggestion
      warn "Ignoring ${executable} (relative paths don't work)"
```

MikeMcQuaid (approved):

install.sh
```diff
@@ -447,7 +450,7 @@ fi
 cd "/usr" || exit 1
 
 ####################################################################### script
-USABLE_GIT="$(command -v git)"
+USABLE_GIT="$(find_tool git)"
```
MikeMcQuaid, line 453: This doesn't work on macOS

MikeMcQuaid (commented):
Thanks for trying again @ChillerDragon, looking good so far!

install.sh
ChillerDragon: Yes I am on it its a bit tricky tho. The troll operating system macOS puts a git in PATH that is not installed :D
So when running ``git --version`` it prints this "version" ``xcode-select: note: no developer tools were found at '/Applications/Xcode.app', requesting install. Choose an option in the dialog to download the command line developer tools.``

Which does not pass the ``test_git`` version check. The CI was working because later in the install script it does run the xcode install. So I have to think about how to handle this.

ChillerDragon: I haven't even red all lines of the install script. How messy would it be to move the [xcode install lines](https://github.com/Homebrew/install/blob/16aefe107e14322a2104c65bff42de5f6ac6297c/install.sh#L840-L862) before the set of USABLE_GIT?

ChillerDragon:
I am still working on this. Current state is trying to debug/fix git not being installed on macOS yet. Will probably take me a few more days to find enough time to finish it.

ChillerDragon:
Okay figured something out. The current version check can not parse the git version that the macOS CI installs.
Running ``git --version`` on my system returns ``git version 2.39.2`` so this works fine:

```bash
version_gt() {
  [[ "${1%.*}" -gt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -gt "${2#*.}" ]]
}
version_ge() {
  [[ "${1%.*}" -gt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -ge "${2#*.}" ]]
}
version_lt() {
  [[ "${1%.*}" -lt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -lt "${2#*.}" ]]
}
major_minor() {
  echo "${1%%.*}.$(
    x="${1#*.}"
    echo "${x%%.*}"
  )"
}

git_version_output="git version 2.39.2"
major_minor "${git_version_output##* }"
```

prints ``2.39``

But the git in the macOS CI prints ``git version 2.37.1 (Apple Git-137.1)``

```bash
version_gt() {
  [[ "${1%.*}" -gt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -gt "${2#*.}" ]]
}
version_ge() {
  [[ "${1%.*}" -gt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -ge "${2#*.}" ]]
}
version_lt() {
  [[ "${1%.*}" -lt "${2%.*}" ]] || [[ "${1%.*}" -eq "${2%.*}" && "${1#*.}" -lt "${2#*.}" ]]
}
major_minor() {
  echo "${1%%.*}.$(
    x="${1#*.}"
    echo "${x%%.*}"
  )"
}

git_version_output="git version 2.37.1 (Apple Git-137.1)"
major_minor "${git_version_output##* }"
```

prints ``Git-137.1)`` and when this is passed to the version check it throws an error:

```bash
version_ge "$(major_minor "${git_version_output##* }")" "$(major_minor "${REQUIRED_GIT_VERSION}")" # => line 19: Git: unbound variable
```

The least intrusive fix I cam up with is making sure git version can only contain three words using awk.
```bash
git_version_output="git version 2.37.1 (Apple Git-137.1)"
echo "$git_version_output" | awk '{ print $1 " " $2 " " $3}'

# => git version 2.37.1
```

Honestly I do not fully understand or even bothered to understand the version check but it seems a bit flawed. Matching the semver version with a regex and then passing it into ``sort -V`` seems more stable and correct but thats for another day.

ChillerDragon:
Okay should be working now. I am a bit scared something broke since I touched so much code. But here is a quick summary of that the current change does.

- use ``find_tool`` instead of ``command -v``
- make ``find_tool`` drop local paths
- change ``warn`` to print to stderr so ``warn`` can be called in ``find_tool`` without getting catched by the subshell
- ensure the git version that is checked is at most 3 words so the version checker does not break
- check curl and git version after installing xcode instead of before because the version is not know before

MikeMcQuaid:
@gromgit @vitorgalvao and any other @Homebrew/members and @Homebrew/maintainers: if you can test this out and review it I'd be really grateful! This stuff is really hard to do robustly.

fxcoudert:
There are people who use relative paths in `PATH`? But why?!

install.sh
```diff
@@ -383,7 +383,7 @@ test_git() {
   fi
 
   local git_version_output
-  git_version_output="$("$1" --version 2>/dev/null)"
+  git_version_output="$("$1" --version 2>/dev/null | awk '{ print $1 " " $2 " " $3 }')"
   version_ge "$(major_minor "${git_version_output##* }")" "$(major_minor "${REQUIRED_GIT_VERSION}")"
```
gromgit, line 387:
```suggestion
  git_version_output="$("$1" --version 2>/dev/null)"
  if [[ "$git_version_output" =~ "git version "([^ ]*).* ]]; then
    version_ge "$(major_minor "${BASH_REMATCH[1]}")" "$(major_minor "${REQUIRED_GIT_VERSION}")"
  else
    abort "Unexpected Git version string '${git_version_output}' reported"
  fi
```

gromgit (changes requested):
Let's not rely on external utilities when `bash` itself has the necessary tools. Regex pattern matching is an old `bash` feature, definitely available before macOS's ancient v3.2 version.

gromgit:
> There are people who use relative paths in `PATH`? But why?!

Bad tutorials and/or advice. :man_shrugging:

install.sh
ChillerDragon: > Matching the semver version with a regex and then passing it into sort -V seems more stable and correct but thats for another day.

As I said I would like to delay the version check refactor for a new pr and keep this about making sure the installer works on all systems again. But if you have hard feelings on this I can click the commit suggestion button.

ChillerDragon:
> There are people who use relative paths in `PATH`? But why?!

Convenience of not having to type ``./bin/application`` when using projects/frameworks that provide this folder to set a application in a specific version. Its a thing trust me. It has not been a big problem so far because usually people install brew first and then their custom settings. And also since the brew installer only started breaking just recently when the new USABLE_GIT was used instead of calling git directly.

install.sh
gromgit: The change I'm requesting actually _matches the version string prefix_, rather than just assuming it's "git version". This is additional assurance that the installer is reading the right information.

ChillerDragon: Yes I understand the code ._. as I said I planned to do the same in a new pr. Since I want to keep this hot fix as simple as possible to ensure that the known bug is fixed as fast as possible. But since you seem to be insisting on this change I made sure its working as expected and tested it locally and committed it :slightly_smiling_face:

install.sh
```diff
@@ -384,7 +384,12 @@ test_git() {
 
   local git_version_output
   git_version_output="$("$1" --version 2>/dev/null)"
-  version_ge "$(major_minor "${git_version_output##* }")" "$(major_minor "${REQUIRED_GIT_VERSION}")"
+  if [[ "${git_version_output}" =~ "git version "([^ ]*).* ]]
+  then
+    version_ge "$(major_minor "${BASH_REMATCH[1]}")" "$(major_minor "${REQUIRED_GIT_VERSION}")"
+  else
+    abort "Unexpected Git version string '${git_version_output}' reported"
```
MikeMcQuaid, line 391:
```suggestion
    abort "Unexpected Git version: '${git_version_output}'!"
```

MikeMcQuaid (approved):

PR #757, opened 2023-03-24
Use xcode-select for CLDT detection

jamestut:
CLDT is not required when Xcode is installed, as Xcode is a superset of CLDT. I think having both installed is just a waste of disk space.

This PR updates the installer to use the macOS' built-in `xcode-select` command rather than checking for a hardcoded path. `xcode-select -p` will return an error code if there is no CLDT or Xcode installed.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -317,7 +317,7 @@ should_install_command_line_tools() {
 
   if version_gt "${macos_version}" "10.13"
   then
-    ! [[ -e "/Library/Developer/CommandLineTools/usr/bin/git" ]]
+    ! xcode-select -p &> /dev/null
   else
     ! [[ -e "/Library/Developer/CommandLineTools/usr/bin/git" ]] ||
       ! [[ -e "/usr/include/iconv.h" ]]


SMillerDev:
As far as I know it is not a superset. Do you have a list of all tools available in both so we can compare?

jamestut:
I'll come back to it later as I am outside now. In the meantime, can you
provide a reference if what I mentioned is not the case? E.g. as far as you
know, what tools are available in CLDT and not in Xcode?

During my experience in using brew (I modified it to avoid installing
CLDT), all things are working fine on my M1 machine, including installing
packages that normally requires CLDT. I never used an Intel mac tho so
can't tell if that would be the case pre Big Sur.

carlocab (commented):
Checking for the CLT specifically (rather than just any Xcode install) is intentional here. If you'd rather not install the CLT with Homebrew use an [alternative installation method](https://docs.brew.sh/Installation#alternative-installs).

Note that alternative installation methods are unsupported, and you will not be able to use bottles for some formulae (e.g. `python@3.*`) if you do not have a CLT installation at `/Library/Developer/CommandLineTools`. You are also likely to experience problems with building formulae from source unless you are on the latest version of macOS.

carlocab (changes requested):
I meant to request changes, not comment.

MikeMcQuaid:
> CLDT is not required when Xcode is installed,

This is not true. Closing.

jamestut:
Hi @MikeMcQuaid, I'd curious what formulae would specifically require CLDT and won't work with only Xcode installed?

I've installed Pillow, Numpy, and other Python 3 libraries, and all of them works fine without CLDT. I encountered some formulas (I think it was `llvm`) which explicitly asked for CLDT, but I modified the formula to take the path from `xcode-select -p`, and it installed and worked fine. In fact, I was on Xcode-beta at that time which again homebrew did not properly detect due to hardcoding the path to `/Applications/Xcode.app`.

I did `grep` in the `homebrew-core:/Formula` and indeed it was only `llvm` and `git` formula that have the string `/Library/Developer/CommandLineTools` hardcoded to it.

I apologize beforehand if I might sound pushy, but I really curious why you all insist to have CLDT installed, and would like to know the real reason. 🙏

MikeMcQuaid:
@jamestut @carlocab has given some examples above. To find more: go digging in homebrew/core.

> I really curious why you all insist to have CLDT installed

We don't. The official installer will install it. You can choose to not use this installer.

> and would like to know the real reason.

We told you the real reason.

Another reason: we've been running a package manager for millions of people for over a decade. We would rather have a default behaviour that definitely works for most of those people, even if it installs things unnecessary for some people, than change it because someone like yourself consider it redundant which then breaks things for many more users.

PR #760, opened 2023-04-01
install: support OCI containers as root

b8rky:
similar to the check for `/.dockerenv`, this adds a check for `/run/.containerenv`. if it exists, bypass the root user check. 

without this, you cannot execute this script as root in a container being managed by something like podman. it will always abort on line 310. 

see the [podman-run description](https://docs.podman.io/en/latest/markdown/podman-run.1.html#description) for more information about this file and my reasoning for including it.


dawidd6 (approved):

carlocab (approved):

MikeMcQuaid:
Thanks so much for your first contribution (hopefully of many)! Without people like you submitting PRs we couldn't run this project. You rock, @b8rky!

PR #774, opened 2023-05-17
Fix git outdated error saying git is not installed

ChillerDragon:
Fixes #773

In https://github.com/Homebrew/install/pull/756 I introduced the bug of changing the error message "git is too old" into "git is not installed"

I tested this pr by uninstalling git and then running

```
echo -e '#!/bin/bash\necho git version 1.8.3.1' | sudo tee /usr/local/bin/git
sudo chmod +x /usr/local/bin/git
```

Before the pr it said

```
You must install Git before installing Homebrew. See:
  https://docs.brew.sh/Installation
```

Now it says

```
The version of Git that was found does not satisfy requirements for Homebrew.
Please install Git 2.7.0 or newer and add it to your PATH.
```

Issue #773, opened 2023-05-17
Install script CentOS 7 can't find git


osalbahr:
- [x] your problem was from running the official `install` or `uninstall` script?
- [x] after installation: ran `brew config` and `brew doctor` and included their output with your issue? If you couldn't install: provided your OS version with the output of your issue?
```
[osalbahr@vclvm177-222 ~]$ cat /etc/redhat-release
CentOS Linux release 7.9.2009 (Core)
[osalbahr@vclvm177-222 ~]$ uname -a
Linux vclvm177-222.vcl.ncsu.edu 3.10.0-1160.81.1.el7.x86_64 #1 SMP Fri Dec 16 17:29:43 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
[osalbahr@vclvm177-222 ~]$ 
```

## What you were trying to do (and why)

I was trying to install Homebrew so I can `brew install neofetch`.

## What happened (include command output)

<details>
  <summary>You must install Git before installing Homebrew</summary>
  <pre>
[osalbahr@vclvm177-222 ~]$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
==> Checking for `sudo` access (which may request your password)...
==> This script will install:
/home/linuxbrew/.linuxbrew/bin/brew
/home/linuxbrew/.linuxbrew/share/doc/homebrew
/home/linuxbrew/.linuxbrew/share/man/man1/brew.1
/home/linuxbrew/.linuxbrew/share/zsh/site-functions/_brew
/home/linuxbrew/.linuxbrew/etc/bash_completion.d/brew
/home/linuxbrew/.linuxbrew/Homebrew

Press RETURN/ENTER to continue or any other key to abort:
==> /usr/bin/sudo /bin/chown -R osalbahr:NCSU /home/linuxbrew/.linuxbrew/Homebrew
You must install Git before installing Homebrew. See:
  https://docs.brew.sh/Installation
[osalbahr@vclvm177-222 ~]$ 
  </pre>
</details>

## What you expected to happen

<details>
  <summary>I expected the install script to go through, since I do have git installed:</summary>
  <pre>
[osalbahr@vclvm177-222 ~]$ git --version
git version 1.8.3.1
[osalbahr@vclvm177-222 ~]$ which -a git
/usr/bin/git
[osalbahr@vclvm177-222 ~]$ echo $PATH
[...]

PR #775, opened 2023-05-19
Fix `./bin` being in `$PATH` (mac)

ChillerDragon:
https://github.com/Homebrew/install/pull/774 broke https://github.com/Homebrew/install/pull/756 for mac.

Thanks to @Bo98 for [reporting](https://github.com/Homebrew/install/pull/774#issuecomment-1555142496)

So my fix for the fix in the fix that fixed a bug in the previous fix had a bug. This probably introduces a new bug. Hard to tell without proper test coverage.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -833,28 +833,29 @@ EOABORT
   )"
 fi
 
-USABLE_GIT="$(command -v git)"
-if [[ -z "${USABLE_GIT}" ]]
+USABLE_GIT=/usr/bin/git
+if [[ -n "${HOMEBREW_ON_LINUX-}" ]]
 then
-  abort "$(
-    cat <<EOABORT
-You must install Git before installing Homebrew. See:
-  ${tty_underline}https://docs.brew.sh/Installation${tty_reset}
-EOABORT
-  )"
-elif [[ -n "${HOMEBREW_ON_LINUX-}" ]]
-then
-  suitable_git="$(find_tool git)"
-  if [[ -z "${suitable_git}" ]]
+  USABLE_GIT="$(find_tool git)"
+  git_installed="$(command -v git)"
+  if [[ -z "${git_installed}" ]]
   then
     abort "$(
       cat <<EOABORT
-The version of Git that was found does not satisfy requirements for Homebrew.
-Please install Git ${REQUIRED_GIT_VERSION} or newer and add it to your PATH.
+  You must install Git before installing Homebrew. See:
+    ${tty_underline}https://docs.brew.sh/Installation${tty_reset}
 EOABORT
     )"
   fi
-  USABLE_GIT="${suitable_git}"
+  if [[ -z "${USABLE_GIT}" ]]
+  then
+    abort "$(
+      cat <<EOABORT
+  The version of Git that was found does not satisfy requirements for Homebrew.
+  Please install Git ${REQUIRED_GIT_VERSION} or newer and add it to your PATH.
+EOABORT
+    )"
+  fi
   if [[ "${USABLE_GIT}" != /usr/bin/git ]]
   then
     export HOMEBREW_GIT_PATH="${USABLE_GIT}"
--- a/install.sh
+++ b/install.sh
@@ -837,8 +837,7 @@ USABLE_GIT=/usr/bin/git
 if [[ -n "${HOMEBREW_ON_LINUX-}" ]]
 then
   USABLE_GIT="$(find_tool git)"
-  git_installed="$(command -v git)"
-  if [[ -z "${git_installed}" ]]
+  if [[ -z "$(command -v git)" ]]
   then
     abort "$(
       cat <<EOABORT


osalbahr:
FYI, this PR while doesn't re-introduce #773, it behaves a little differently. It fails later in the install. I think this is ok.

```
[osalbahr@vclvm176-217 ~]$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/ChillerDragon/install/pr_fix_fix_fix_path/install.sh)"
==> This script will install:
==> The following new directories will be created:
/home/linuxbrew/.linuxbrew/bin
/home/linuxbrew/.linuxbrew/etc
/home/linuxbrew/.linuxbrew/include
/home/linuxbrew/.linuxbrew/lib
/home/linuxbrew/.linuxbrew/sbin
/home/linuxbrew/.linuxbrew/share
/home/linuxbrew/.linuxbrew/var
/home/linuxbrew/.linuxbrew/opt
/home/linuxbrew/.linuxbrew/share/zsh
/home/linuxbrew/.linuxbrew/share/zsh/site-functions
/home/linuxbrew/.linuxbrew/var/homebrew
/home/linuxbrew/.linuxbrew/var/homebrew/linked
/home/linuxbrew/.linuxbrew/Cellar
/home/linuxbrew/.linuxbrew/Caskroom
/home/linuxbrew/.linuxbrew/Frameworks

==> /usr/bin/sudo /usr/bin/install -d -o osalbahr -g NCSU -m 0755 /home/linuxbrew/.linuxbrew
==> /usr/bin/sudo /bin/mkdir -p /home/linuxbrew/.linuxbrew/bin /home/linuxbrew/.linuxbrew/etc /home/linuxbrew/.linuxbrew/include /home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/sbin /home/linuxbrew/.linuxbrew/share /home/linuxbrew/.linuxbrew/var /home/linuxbrew/.linuxbrew/opt /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions /home/linuxbrew/.linuxbrew/var/homebrew /home/linuxbrew/.linuxbrew/var/homebrew/linked /home/linuxbrew/.linuxbrew/Cellar /home/linuxbrew/.linuxbrew/Caskroom /home/linuxbrew/.linuxbrew/Frameworks
==> /usr/bin/sudo /bin/chmod ug=rwx /home/linuxbrew/.linuxbrew/bin /home/linuxbrew/.linuxbrew/etc /home/linuxbrew/.linuxbrew/include /home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/sbin /home/linuxbrew/.linuxbrew/share /home/linuxbrew/.linuxbrew/var /home/linuxbrew/.linuxbrew/opt /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions /home/linuxbrew/.linuxbrew/var/homebrew /home/linuxbrew/.linuxbrew/var/homebrew/linked /home/linuxbrew/.linuxbrew/Cellar /home/linuxbrew/.linuxbrew/Caskroom /home/linuxbrew/.linuxbrew/Frameworks
==> /usr/bin/sudo /bin/chmod go-w /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions
==> /usr/bin/sudo /bin/chown osalbahr /home/linuxbrew/.linuxbrew/bin /home/linuxbrew/.linuxbrew/etc /home/linuxbrew/.linuxbrew/include /home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/sbin /home/linuxbrew/.linuxbrew/share /home/linuxbrew/.linuxbrew/var /home/linuxbrew/.linuxbrew/opt /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions /home/linuxbrew/.linuxbrew/var/homebrew /home/linuxbrew/.linuxbrew/var/homebrew/linked /home/linuxbrew/.linuxbrew/Cellar /home/linuxbrew/.linuxbrew/Caskroom /home/linuxbrew/.linuxbrew/Frameworks
==> /usr/bin/sudo /bin/chgrp NCSU /home/linuxbrew/.linuxbrew/bin /home/linuxbrew/.linuxbrew/etc /home/linuxbrew/.linuxbrew/include /home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/sbin /home/linuxbrew/.linuxbrew/share /home/linuxbrew/.linuxbrew/var /home/linuxbrew/.linuxbrew/opt /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions /home/linuxbrew/.linuxbrew/var/homebrew /home/linuxbrew/.linuxbrew/var/homebrew/linked /home/linuxbrew/.linuxbrew/Cellar /home/linuxbrew/.linuxbrew/Caskroom /home/linuxbrew/.linuxbrew/Frameworks
==> /usr/bin/sudo /bin/mkdir -p /home/linuxbrew/.linuxbrew/Homebrew
==> /usr/bin/sudo /bin/chown -R osalbahr:NCSU /home/linuxbrew/.linuxbrew/Homebrew
The version of Git that was found does not satisfy requirements for Homebrew.
Please install Git 2.7.0 or newer and add it to your PATH.
[osalbahr@vclvm176-217 ~]$ git --version
git version 1.8.3.1
[osalbahr@vclvm176-217 ~]$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
==> This script will install:

==> /usr/bin/sudo /bin/chown -R osalbahr:NCSU /home/linuxbrew/.linuxbrew/Homebrew
The version of Git that was found does not satisfy requirements for Homebrew.
Please install Git 2.7.0 or newer and add it to your PATH.
[osalbahr@vclvm176-217 ~]$ 
```

ChillerDragon:
> FYI, this PR while doesn't re-introduce #773, it behaves a little differently. It fails later in the install. I think this is ok.
> 
> 
> 
> ```
> 
> [osalbahr@vclvm176-217 ~]$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/ChillerDragon/install/pr_fix_fix_fix_path/install.sh)"
> 
> ==> Checking for `sudo` access (which may request your password)...
> 
> ==> This script will install:
> 
> /home/linuxbrew/.linuxbrew/bin/brew
> 
> /home/linuxbrew/.linuxbrew/share/doc/homebrew
> 
> /home/linuxbrew/.linuxbrew/share/man/man1/brew.1
> 
> /home/linuxbrew/.linuxbrew/share/zsh/site-functions/_brew
> 
> /home/linuxbrew/.linuxbrew/etc/bash_completion.d/brew
> 
> /home/linuxbrew/.linuxbrew/Homebrew
> 
> ==> The following new directories will be created:
> 
> /home/linuxbrew/.linuxbrew/bin
> 
> /home/linuxbrew/.linuxbrew/etc
> 
> /home/linuxbrew/.linuxbrew/include
> 
> /home/linuxbrew/.linuxbrew/lib
> 
> /home/linuxbrew/.linuxbrew/sbin
> 
> /home/linuxbrew/.linuxbrew/share
> 
> /home/linuxbrew/.linuxbrew/var
> 
> /home/linuxbrew/.linuxbrew/opt
> 
> /home/linuxbrew/.linuxbrew/share/zsh
> 
> /home/linuxbrew/.linuxbrew/share/zsh/site-functions
> 
> /home/linuxbrew/.linuxbrew/var/homebrew
> 
> /home/linuxbrew/.linuxbrew/var/homebrew/linked
> 
> /home/linuxbrew/.linuxbrew/Cellar
> 
> /home/linuxbrew/.linuxbrew/Caskroom
> 
> /home/linuxbrew/.linuxbrew/Frameworks
> 
> 
> 
> Press RETURN/ENTER to continue or any other key to abort:
> 
> ==> /usr/bin/sudo /usr/bin/install -d -o osalbahr -g NCSU -m 0755 /home/linuxbrew/.linuxbrew
> 
> ==> /usr/bin/sudo /bin/mkdir -p /home/linuxbrew/.linuxbrew/bin /home/linuxbrew/.linuxbrew/etc /home/linuxbrew/.linuxbrew/include /home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/sbin /home/linuxbrew/.linuxbrew/share /home/linuxbrew/.linuxbrew/var /home/linuxbrew/.linuxbrew/opt /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions /home/linuxbrew/.linuxbrew/var/homebrew /home/linuxbrew/.linuxbrew/var/homebrew/linked /home/linuxbrew/.linuxbrew/Cellar /home/linuxbrew/.linuxbrew/Caskroom /home/linuxbrew/.linuxbrew/Frameworks
> 
> ==> /usr/bin/sudo /bin/chmod ug=rwx /home/linuxbrew/.linuxbrew/bin /home/linuxbrew/.linuxbrew/etc /home/linuxbrew/.linuxbrew/include /home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/sbin /home/linuxbrew/.linuxbrew/share /home/linuxbrew/.linuxbrew/var /home/linuxbrew/.linuxbrew/opt /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions /home/linuxbrew/.linuxbrew/var/homebrew /home/linuxbrew/.linuxbrew/var/homebrew/linked /home/linuxbrew/.linuxbrew/Cellar /home/linuxbrew/.linuxbrew/Caskroom /home/linuxbrew/.linuxbrew/Frameworks
> 
> ==> /usr/bin/sudo /bin/chmod go-w /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions
> 
> ==> /usr/bin/sudo /bin/chown osalbahr /home/linuxbrew/.linuxbrew/bin /home/linuxbrew/.linuxbrew/etc /home/linuxbrew/.linuxbrew/include /home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/sbin /home/linuxbrew/.linuxbrew/share /home/linuxbrew/.linuxbrew/var /home/linuxbrew/.linuxbrew/opt /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions /home/linuxbrew/.linuxbrew/var/homebrew /home/linuxbrew/.linuxbrew/var/homebrew/linked /home/linuxbrew/.linuxbrew/Cellar /home/linuxbrew/.linuxbrew/Caskroom /home/linuxbrew/.linuxbrew/Frameworks
> 
> ==> /usr/bin/sudo /bin/chgrp NCSU /home/linuxbrew/.linuxbrew/bin /home/linuxbrew/.linuxbrew/etc /home/linuxbrew/.linuxbrew/include /home/linuxbrew/.linuxbrew/lib /home/linuxbrew/.linuxbrew/sbin /home/linuxbrew/.linuxbrew/share /home/linuxbrew/.linuxbrew/var /home/linuxbrew/.linuxbrew/opt /home/linuxbrew/.linuxbrew/share/zsh /home/linuxbrew/.linuxbrew/share/zsh/site-functions /home/linuxbrew/.linuxbrew/var/homebrew /home/linuxbrew/.linuxbrew/var/homebrew/linked /home/linuxbrew/.linuxbrew/Cellar /home/linuxbrew/.linuxbrew/Caskroom /home/linuxbrew/.linuxbrew/Frameworks
> 
> ==> /usr/bin/sudo /bin/mkdir -p /home/linuxbrew/.linuxbrew/Homebrew
> 
> ==> /usr/bin/sudo /bin/chown -R osalbahr:NCSU /home/linuxbrew/.linuxbrew/Homebrew
> 
> The version of Git that was found does not satisfy requirements for Homebrew.
> 
> Please install Git 2.7.0 or newer and add it to your PATH.
> 
> [osalbahr@vclvm176-217 ~]$ git --version
> 
> git version 1.8.3.1
> 
> [osalbahr@vclvm176-217 ~]$ /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
> 
> ==> Checking for `sudo` access (which may request your password)...
> 
> ==> This script will install:
> 
> /home/linuxbrew/.linuxbrew/bin/brew
> 
> /home/linuxbrew/.linuxbrew/share/doc/homebrew
> 
> /home/linuxbrew/.linuxbrew/share/man/man1/brew.1
> 
> /home/linuxbrew/.linuxbrew/share/zsh/site-functions/_brew
> 
> /home/linuxbrew/.linuxbrew/etc/bash_completion.d/brew
> 
> /home/linuxbrew/.linuxbrew/Homebrew
> 
> 
> 
> Press RETURN/ENTER to continue or any other key to abort:
> 
> ==> /usr/bin/sudo /bin/chown -R osalbahr:NCSU /home/linuxbrew/.linuxbrew/Homebrew
> 
> The version of Git that was found does not satisfy requirements for Homebrew.
> 
> Please install Git 2.7.0 or newer and add it to your PATH.
> 
> [osalbahr@vclvm176-217 ~]$ 
> 
> ```
> 
> 
> 
> 

As far as I understood it that was never a bug but a feature. Do you have a git installed that is 2.7.0 or later? If not then the installer can not proceed.

osalbahr:
It might be needed to further modify `install.sh` so that the correct version of `git` is checked before any directories are created, just like it used to before #774. Idk

MikeMcQuaid:
> Hard to tell without proper test coverage.

Could some be added?

Bo98 (dismissed):
I'm fine with merging this as is, as it fixes a bug. However there's a couple problems I have with this general code (which is not really because of this PR in particular):

* I don't see the value of running it at all on macOS. We install the CLT as a part of the install process, and we can guarantee a git binary at a fixed location (`/usr/bin/git`). We currently support all Git versions provided by Apple (even on El Capitan, where Xcode provides Git 2.7.4 - not that the install script is allowed on those macOS versions anyway). We do not support abritrary Git executables in the PATH on macOS in brew itself, so I don't see why we need or want to use them here.
* This check really should be placed _before_ we do any part of the actual install.

Bo98:
Placing all this in a `HOMEBREW_ON_LINUX` check and moving it as early as possible seems sufficient.

MikeMcQuaid:
> Would it be better to handle macOS as a special case then?

Yes.

> Placing all this in a `HOMEBREW_ON_LINUX` check and moving it as early as possible seems sufficient.

Agreed 👍🏻

osalbahr:
I see. How about assuming that `git` exists at the usual location with an appropriate version and *then* fall back to `find_tool` if needed in case of an unconventional setup? And maybe issuing a warning if `/usr/bin/git` doesn’t exist or outdated

osalbahr:
Does `brew` itself even do a similar thing or just directly call git? I couldn’t find a similar function other than in the installer.

Bo98:
In `brew`, we do not check the Git version on macOS. All our checks on macOS are based on the macOS version itself.

The `git` installed by Homebrew itself can be used, but that's not applicable when Homebrew isn't actually installed yet. Beyond that we technically do this:

- Check `/usr/bin/xcrun -find git` and use that at the Git path.
- If `xcode-select -print-path` is misconfigured, it will check Xcode.app (`/Applications/Xcode.app/Contents/Developer/usr/bin/git` - though IMO this should check the CLT instead).

However we don't need to mimic precisely this in the install script. The CLT check in the install script checks for the presence of `/Library/Developer/CommandLineTools/usr/bin/git` and installs the CLT if it's not there. Given we guarantee the CLT is installed, we can guarantee `/Library/Developer/CommandLineTools/usr/bin/git` is available.

`/usr/bin/git` is a stub that points to a git configured by `xcode-select -print-path` - so it's really the same thing as `xcrun git`. I'm however happy to use `/Library/Developer/CommandLineTools/usr/bin/git` directly instead, though a misconfigured `xcode-select -print-path` will may cause issues when running `brew` anyway or even in the Xcode license check that does `xcrun clang`.

osalbahr:
Yeah it seems that shortcutting the checks as a special case for macOS is reasonable and consistent with `brew`. Not sure how `brew` determines which `git` to use on Linux though

Bo98:
First on the PATH basically, though this path is hardwired to be `/usr/bin:/bin:/usr/sbin:/sbin` so doesn't really consider user's configured PATHs. However `HOMEBREW_GIT_PATH` can be set to use a custom git, which the install script also uses.

ChillerDragon:
I did force push incorporating your feedback.

- :heavy_check_mark: handle macOS as [special case](https://github.com/Homebrew/install/pull/775#issuecomment-1557545148) 

  ``find_tool`` is now only used on linux

- :x: This check really should be placed before we do any part of the actual install.

  Yes I do agree. But do not have the motivation yet to fix it in this pr. And would love to delay it into a new pr.
  Or even let someone else do it if you do not want to wait for me.
  Iirc I did move the whole git version check into a later part of the code to ensure the installer did already install git.
  Because if git is not installed on macOS yet the version check fails. This happend in #756.
  This pr is fixing an actual installation bug and should be merged asap. The install order is more of an nice to have and
  requires more code restructure and testing.
  
I did only test this code on linux. Covering all of the edge cases I know: No git installed, too old git installed, ./bin in PATH

Something noteworthy is that I am now hardcoding /usr/bin/git instead of searching git with ``command -v git`` that is new and has never been done so far as far as I know. I did this to avoid using ``find_tool`` while also avoiding the ./bin PATH bug. Since we assume the git on mac is there anyways that seems sane. But this possibly brings some new issues and next week we will have some angry user feedback about the one mac user that managed to delete /usr/bin/git and uses his selfcompiled git in /usr/local/bin/git

ChillerDragon:
> > Hard to tell without proper test coverage.
> 
> Could some be added?

Yes technically all the edge cases on linux and macOS with old git and PATH shenanigans should be coverable in the CI. But someone has to do it :stuck_out_tongue_winking_eye:

osalbahr:
So what’s the consensus? Should the CI be written under the assumption that user’s custom PATH will be ignored at install time as consistent with `brew`?

Because if so then that simplifies testing a lot and no need for handling `./bin` in PATH

osalbahr:
I might write some tests soon-ish (sometime this week) but I don’t know yet what the end goal is.

Also, how can I `grep` for something in the installer’s output? I know that `| tee out.txt` breaks the install, but there’s probably global variables to disable this behavior that I don’t know of. This is so I can see that the check happened early enough. Probably `wc -l` would suffice too. Not sure which testing approach is more versatile against potential changes in the future

ChillerDragon:
> So what’s the consensus? Should the CI be written under the assumption that user’s custom PATH will be ignored at install time as consistent with `brew`?
> 
> 
> 
> Because if so then that simplifies testing a lot and no need for handling `./bin` in PATH

The CI should ensure that the install fails with a nice error if there is no recent enough git installed.

And the CI should ensure that the installer does not fail even if ./bin is in the PATH

osalbahr:
Should the installer ignore the user’s `PATH`?

ChillerDragon:
> Should the installer ignore the user’s `PATH`? 

I thought about that in the very beginning when fixing the ./bin issue. But then the find_tool seems a bit pointless. I did not implement the whole git version check and did not read up on its reasoning but I can imagine it is to support some exotic Linux user setups that might only install all their programs per user in some ~/.local/bin folder.

But now that I know that brew it self won't work with that because it only looks in the default PATH that seems to make less sense.

MikeMcQuaid:
> * Yes I do agree. But do not have the motivation yet to fix it in this pr.

That's fine with me 👍🏻

install.sh
```diff
@@ -833,28 +833,29 @@ EOABORT
...
+USABLE_GIT=/usr/bin/git
+if [[ -n "${HOMEBREW_ON_LINUX-}" ]]
 then
-  abort "$(
-    cat <<EOABORT
-You must install Git before installing Homebrew. See:
-  ${tty_underline}https://docs.brew.sh/Installation${tty_reset}
+  USABLE_GIT="$(find_tool git)"
+  git_installed="$(command -v git)"
+  if [[ -z "${git_installed}" ]]
```
MikeMcQuaid, line 841:
```suggestion
  if [[ -z "$(command -v git)" ]]
```

MikeMcQuaid (approved):

MikeMcQuaid:
Thanks so much for your repeated contribution! Without people like you submitting PRs we couldn't run this project. You rock, @ChillerDragon!

install.sh
```diff
@@ -833,28 +833,28 @@ EOABORT
...
-if [[ -z "${USABLE_GIT}" ]]
+USABLE_GIT=/usr/bin/git
+if [[ -n "${HOMEBREW_ON_LINUX-}" ]]
 then
-  abort "$(
-    cat <<EOABORT
-You must install Git before installing Homebrew. See:
-  ${tty_underline}https://docs.brew.sh/Installation${tty_reset}
+  USABLE_GIT="$(find_tool git)"
+  if [[ -z "$(command -v git)" ]]
```
maxim-belkin, line 840: What is the point of using `find_tool` if you `abort` on the next line depending on the output of `command -v`?

ChillerDragon: Yea the code is a bit of a mess sorry. I shuffled it around until it worked :sweat_smile: 

It is used later here

https://github.com/Homebrew/install/blob/716a1d024f32890ef75ea82c18a769abc24e9475/install.sh#L849

The code would be more readable if this line was moved to 848.

PR #794, opened 2023-07-27
doc: add `install.sh --help`. Fix #790

### LINKED ISSUE #790: doc: add `install.sh --help`

```console
$ brew doctor
Your system is ready to brew.
$ brew config
HOMEBREW_VERSION: 4.1.0
ORIGIN: https://github.com/Homebrew/brew
HEAD: ac93842f8d13913f462aedb246cfacd08d6b7f05
Last commit: 4 days ago
Core tap JSON: 23 Jul 20:48 UTC
HOMEBREW_PREFIX: /home/linuxbrew/.linuxbrew
HOMEBREW_CASK_OPTS: []
HOMEBREW_MAKE_JOBS: 16
Homebrew Ruby: 2.6.10 => /home/linuxbrew/.linuxbrew/Homebrew/Library/Homebrew/vendor/portable-ruby/2.6.10_1/bin/ruby
CPU: 16-core 64-bit haswell
Clang: N/A
Git: 2.34.1 => /bin/git
Curl: 7.81.0 => /bin/curl
Kernel: Linux 5.15.0-76-generic x86_64 GNU/Linux
OS: Ubuntu 22.04.2 LTS (jammy)
Host glibc: 2.35
/usr/bin/gcc: 11.3.0
/usr/bin/ruby: 3.0.2
glibc: N/A
gcc@11: N/A
gcc: N/A
xorg: N/A
```

I was trying to see a descriptions of the `install.sh` options. For example, I wasn't sure if there was a difference between setting `NONINTERACTIVE` or `CI`.

Related:
- https://github.com/Homebrew/install/issues/369#issuecomment-1628267813

```console
$ ./install.sh --help
==> This script will install:

```

Help message, similar to `./uninstall.sh --help`:
```console
$ ./uninstall.sh --help
Homebrew Uninstaller
Usage: [NONINTERACTIVE=1] ./uninstall.sh [options]
    -p, --path=PATH  Sets Homebrew prefix. Defaults to /home/linuxbrew/.linuxbrew.
        --skip-cache-and-logs
                     Skips removal of HOMEBREW_CACHE and HOMEBREW_LOGS.
    -f, --force      Uninstall without prompting.
    -q, --quiet      Suppress all output.
    -n, --dry-run    Simulate uninstall but don't remove anything.
    -h, --help       Display this message.
    NONINTERACTIVE   Imply --force if NONINTERACTIVE is non-empty.
```

Would it make sense to also have an `./install.sh --help`?

```shell
./install.sh --help
```

osalbahr:
Adapted from `uninstall.sh`:

```console
$ ./install.sh --help
Homebrew Installer
Usage: [NONINTERACTIVE=1] [CI=1] ./install.sh [options]
    -h, --help       Display this message.
    NONINTERACTIVE   Install without prompting.
    CI               Imply NONINTERACTIVE.
$ ./install.sh -h    
Homebrew Installer
Usage: [NONINTERACTIVE=1] [CI=1] ./install.sh [options]
    -h, --help       Display this message.
    NONINTERACTIVE   Install without prompting.
    CI               Imply NONINTERACTIVE.
$ ./install.sh --foo
Warning: Unrecognized option: '--foo'
Homebrew Installer
Usage: [NONINTERACTIVE=1] [CI=1] ./install.sh [options]
    -h, --help       Display this message.
    NONINTERACTIVE   Install without prompting.
    CI               Imply NONINTERACTIVE.
```

I didn't modify the README since it already mentions the only supported option.

ShellCheck complained about the `shift` (unreachable), so I commented it out. I didn't remove it to make it easier in case a new option gets added in the future.

The ShellCheck warning:

```
$ shellcheck install.sh 

In install.sh line 99:
  shift # Command appears to be unreachable. Check usage (or ignore if invoked indirectly). shellcheck (SC2317)
  ^-- SC2317 (info): Command appears to be unreachable. Check usage (or ignore if invoked indirectly).

For more information:
  https://www.shellcheck.net/wiki/SC2317 -- Command appears to be unreachable...
```

---

Note: I assumed `CI` and `NONINTERACTIVE` are interchangeable. There was a closed issue that was solved by setting `CI=1` and I don't know if `NONINTERACTIVE=1` would have worked as well.
- https://github.com/Homebrew/install/issues/369

If they are equivalent, should `CI` be added to `uninstall.sh` for consistency?

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -39,6 +39,29 @@ then
   abort 'Bash must not run in POSIX mode. Please unset POSIXLY_CORRECT and try again.'
 fi
 
+usage() {
+  cat <<EOS
+Homebrew Installer
+Usage: [NONINTERACTIVE=1] [CI=1] $0 [options]
+    -h, --help       Display this message.
+    NONINTERACTIVE   Install without prompting.
+    CI               Imply NONINTERACTIVE.
+EOS
+  exit "${1:-0}"
+}
+
+while [[ $# -gt 0 ]]
+do
+  case "$1" in
+    -h | --help) usage ;;
+    *)
+      warn "Unrecognized option: '$1'"
+      usage 1
+      ;;
+  esac
+  # shift # Command appears to be unreachable. Check usage (or ignore if invoked indirectly). shellcheck (SC2317)
+done
+
 # string formatters
 if [[ -t 1 ]]
 then
--- a/install.sh
+++ b/install.sh
@@ -42,7 +42,7 @@ fi
 usage() {
   cat <<EOS
 Homebrew Installer
-Usage: [NONINTERACTIVE=1] [CI=1] $0 [options]
+Usage: [NONINTERACTIVE=1] [CI=1] install.sh [options]
     -h, --help       Display this message.
     NONINTERACTIVE   Install without prompting for user input
     CI               Imply NONINTERACTIVE.
--- a/install.sh
+++ b/install.sh
@@ -44,7 +44,7 @@ usage() {
 Homebrew Installer
 Usage: [NONINTERACTIVE=1] [CI=1] $0 [options]
     -h, --help       Display this message.
-    NONINTERACTIVE   Install without prompting.
+    NONINTERACTIVE   Install without prompting for user input
     CI               Imply NONINTERACTIVE.
 EOS
   exit "${1:-0}"
--- a/install.sh
+++ b/install.sh
@@ -45,7 +45,7 @@ Homebrew Installer
 Usage: [NONINTERACTIVE=1] [CI=1] install.sh [options]
     -h, --help       Display this message.
     NONINTERACTIVE   Install without prompting for user input
-    CI               Imply NONINTERACTIVE.
+    CI               Install in CI mode (e.g. do not prompt for user input)
 EOS
   exit "${1:-0}"
 }
--- a/install.sh
+++ b/install.sh
@@ -59,7 +59,6 @@ do
       usage 1
       ;;
   esac
-  # shift # Command appears to be unreachable. Check usage (or ignore if invoked indirectly). shellcheck (SC2317)
 done
 
 # string formatters


install.sh
```diff
@@ -39,6 +39,29 @@ then
   abort 'Bash must not run in POSIX mode. Please unset POSIXLY_CORRECT and try again.'
 fi
 
+usage() {
+  cat <<EOS
+Homebrew Installer
+Usage: [NONINTERACTIVE=1] [CI=1] $0 [options]
+    -h, --help       Display this message.
+    NONINTERACTIVE   Install without prompting.
```
MikeMcQuaid, line 47:
```suggestion
    NONINTERACTIVE   Install without prompting for user input
```

install.sh
```diff
@@ -39,6 +39,29 @@ then
   abort 'Bash must not run in POSIX mode. Please unset POSIXLY_CORRECT and try again.'
 fi
 
+usage() {
+  cat <<EOS
+Homebrew Installer
+Usage: [NONINTERACTIVE=1] [CI=1] $0 [options]
```
MikeMcQuaid, line 45:
```suggestion
Usage: [NONINTERACTIVE=1] [CI=1] install.sh [options]
```

install.sh
```diff
@@ -39,6 +39,29 @@ then
...
+    -h, --help       Display this message.
+    NONINTERACTIVE   Install without prompting.
+    CI               Imply NONINTERACTIVE.
```
MikeMcQuaid, line 48:
```suggestion
    CI               Install in CI mode (e.g. do not prompt for user input)
```

install.sh
```diff
@@ -39,6 +39,29 @@ then
...
+while [[ $# -gt 0 ]]
+do
+  case "$1" in
+    -h | --help) usage ;;
+    *)
+      warn "Unrecognized option: '$1'"
+      usage 1
+      ;;
+  esac
+  # shift # Command appears to be unreachable. Check usage (or ignore if invoked indirectly). shellcheck (SC2317)
```
MikeMcQuaid, line 62:
```suggestion
```

MikeMcQuaid (commented):
Thanks @osalbahr! Looking good so far.

I don't think you need to add `NONINTERACTIVE` or `CI` to `uninstall.sh` given that no-one has ever requested it.

install.sh
osalbahr: I’ve used $0 because it’s what the uninstaller uses, but I’m fine with either

install.sh
osalbahr: Is there a difference between CI and NONINTERACTIVE?

MikeMcQuaid: Not currently but want to leave the door open for it in future. There's a semantic difference.

osalbahr: What’s the semantic difference? To the best of my understanding, CI needs to be non-interactive but not sure what else might be into it.

Also, in `uninstall.sh` they make it clear that `NONINTERACTIVE` just implies `--force` so I adopted the same idea.

MikeMcQuaid: > CI needs to be non-interactive but not sure what else might be into it.

CI does need to be non-interactive but we may additionally change behaviour in future.

MikeMcQuaid:
Thanks again @osalbahr, you rock!

PR #795, opened 2023-07-28
install.sh: Change argument check from "while" to "if"

vitorgalvao:
This is not a loop but a single check.


MikeMcQuaid:
@vitorgalvao Sorry, I don't agree with the change. It's standard practise for argument parsing in Bash to use a `while` like this and it'll make adding any future arguments easier/more obvious.

PR #801, opened 2023-08-22
Delete legacy unused issue template

kevinji:
The `issue_template.md` file is not used if `blank_issues_enabled: false` is set: https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#configuring-the-template-chooser


Bo98:
Unfortunately, `blank_issues_allowed` isn't exactly enforced: https://github.com/Homebrew/install/issues/new

kevinji:
That sucks, I didn't think to check the direct link. Closing for now then unless GitHub decides to enforce the use of issue templates.

PR #810, opened 2023-10-04
install.sh: Condense (grep | awk | sed) to awk only

toobuntu:
Condense (grep | awk | sed) to awk only.

Awk can find and substitute, too.

---
Explanation:
We combine the functionality of the `grep` (find, match) and `sed` (in this case, substitute) into `awk` itself. `awk` was being used anyway, so it made me wonder why bother doing things with other tools when `awk` can also do those same things.

When assigning a value to a field variable--in this case field `$1` is "assigned" the value of itself--awk recompiles the record (field `$0`) by concatenating all of the fields with the output field separator (OFS), which is space by default. This will also remove all leading and trailing spaces from the record when using the default FS. (https://unix.stackexchange.com/questions/568666/how-does-awk-1-1-remove-extra-spaces)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -817,9 +817,7 @@ then
   execute_sudo "${TOUCH[@]}" "${clt_placeholder}"
 
   clt_label_command="/usr/sbin/softwareupdate -l |
-                      grep -B 1 -E 'Command Line Tools' |
-                      awk -F'*' '/^ *\\*/ {print \$2}' |
-                      sed -e 's/^ *Label: //' -e 's/^ *//' |
+                      awk '/Label:[[:blank:]]+Command Line Tools/ {\$1=\$1; sub(/^.*Label:[[:blank:]]+/,\"\"); print; next}' |
                       sort -V |
                       tail -n1"
   clt_label="$(chomp "$(/bin/bash -c "${clt_label_command}")")"


MikeMcQuaid (commented):
@toobuntu Thanks for the PR!

> Condense (grep | awk | sed) to awk only. Awk can find and substitute, too.

It can but: in cases like this: this is considerably less readable to the typical user who has typical levels of familiarity with these tools. By the time that `awk` is being used at this level: I think `perl` or similar provides a better experience.

In this case, though: given it's not performance critical in anyway: I think the multi-pipe, multi-command approach from before leaves this as more readable.

Open to disagreement from other @Homebrew/maintainers

fxcoudert:
Strongly agree: readability is very important, especially in the install code, which is something that has very little testing in CI and where changes are deployed directly to new users.

vitorgalvao:
If `awk` were to be used in this context, it could replace the `sort` and `tail` as well. Though I’m not advocating for it.

> I think `perl` or similar provides a better experience.

I understand you’re giving an example and not advocating for using `perl`, but for anyone who bumps into this in the future I’ll note that `perl` is one of the languages Apple will allegedly remove from macOS at some point, so avoid it on the install script.

On the subject of readability, this PR does highlight that there could be improvements even in these lines. Specifically, in a script we should always use the long form of flags (`grep --before-context 1 --extended-regexp`) instead of the short versions (`grep -B 1 -E`) because the latter are less readable and make users have to constantly check the manual.

toobuntu:
> this is considerably less readable to the typical user who has typical levels of familiarity with these tools.

I understand your concern.

(And as for re-implementing `sort -V | tail -n1` in awk, I agree that would be way too complex and completely unnecessary for "the typical user who has typical levels of familiarity with these tools.")

---

Just to note for discussion, as it would apply to Strap as well as Homebrew:
What is the function of `grep -B 1` here? Was there something important about the line preceding the match?

As mentioned in https://github.com/Homebrew/install/issues/233, the output of `softwareupdate -l` changed in macOS 15 Catalina. When the grep was first implemented in Boxen, and carried over here, the output was something closer to
```
* Command Line Tools (macOS Mojave version 10.14) for Xcode-10.3
        #        Command Line Tools (macOS Mojave version 10.14) for Xcode (10.3), 199140K [recommended]
```
and since then Apple added the `Label: ` and `Title: ` prefixes.
```
* Label: Command Line Tools beta 5 for Xcode-11.0
        #     Title: Command Line Tools beta 5 for Xcode, Version: 11.0, Size: 224804K, Recommended: YES,
```

Would it be reasonable to drop the `grep -B 1` and instead match only lines containing `Label: `?

`sed` alone can do it:
```
sed -E '/Label:[[:blank:]]+Command Line Tools/ s/^[^.]*Command Line Tools/Command Line Tools/'
```

But that might also be a readability issue.

```
grep -E 'Label:[[:blank:]]+Command Line Tools' | 
sed 's/^[^.]*Command Line Tools/Command Line Tools/'
```
or `sed -e 's/^[^.]*Label: //'`, for instance.

Bo98:
> Just to note for discussion, as it would apply to Strap as well as Homebrew:
> What is the function of `grep -B 1` here? Was there something important about the line preceding the match?

First line is the ID you pass to `softwareupdate -i` and the second is the display name. Originally the ID wasn't really greppable.

Old example:
```
* 031-1006-5.0.1.0
    Command Line Developer Tools for OS X Mavericks (5.0.1.0), 99103K [recommended]
```

Nowadays it probably matters less. I guess it depends if you trust the ID to remain stable.

Homebrew installer probably doesn't need to support anything before macOS 12 (though I did use it for 10.13 CI semi-recently), but certainly not anything before macOS 10.11. I don't know what Strap targets to support.

toobuntu:
Closed as not planned.

MikeMcQuaid:
Thanks for the PR and for understanding @toobuntu! Hope we can get another PR from you in future that we can merge ❤️

PR #811, opened 2023-10-22
install: add fish post install instructions

apainintheneck:
This adds the correct config file for the fish shell needed to load the brew environment. Fish ends up using the `~/.config/fish/config.fish` file during startup to load environment variables and scripts most of the time. This is also used by asdf and ghcup for instance.

This was brought up in discussion here:
- https://github.com/orgs/Homebrew/discussions/4412
- https://github.com/orgs/Homebrew/discussions/4722

More info:
- https://fishshell.com/docs/current/index.html#configuration

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1029,6 +1029,9 @@ case "${SHELL}" in
       shell_rcfile="${ZDOTDIR:-"${HOME}"}/.zprofile"
     fi
     ;;
+  */fish*)
+    shell_rcfile="${HOME}/.config/fish/config.fish"
+    ;;
   *)
     shell_rcfile="${ENV:-"${HOME}/.profile"}"
     ;;


BarbzYHOOL:
it adds a line in the fish config?
it could also add a fish function in `.config/fish/conf.d` either in home directory or in /etc/

apainintheneck:
> it adds a line in the fish config?
> it could also add a fish file in `.config/fish/conf.d` either in home directory or in /etc/
> like "homebrew_env.fish"

@BarbzYHOOL Yep, that's the idea. Add a file to `.config/fish/conf.d` is another option but it means adding more special handling for the fish shell. This works similar to how `zsh` and `bash` are handled already. Is there any advantage to doing it this way?

apainintheneck:
CC: @ZhongRuoyu just in case you have any thoughts on this since you were involved in one of the threads.

SMillerDev (approved):

MikeMcQuaid (approved):

ZhongRuoyu (approved):
Not actually a fish user, but this looks good to me 👍

BarbzYHOOL:
@apainintheneck yes, the advantage is to not tamper with the user's config

For example, if you auto install stuff in your config
It's the same thing with all config file systems, when there is a directory dedicated to subconfig files, imo it should be used

apainintheneck:
@BarbzYHOOL I see your point. I misunderstood your initial question though. This doesn't actually add anything to your config file directly. It just prints a message to suggest adding those lines to your config file if they don't already exist there. Of course, you can add them to a separate config file if you prefer that.

You can see how the `shell_rcfile` variable is used here:
https://github.com/Homebrew/install/blob/85c5f4b57452dbd1c7ebc01a021548d2ceaf2b64/install.sh#L1040-L1068

BarbzYHOOL:
ok ok fine then

But I don't see it in the snippet above, it just says to run "eval"

PR #812, opened 2023-10-22
Fix installation in environments where sudo is missing

slonopotamus:
Homebrew allows running as root in some environments (Docker, Kubernetes, GitHub Actions, etc). When we're root, there is no point using sudo.
So, just invoke commands directly, thus removing requirement to have a functional `sudo` command.

I personally plan to use this in [macOS containers](https://macoscontainers.org).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -262,7 +262,7 @@ execute() {
 
 execute_sudo() {
   local -a args=("$@")
-  if have_sudo_access
+  if [[ "${EUID:-${UID}}" != "0" ]] && have_sudo_access
   then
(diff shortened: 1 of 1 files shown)


osalbahr (commented):
Is there a reason you need `"${EUID:-${UID}}"` twice, before calling the function as well as at the top of the function?

slonopotamus:
Reorganised code a bit so it should be clearer why I need two checks.

MikeMcQuaid (approved):
Thanks so much for your first contribution! Without people like you submitting PRs we couldn't run this project. You rock, @slonopotamus!

PR #815, opened 2023-11-03
correct messages to "i.e." (i.e. id est)

bgerrity:
https://english.stackexchange.com/questions/1629/e-g-versus-i-e


vitorgalvao:
Current usage is correct. You don’t need to be an administrator to have `sudo` access.

bgerrity:
> Only administrator users can use sudo.

It may be possible to be on the sudoers file and not an administrator but would that be subverting the design ([docs](https://support.apple.com/guide/terminal/enter-administrator-commands-apd5b0b6259-a7d4-4435-947d-0dff528912ba/mac))?

PR #833, opened 2024-01-20
fix - changes in the install.sh for misplaced single quote.

Rushour0:
A single quote was misplaced in the section of `Run these two commands in your terminal to add Homebrew`
Fixing this will make the post install process run smoothly without any hiccup.


Bo98:
Original looks correct:

```
            v_______________________________________________v
(echo; echo 'eval "\$(${HOMEBREW_PREFIX}/bin/brew shellenv)"')
^____________________________________________________________^
                     ^____________________________________^
```

Rushour0:
The original was probably misse read on my terminal

Issue #829, opened 2023-12-11
Don't checkout Homebrew/brew `master` for non-developers


Bo98:
Currently we check out `master` when cloning `Homebrew/brew`: https://github.com/Homebrew/install/blob/3d2b88baf18424b561766fcdac31be06553e0e07/install.sh#L933

`brew update` will correct this to be on a stable tag, but picking `master` initially is not ideal because it means `brew update` will perform a downgrade and we don't test downgrades (but will accept fixes where there are issues). An example of an issue caused by a downgrade is #825.

Instead, short of a full stable-branch setup that `brew update` does, an initial detached tag checkout would be much better. Detached tag checkouts now work without downgrades since https://github.com/Homebrew/brew/commit/2a02038690db452b71c26d5beb9d1f07b068cacf.

MikeMcQuaid:
Fixed by https://github.com/Homebrew/install/pull/839

PR #839, opened 2024-03-14
install: use stable tag.

MikeMcQuaid:
Currently we always install onto the latest `master`. Instead, let's correctly use the latest tag.

See also https://github.com/Homebrew/brew/pull/16891

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -914,8 +914,14 @@ ohai "Downloading and installing Homebrew..."
 
   execute "${USABLE_GIT}" "fetch" "--force" "origin"
   execute "${USABLE_GIT}" "fetch" "--force" "--tags" "origin"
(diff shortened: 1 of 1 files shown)


vitorgalvao:
Should the [“untar anywhere” instructions](https://docs.brew.sh/Installation#alternative-installs) be updated as well, then, for parity?

Bo98 (approved):

Bo98:
Also fixes https://github.com/Homebrew/install/issues/829 I think?

Bo98:
`update.sh` technically does `refs/tags/#{LATEST_GIT_TAG}` to avoid any name clash, though not sure if that matters much.

MikeMcQuaid:
> `update.sh` technically does `refs/tags/${LATEST_GIT_TAG}` to avoid any name clash, though not sure if that matters much.

@Bo98 yeh, can't see it mattering here but can update if it does! 

Thanks all!

PR #849, opened 2024-05-17
Update install.sh to use /usr/bin/stat

DG12:
Using customized stat in $PATH can be in conflict with standard.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -160,7 +160,7 @@ then
   fi
   HOMEBREW_CACHE="${HOME}/Library/Caches/Homebrew"
 
-  STAT_PRINTF=("stat" "-f")
+  STAT_PRINTF=("/usr/bin/stat" "-f")
   PERMISSION_FORMAT="%A"
   CHOWN=("/usr/sbin/chown")
   CHGRP=("/usr/bin/chgrp")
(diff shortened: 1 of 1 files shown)


install.sh
```diff
@@ -160,7 +160,7 @@ then
   fi
   HOMEBREW_CACHE="${HOME}/Library/Caches/Homebrew"
 
-  STAT_PRINTF=("stat" "-f")
+  STAT_PRINTF=("/usr/bin/stat" "-f")
```
woodruffw, line 163: In practice this should always be right, but I can't find a good POSIX/similar source that asserts that `stat` is always at `/usr/bin/stat`. Then again maybe this doesn't matter, since we assume similar things for `chown` and `chgrp` below.

mistydemeo: I checked the file list for all the distros we advertise support for (Debian, Ubuntu, Fedora, CentOS, Red Hat, Arch). In all cases `stat` is provided by coreutils and present at `/usr/bin/stat`. I agree though, we don't necessarily have a guarantee it'll always be provided there in *every* distro; for example, alpine's busybox lives in `/bin`.

MikeMcQuaid (approved):
Thanks @DG12 for the PR and @mistydemeo for the research!

install.sh
DG12: To be more compatible perhaps someone could enhance this (and others if necessary )  to include _something like_:

```
UTILDIR ='/usr/bin'
if [ ! -x $UTILDIR/stat ]; then UTILDIR='/bin' ; fi
STAT_PRINTF=("$UTILDIR/stat" "-f")
...
STAT_PRINTF=("$UTILDIR/stat" "--printf")
```

MikeMcQuaid: It's fine as-is until someone complains.

PR #871, opened 2024-07-26
Remove erroneous parentheses in sample command

jsquaredz:
There is an extra parentheses in the sample which gives a parsing error if you copy the command directly from the sample output.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1048,7 +1048,7 @@ EOS
 else
   cat <<EOS
 - Run these two commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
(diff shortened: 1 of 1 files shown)


mistydemeo:
That outside paranthesis closes the open one that comes before `echo`. Did you include the open paren in the command you pasted?

p-linnane:
These are also two separate commands to run, as it says above, rather than one.

gromgit:
A simple visual inspection shows that the parentheses _and_ braces in the original code balance perfectly, so I don't know why you think the final parenthesis is erroneous.

jsquaredz:
Yikes.  That's embarrassing.  When I copied it I missed the first paren.  Sorry folks . Keep up the great work!

PR #883, opened 2024-09-26
Make some git invocations quiet

ZhongRuoyu:
Currently, all the branches and tags are printed out by `git fetch`, which I believe our end users are not concerned about.

This change suppresses some of the Git outputs with the `--quiet` option, so no branches or tags are printed out. As the progress bar can still be useful, `--progress` is used to keep it enabled.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -927,8 +927,8 @@ ohai "Downloading and installing Homebrew..."
   # make sure symlinks are saved as-is
   execute "${USABLE_GIT}" "config" "--bool" "core.symlinks" "true"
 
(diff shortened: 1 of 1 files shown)


ZhongRuoyu:
For comparison, see [old](https://github.com/Homebrew/install/actions/runs/11038271869/job/30661060500#step:8:39) versus [new](https://github.com/Homebrew/install/actions/runs/11045312709/job/30682716118?pr=883#step:12:35).

carlocab (approved):

MikeMcQuaid (approved):
Makes sense, thanks @ZhongRuoyu!

PR #898, opened 2024-11-22
Remove redundant escape characters

Rylan12:
Since we write the insufficient permissions message in a heredoc, the `\"` characters are rendered as-is instead of as an escaped quote character.

Before:

```
Insufficient permissions to install Homebrew to \"/home/linuxbrew/.linuxbrew\" (the default prefix).
```

After:

```
Insufficient permissions to install Homebrew to "/home/linuxbrew/.linuxbrew" (the default prefix).
```


MikeMcQuaid (approved):
Thanks @Rylan12!

Issue #840, opened 2024-03-23
Installation fails on Linux with snap installed curl


orsenthil:
- [x] your problem was from running the official `install` or `uninstall` script?
- [ ] after installation: ran `brew config` and `brew doctor` and included their output with your issue? If you couldn't install: provided your OS version with the output of your issue?

<!-- To help us debug your issue, please complete these sections: -->

## What you were trying to do (and why)

I was trying to install homebrew on Ubuntu Linux 22.04
I realized it didn't come with curl, but recommended me to do `snap` install curl.

> sudo snap install curl

It installed curl 8.1.2

```
$ /snap/bin/curl --version
curl 8.1.2 (x86_64-pc-linux-gnu) libcurl/8.1.2 OpenSSL/1.1.1f zlib/1.2.11 brotli/1.0.7 zstd/1.4.4 libidn2/2.2.0 libpsl/0.21.0 (+libidn2/2.2.0) libssh/0.9.3/openssl/zlib nghttp2/1.40.0 librtmp/2.3 libgsasl/1.8.1
Release-Date: 2023-05-30
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtmp rtsp scp sftp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS brotli gsasl GSS-API HSTS HTTP2 HTTPS-proxy IDN IPv6 Kerberos Largefile libz NTLM NTLM_WB PSL SPNEGO SSL threadsafe TLS-SRP UnixSockets zstd
```

And with this curl version, the Homebrew installation fails in a reproducible manner.

## What happened (include command output)
```
Switched to a new branch 'stable'
==> Tapping homebrew/core because `$HOMEBREW_NO_INSTALL_FROM_API` is set.
remote: Enumerating objects: 1918864, done.
remote: Counting objects: 100% (312/312), done.
remote: Compressing objects: 100% (153/153), done.
remote: Total 1918864 (delta 196), reused 263 (delta 159), pack-reused 1918552
Receiving objects: 100% (1918864/1918864), 722.60 MiB | 8.09 MiB/s, done.
Resolving deltas: 100% (1369295/1369295), done.
From https://github.com/Homebrew/homebrew-core
 * [new branch]              master     -> origin/master
HEAD is now at 0f8742774a7 Merge pull request #166913 from Homebrew/tests-stable-syntax
==> Updating Homebrew...
[...]

PR #902, opened 2024-12-02
install.sh: ignore curl snap if found

gromgit:
Snap upstream [refuses to relax strict security confinement](https://forum.snapcraft.io/t/classic-confinement-request-for-curl/24611/13) on curl. This breaks the install process, as at least two different reports reveal: https://github.com/Homebrew/install/issues/840 https://github.com/orgs/Homebrew/discussions/5776#discussioncomment-11427538

This patch also warns the user about this rejection, because the curl snap is quite likely the only curl installed on a user's system _and_ is otherwise recent enough to pass the version test, so it provides the necessary clarification to the next thing the user sees:
```
The version of cURL that was found does not satisfy requirements for Homebrew.
Please install cURL ${REQUIRED_CURL_VERSION} or newer and add it to your PATH.
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -393,6 +393,12 @@ test_curl() {
     return 1
   fi
 
+  if [[ "$1" == "/snap/bin/curl" ]]
+  then
+    warn "Ignoring $1 (curl snap is too restricted)"
+    return 1
+  fi
+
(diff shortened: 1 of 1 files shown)


carlocab (approved):
Thanks!

PR #912, opened 2024-12-13
install.sh: update comment for non-default Git remote setup

ZhongRuoyu:
The existing comment is wrong, the commands are not for setting up PATHs:

```
- Run these commands in your terminal to add the non-default Git remote for Homebrew/brew:
    echo '# Set PATH, MANPATH, etc., for Homebrew.' >> /Users/ruoyu/.zprofile
    echo 'export HOMEBREW_BREW_GIT_REMOTE="https://example.org/brew.git"' >> /Users/ruoyu/.zprofile
    export HOMEBREW_BREW_GIT_REMOTE="https://example.org/brew.git"
```

With this change, the comment is updated to reflect the correct purpose of the commands:

```
- Run these commands in your terminal to add the non-default Git remote for Homebrew/brew:
    echo '# Set non-default Git remote for Homebrew/brew.' >> /Users/ruoyu/.zprofile
    echo 'export HOMEBREW_BREW_GIT_REMOTE="https://example.org/brew.git"' >> /Users/ruoyu/.zprofile
    export HOMEBREW_BREW_GIT_REMOTE="https://example.org/brew.git"
```

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1076,7 +1076,7 @@ then
     plural="s"
   fi
   printf -- "- Run these commands in your terminal to add the non-default Git remote%s for %s:\n" "${plural}" "${non_default_repos}"
-  printf "    echo '# Set PATH, MANPATH, etc., for Homebrew.' >> %s\n" "${shell_rcfile}"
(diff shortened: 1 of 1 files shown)


SMillerDev (approved):

MikeMcQuaid:
Thanks @ZhongRuoyu, good catch.

PR #924, opened 2024-12-30
uninstall.sh: don't check /usr/local for residual files

gromgit:
It's a very common install prefix, so any residual files found here are likely false positives.

Followup to #923.

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -459,7 +459,7 @@ then
   fi
 fi
 
-residual_dirs=("${HOMEREW_REPOSITORY}")
+residual_dirs=("${HOMEBREW_REPOSITORY}")
 if [[ "${HOMEBREW_PREFIX}" != "/usr/local" ]]
 then
   residual_dirs+=("${HOMEBREW_PREFIX}")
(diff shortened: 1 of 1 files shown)


uninstall.sh
```diff
@@ -459,7 +459,12 @@ then
   fi
 fi
 
-dir_children "${HOMEBREW_REPOSITORY}" "${HOMEBREW_PREFIX}" |
+residual_dirs=("${HOMEREW_REPOSITORY}")
```
mistydemeo, line 462: Noticed there's a typo in this variable name:

```suggestion
residual_dirs=("${HOMEBREW_REPOSITORY}")
```

maxim-belkin (approved):
I think this is very reasonable.

uninstall.sh
gromgit: <facepalm>

gromgit:
>```
> WslRegisterDistribution failed with error: 0x800701bc
>
> Error: 0x800701bc WSL 2 requires an update to its kernel component. For information please visit https://aka.ms/wsl2kernel
>```

MikeMcQuaid (approved):
Makes sense to me, thanks @gromgit!

PR #933, opened 2025-01-31
Fix installation issue when originally attempted at broken state

### LINKED ISSUE #766: Installation keeps on failing if originally attempted at broken state

cokyman:
Fixes #766

Add a check for existing broken installation in `install.sh`.

* Add a check for an existing broken installation at the start of the script
* Add a function to remove the broken installation if detected
* Call the function to remove the broken installation if detected

---

For more details, open the [Copilot Workspace session](https://copilot-workspace.githubnext.com/Homebrew/install/pull/933?shareId=2b505cfd-4204-40e0-a495-e2d919c4ee00).


gromgit:
A better test might be if `git -C $HOMEBREW_PREFIX status -s` exits non-zero. That directly addresses the failure condition you referenced, rather than giving a false positive on a healthy installation.

MikeMcQuaid:
> A better test might be if `git -C $HOMEBREW_PREFIX status -s` exits non-zero. That directly addresses the failure condition you referenced, rather than giving a false positive on a healthy installation.

Agreed. This doesn't seem to detect whether an installation is actually broken but just that it already exists.

PR #939, opened 2025-02-08
userspecified install prefix

igueths:
- **Install: allow the user to specify an install prefix.**
- **install: modify variable names to more closely match the uninstaller.**
- **install: fix variable interpolation, make sure the flag variable has a default value.**
- **install: remove comment.**
- **README: remove duplicate word.**
- **README: document how to use HOMEBREW_INSTALL_PREFIX.**


p-linnane:
Thanks for taking the time to open a PR. Installing outside the default prefix is [specifically called out as unsupported](https://docs.brew.sh/Installation#untar-anywhere-unsupported).

PR #950, opened 2025-03-04
Add caching to CLT installation check

willmcginnis:
Improve the should_install_command_line_tools function by:
1. Caching the result in SHOULD_INSTALL_CLT to avoid redundant file existence checks.
2. Using a local variable (ret) to capture return values for added clarity.

This change slightly increases code size but provides a more efficient
and readable approach.

**Upsides:**
- Efficiency improvement by skipping redundant checks on subsequent calls.

**Downsides:**
- Introduces a new global variable (SHOULD_INSTALL_CLT).
- Increases the function length by about 10 lines.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -352,18 +352,28 @@ check_run_command_as_root() {
 }
 
 should_install_command_line_tools() {
-  if [[ -n "${HOMEBREW_ON_LINUX-}" ]]
-  then
+  if [[ -n "${SHOULD_INSTALL_CLT-}" ]]; then
+    return "${SHOULD_INSTALL_CLT}"
+  fi
+
+  if [[ -n "${HOMEBREW_ON_LINUX-}" ]]; then
+    SHOULD_INSTALL_CLT=1
     return 1
   fi
 
-  if version_gt "${macos_version}" "10.13"
-  then
+  local ret
+
+  if version_gt "${macos_version}" "10.13"; then
     ! [[ -e "/Library/Developer/CommandLineTools/usr/bin/git" ]]
+    ret=$?
   else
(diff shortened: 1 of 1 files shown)


gromgit:
I'm not sure calling `should_install_command_line_tools` is computationally expensive enough to warrant caching, but even if it is, the required change can be reduced to just 3 lines (1 new, 2 changed):
```
diff --git a/install.sh b/install.sh
index 71ca08b..4b588fa 100755
--- a/install.sh
+++ b/install.sh
@@ -732,6 +732,7 @@ fi
 if should_install_command_line_tools
 then
   ohai "The Xcode Command Line Tools will be installed."
+  SHOULD_INSTALL_CLT=1
 fi
 
 non_default_repos=""
@@ -834,7 +835,7 @@ then
   execute "${TOUCH[@]}" "${HOMEBREW_CACHE}/.cleaned"
 fi
 
-if should_install_command_line_tools && version_ge "${macos_version}" "10.13"
+if [[ -n "${SHOULD_INSTALL_CLT-}" ]] && version_ge "${macos_version}" "10.13"
 then
   ohai "Searching online for the Command Line Tools"
   # This temporary file prompts the 'softwareupdate' utility to list the Command Line Tools
@@ -859,7 +860,7 @@ then
 fi
 
 # Headless install may have failed, so fallback to original 'xcode-select' method
-if should_install_command_line_tools && test -t 0
+if [[ -n "${SHOULD_INSTALL_CLT-}" ]] && test -t 0
 then
   execute "/usr/bin/xcode-select" "--install"
```

MikeMcQuaid:
Passing on this, sorry. As @gromgit mentioned: this is not expensive enough to warrant caching. Also, there's a non-zero chance at install time that these files are present at some points and not at others.

PR #960, opened 2025-03-21
linux: cleanup default Homebrew parent dir

gromgit:
Addresses https://github.com/orgs/Homebrew/discussions/6033

Also check for `brew` before trying to run it.


MikeMcQuaid (approved):
Makes sense, thanks @gromgit!

PR #990, opened 2025-06-25
install.sh: Homebrew/brew uses `main` default branch.

MikeMcQuaid:
`master` is still around so it will keep working but let's do things nicely here.


install.sh
```diff
@@ -943,11 +943,12 @@ ohai "Downloading and installing Homebrew..."
   cd "${HOMEBREW_REPOSITORY}" >/dev/null || return
 
   # we do it in four steps to avoid merge errors when reinstalling
-  execute "${USABLE_GIT}" "-c" "init.defaultBranch=master" "init" "--quiet"
+  execute "${USABLE_GIT}" "init" "--quiet"
```
ZhongRuoyu, line 946: I believe the reason we've had this here is to override user settings when people have it set to `main` or something else in their Git config. (Git's default is `master`.) So I think we should continue to override this but instead change the branch name to `main`?

ZhongRuoyu, line 946:
```suggestion
  execute "${USABLE_GIT}" "-c" "init.defaultBranch=main" "init" "--quiet"
```

ZhongRuoyu:
Whoops, this was automerged -- #991

MikeMcQuaid:
@ZhongRuoyu thought I applied, weird, sorry! thanks for #991

PR #1001, opened 2025-07-08
install.sh: move PATH warning after update

gromgit:
Otherwise, it'll get lost in the voluminous `brew update` output.

Also fix the string spacing, to maintain a consistent two-space overflow indent with other multiline messages.


MikeMcQuaid (approved):
Thanks!

carlocab (approved):
Thanks!

PR #1014, opened 2025-07-30
Prevent git column.ui configuration from affecting install script

jacobwhall:
If a user has `column.ui` set to `always`, the Homebrew installation script fails. This PR fixes that by guaranteeing a non-columnar layout when parsing tags.


MikeMcQuaid (approved):
Good catch, thanks!

PR #1022, opened 2025-08-19
Add flag to allow root install

Emma-Fitzmaurice:
This PR adds a flag to allow the script to run as root. 
check_run_command_as_root() already makes some exceptions to allow for running as root, but it's non-exhaustive so I think it is reasonable to allow for an override flag. 

Sorry if this has been considered and rejected before, i couldn't find anything to that effect


MikeMcQuaid:
Passing on this, sorry. The calls to Homebrew itself within this script will fail when run under root too.

PR #1030, opened 2025-09-17
Add a check for Xcode.app presence

ghost:
The Xcode application ships all the tools from the standalone CLT at ``/Applications/Xcode.app/Contents/Developer``, but installing brew currently forces a duplicate CLT installation separate from the Xcode application. This change should now make the install script respect the Xcode application if present.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -378,7 +378,8 @@ should_install_command_line_tools() {
 
   if version_gt "${macos_version}" "10.13"
   then
-    ! [[ -e "/Library/Developer/CommandLineTools/usr/bin/git" ]]
+    ! [[ -e "/Library/Developer/CommandLineTools/usr/bin/git" ]] ||
+      ! [[ -e "/Applications/Xcode.app/Contents/Developer/usr/bin/git" ]]
   else
     ! [[ -e "/Library/Developer/CommandLineTools/usr/bin/git" ]] ||
       ! [[ -e "/usr/include/iconv.h" ]]


nesnrve:
@SMillerDev that'd be news to me, what's your reference?

nesnrve:
@carlocab Cc: @SMillerDev

I've now read https://github.com/Homebrew/install/pull/509, https://github.com/Homebrew/install/pull/586, https://github.com/Homebrew/install/pull/757 and https://github.com/Homebrew/install/pull/340. All of which are rehearsing the same statement made in here, that the CLT from the Xcode application and the standalone CLT are "not the same thing", without any references to support that claim whatsoever.

As to my observation above, the Xcode CLT ships everything you get from the standalone CLT, so if a dependance can be declared on the latter then I see no reason why not former... if it were vice versa, then it'd make sense given the standalone CLT doesn't ship everything that's available from the Xcode application.

p-linnane:
The fact that major formulae like `gcc` and `python@3.x` have issues without CLT being installed is reason enough. Check out the formula files if you want, but there are comments explaining issues with incorrect sysroot, headers that can't be found, etc. unless CLT is installed.

Further, multiple maintainers have stated both in this thread, and the ones you linked, that we won't support this. There's no real obligation on our part to 'prove' to you, or anyone else, why that is.

MikeMcQuaid:
> without any references to support that claim whatsoever.

Here's a reference: we're Homebrew maintainers. I've been doing this for 16 years. I'm in direct communication with Apple around things like the CLT. Homebrew is part of the reason the CLT exists. We know than you about this. We're not obliged (nor is it a good use of our time) to explain to you any more details than have already been provided.

PR #1063, opened 2025-12-28
install.sh: do not echo newline to `shell_rcfile`

fkobi:
The initial installation should be as simple as possible. The newline is nice but it is 1/3rd of what the user has to do after the install.

If we want to keep it I suggest swapping the echo to printf and using `\n`.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1108,7 +1108,6 @@ EOS
 else
   cat <<EOS
 - Run these commands in your terminal to add Homebrew to your ${tty_bold}PATH${tty_reset}:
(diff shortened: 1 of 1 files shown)


woodruffw:
> From my experiences people not used to working in the terminal just copy-paste commands without trying to comprehend them so in my opinion having 2 lines to copy-paste is better UX than 3.

Sorry, I don't totally understand this: if they're copy-pasting, what's the material difference between it being two versus three lines? Either way they're doing a multi-line select.

PR #1088, opened 2026-03-24
fix: remove stale origin/master ref before post-install update

### LINKED ISSUE #766: Installation keeps on failing if originally attempted at broken state

saschabuehrle:
Fixes #766

When users retry installation after an old/broken install attempt, the local brew repo can still contain `refs/remotes/origin/master`. If a checked-out brew revision references that stale remote branch, the final `brew update --force --quiet` step can fail with `Not a valid ref: refs/remotes/origin/master`.

This cleans up the stale `origin/master` ref (when present) right after fetch and before the post-install update.

Greetings, saschabuehrle


MikeMcQuaid (approved):
Good idea, seems basically instant. Thanks!

Bo98:
Is `master` still used despite `main` now being the default branch?

PR #1100, opened 2026-05-18
install.sh: print correct dnf group name on non-Fedora distros

TrevorBurnham:
On an Amazon Linux 2023 instance, I got this output from the Homebrew installer:

```
- Install Homebrew's dependencies if you have sudo access:
    sudo dnf group install development-tools
```

But that command failed:

```sh
$ sudo dnf group install development-tools
Amazon Linux 2023 Internal Shared Repository                                                          20 kB/s | 863  B     00:00
Module or Group 'development-tools' is not available.
Error: Nothing to do.
```

The [Homebrew on Linux docs](https://docs.brew.sh/Homebrew-on-Linux) acknowledge this issue: They say that you should run `sudo dnf group install development-tools` under Fedora, but `sudo dnf group install 'Development Tools'` under CentOS Stream or RHEL. (Amazon Linux 2023 follows the CentOS/RHEL convention.) The installer, however, prints the `development-tools` command whenever `dnf` exists.

The fix here is to check `/etc/os-release` for `id=fedora`. If that file indicates that this is a Fedora system, print the `development-tools` command; otherwise, print the `Development Tools` command.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1147,7 +1147,15 @@ then
     echo "    sudo apt-get install build-essential"
   elif [[ -x "$(command -v dnf)" ]]
   then
-    echo "    sudo dnf group install development-tools"
+    # Fedora uses the lowercase `development-tools` group id; most other
+    # dnf-based distros use the `Development Tools` name instead.
+    # shellcheck disable=SC1091
+    if [[ -r /etc/os-release ]] && (. /etc/os-release && [[ "${ID:-}" == "fedora" ]])
+    then
+      echo "    sudo dnf group install development-tools"
+    else
+      echo "    sudo dnf group install 'Development Tools'"
+    fi
   elif [[ -x "$(command -v yum)" ]]
   then
     echo "    sudo yum groupinstall 'Development Tools'"


MikeMcQuaid (approved):
Thanks!

install.sh
```diff
@@ -1147,7 +1147,14 @@ then
     echo "    sudo apt-get install build-essential"
   elif [[ -x "$(command -v dnf)" ]]
   then
-    echo "    sudo dnf group install development-tools"
+    # Fedora uses the lowercase `development-tools` group id; most other
+    # dnf-based distros use the `Development Tools` name instead.
+    if [[ -r /etc/os-release ]] && grep -q '^ID=fedora$' /etc/os-release
```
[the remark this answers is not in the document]
TrevorBurnham: Updated to match this recommendation (with different grep syntax).

install.sh
```diff
@@ -1147,7 +1147,14 @@ then
     echo "    sudo apt-get install build-essential"
   elif [[ -x "$(command -v dnf)" ]]
   then
-    echo "    sudo dnf group install development-tools"
+    # Fedora uses the lowercase `development-tools` group id; most other
+    # dnf-based distros use the `Development Tools` name instead.
+    if [[ -r /etc/os-release ]] && (. /etc/os-release && [[ "${ID:-}" == "fedora" ]])
```
MikeMcQuaid, line 1153: `shellcheck` error that needs fixed 👇🏻

TrevorBurnham: I've added a `shellcheck disable` comment, which is consistent with established patterns in this repo.

PR #1106, opened 2026-06-10
install.sh/ring_bell: make the function more concise

fkobi:
This style is used elsewhere:
https://github.com/Homebrew/install/blob/280cbc9adffcbdef15dd1c9d991ef2d1dd7cfc9c/install.sh#L351

It makes the code more readable


MikeMcQuaid:
I don't think this is more readable, sorry.

PR #1121, opened 2026-07-22
install.sh: use Development Tools group on Amazon Linux 2023

The-Alchemist:
## Summary
- Amazon Linux 2023 sets `ID=amzn` and `ID_LIKE=fedora`, so the post-install dependency hint currently recommends `sudo dnf group install development-tools`.
- On AL2023 that group id does not exist (`Module or Group 'development-tools' is not available`). The available group is `Development Tools` (group id: `development`).
- Exclude `ID=amzn` from the Fedora/RHEL `development-tools` branch so AL2023 gets `sudo dnf group install 'Development Tools'`.

## Questions

Should I add a `Dockerfile` test case or something?  Feedback requested. :)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -1144,8 +1144,11 @@ then
   then
     # Fedora and RHEL-like distros use the lowercase `development-tools`
     # group id; most other dnf-based distros use the `Development Tools` name instead.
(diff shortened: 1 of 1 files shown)


p-linnane (approved):
Good catch. Thanks for contributing to Homebrew @The-Alchemist!

PR #1135, opened 2026-08-30
Retry protected uninstall paths with sudo

### LINKED ISSUE #1134: Uninstall fails due to not enought permission.
# Please note we will close your issue without comment if you delete, do not read or do not fill out the issue checklist below and provide ALL the requested information. If you repeatedly fail to use the issue template, we will block you from ever submitting issues to Homebrew again.

- [x] you carefully read [Git checkout or network failures](https://docs.brew.sh/Common-Issues#git-checkout-or-network-failures) and confirmed that the problem is not caused by your connection to GitHub or `curl` configuration?
- [-] after installation: ran `brew config` and `brew doctor` and included their output with your issue? If you couldn't install: provided your OS version with the output of your issue? (__not applicable since it's uninstallation process__)

I was trying to achieve migration from x64 to arm since I'm on Apple silicon mac. 

The official uninstall script asked password right after when it needed the most. 
And it recommended using `sudo`, which also failed because it no longer detect homebrew is there. 

Although I know I can manually delete directories one by one, the script should have handled it elegantly. 

```
curl -fsSLO https://raw.githubusercontent.com/Homebrew/install/HEAD/uninstall.sh
/bin/bash uninstall.sh --path /usr/local
Warning: This script will remove:
/etc/paths.d/homebrew
/Users/myusername/Library/Caches/Homebrew/
/Users/myusername/Library/Logs/Homebrew/
/usr/local/bin/brew -> /usr/local/bin/brew
/usr/local/Caskroom/
/usr/local/Cellar/
Are you sure you want to uninstall Homebrew? This will remove your installed packages! [y/N] y
==> Removing Homebrew installation...
Warning: Failed to delete /etc/paths.d/homebrew
rm: /etc/paths.d/homebrew: Permission denied
Warning: Failed to delete /usr/local/Caskroom
rm: /usr/local/Caskroom: Permission denied
Warning: Failed to delete /usr/local/Cellar
rm: /usr/local/Cellar: Permission denied
==> Removing empty directories...
Password:
==> /usr/bin/sudo /usr/bin/find /usr/local/bin /usr/local/etc /usr/local/include /usr/local/lib /usr/local/opt /usr/local/sbin /usr/local/share /usr/local/var /usr/local/Caskroom /usr/local/Cellar /usr/local/Homebrew /usr/local/Frameworks -name .DS_Store -delete
==> /usr/bin/sudo /usr/bin/find /usr/local/bin /usr/local/etc /usr/local/include /usr/local/lib /usr/local/opt /usr/local/sbin /usr/local/share /usr/local/var /usr/local/Caskroom /usr/local/Cellar /usr/local/Homebrew /usr/local/Frameworks -depth -type d -empty -exec rmdir {} ;
==> /usr/bin/sudo rmdir /usr/local
rmdir: /usr/local: Operation not permitted
Warning: Failed during: /usr/bin/sudo rmdir /usr/local
Warning: Homebrew partially uninstalled (but there were steps that failed)!
To finish uninstalling rerun this script with `sudo`.
The following possible Homebrew files were not deleted:
/usr/local/bin/
/usr/local/com.cisco.packettracer/
/usr/local/etc/
/usr/local/Frameworks/
/usr/local/Homebrew/
/usr/local/include/
/usr/local/lib/
/usr/local/MacGPG2/
/usr/local/opt/
/usr/local/sbin/
/usr/local/share/
/usr/local/var/
You may wish to remove them yourself.

pyenv:8: command not found: pyenv                                                                                        
[myusername@ ~]% sudo !!
sudo curl -fsSLO https://raw.githubusercontent.com/Homebrew/install/HEAD/uninstall.sh
/bin/bash uninstall.sh --path /usr/local
Password:
Failed to locate Homebrew!

```

Two options but the first one is preferred (second one for just-in-case)
- It asks me for the password right when it needs it. 
- If the permission is not enough, stop the script so that it will not do the uninstall half way through. 

Also consider but not necessarily:
- Re-running should be possible. 

1. Make sure homebrew is installed in /usr/local
2. curl -fsSLO https://raw.githubusercontent.com/Homebrew/install/HEAD/uninstall.sh /bin/bash uninstall.sh --path /usr/local

MikeMcQuaid:
- Retry failed removals through the existing sudo helper
- Cover permission failures without requiring real root access

Fixes #1134

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -245,6 +245,30 @@ do
   shift
 done
 
+homebrew_prefix_markers=(
+  bin/brew
+  etc/bash_completion.d/brew
+  share/doc/homebrew
+  share/man/man1/brew.1
+  share/man/man1/brew-cask.1
+  share/zsh/site-functions/_brew
+  share/zsh/site-functions/_brew_cask
+  share/fish/vendor_completions.d/brew.fish
+  var/homebrew
+)
+homebrew_prefix_paths=(
+  "${homebrew_prefix_markers[@]}"
+  share/man/man1/README.md
+)
+homebrew_path_markers=(
+  .git
+  Homebrew
+  Library/Homebrew
+  Cellar
+  Caskroom
+  "${homebrew_prefix_markers[@]}"
+)
+
 # Attempt to locate Homebrew unless `--path` is passed
 if [[ "${#homebrew_prefix_candidates[@]}" -eq 0 ]]
 then
@@ -264,8 +288,10 @@ HOMEBREW_PREFIX="$(
   for p in "${homebrew_prefix_candidates[@]}"
   do
     [[ -d "${p}" ]] || continue
-    [[ ${p} == "${homebrew_prefix_default}" && -d "${p}/Homebrew/.git" ]] && echo "${p}" && break
-    [[ -d "${p}/.git" || -x "${p}/bin/brew" ]] && echo "${p}" && break
+    for f in "${homebrew_path_markers[@]}"
+    do
+      [[ -e "${p}/${f}" || -L "${p}/${f}" ]] && echo "${p}" && break 2
+    done
   done
 )"
 [[ -n "${HOMEBREW_PREFIX}" ]] || abort "Failed to locate Homebrew!"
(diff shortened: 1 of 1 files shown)


p-linnane (changes requested):
`homebrew_files` isn't a trusted sudo allowlist. This change turns every failed removal into `sudo rm -fr`, but the array includes:

- `/etc/paths.d/homebrew`, which can belong to the active `/opt/homebrew` install while we're uninstalling `/usr/local`
- paths derived from the repository's `.gitignore`, including `..` traversal
- arbitrary `HOMEBREW_CACHE` and `HOMEBREW_LOGS` values

I reproduced all three reaching the sudo fallback with temporary fixtures and a mocked `execute_sudo`.

This also doesn't fully fix #1134. On Apple Silicon we try to `rmdir /usr/local`, and unrelated software there makes the uninstall report failure and tell the user to rerun with sudo even though Homebrew is already gone.

Let's keep the sudo retry to a small set of paths we know belong to the selected installation. The `paths.d` file should only be removed when it's a regular non-symlink file containing `${HOMEBREW_PREFIX}/bin`. `.gitignore` and environment-derived paths shouldn't be retried with sudo, and we shouldn't treat a non-empty shared `/usr/local` as an uninstall failure. We'll need focused tests for those cases too.

MikeMcQuaid:
@p-linnane addressed now (bar adding additional tests as they aren't in this repo, we use a CI integration test, so this didn't feel like right place to add.

p-linnane (approved):

PR #1140, opened 2026-09-04
Remove Intel macOS support from installer

MikeMcQuaid:
- Homebrew on macOS is now Apple Silicon only, so abort on any other architecture before prompting for `sudo` and always install to `/opt/homebrew`, dropping the `/usr/local` prefix and its `paths.d` special case.
- Apple Silicon Macs require macOS 11 or newer, so the checks for macOS 10.7, 10.11 and 10.13 and the `version_gt` helper were unreachable.
- Point CI at `/opt/homebrew/bin`: the arm64 runners never used `/usr/local`.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -67,7 +67,7 @@ jobs:
         run: |
           if [[ "${RUNNER_OS}" = "macOS" ]]
           then
-            echo "/usr/local/bin:/usr/bin:/bin" >> "${GITHUB_PATH}"
+            echo "/opt/homebrew/bin:/usr/bin:/bin" >> "${GITHUB_PATH}"
           elif [[ "${RUNNER_OS}" = "Windows" ]]
           then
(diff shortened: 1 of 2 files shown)


p-linnane (approved):

ZhongRuoyu (approved):

b3z:
Is it still possible to run older versions or is there going to be a day when homebrew won't be usable on intel macs at all?

MikeMcQuaid:
@b3z It won't run at all in one year from now.

ddebernardy:
> @b3z It won't run at all in one year from now.

OSS has only ever thrived on supporting old hardware and platforms. At least let the thing run, for goodness' sake.

p-linnane:
“Letting it run” still requires maintaining compatibility. Homebrew’s volunteers have finite time and resources, and this timeline was announced in November 2025. Anyone who wants to continue maintaining Intel support is welcome to do so in a fork.

PR #1142, opened 2026-09-10
Support macOS Golden Gate 27

p-linnane:
Advances the oldest supported macOS to 15.0 and the prerelease cutoff to 28.0.

Companion to https://github.com/Homebrew/brew/pull/23928.


ZhongRuoyu (approved):
🌉

PR #1144, opened 2026-09-13
Allow custom installation prefixes

MikeMcQuaid:
- Limit resolved prefixes to the default prefix's byte length so bottles can be relocated.
- Support `--path` in unattended installs and check the selected path's permissions.
- Exercise shorter and equal-length prefixes with bottles in CI.

Needs https://github.com/Homebrew/brew/pull/23959 in the 7.0.2 release to avoid `brew doctor` shouting.


p-linnane (approved):
Let's see how it goes.

PR #1148, opened 2026-09-21
Allow installation without sudo access

MikeMcQuaid:
- Use writable prefixes and the installing account's group without requiring administrator membership.
- Honour `HOMEBREW_NO_SUDO` and detect known privilege failures.
- Try filesystem operations before requesting elevation.
- Skip Command Line Tools without sudo and make their installation failures non-fatal when a usable Git is available.
- Validate Git before downloading Homebrew, accepting Xcode's Git.
- Cover permissions and optional tools with Ruby tests and verify installation and package use as a real non-admin account in CI.
- Use the test account's login environment to avoid inheriting the runner's inaccessible working directory.


install.sh
```diff
@@ -179,8 +179,12 @@ then
   CHOWN=("/usr/sbin/chown")
   CHGRP=("/usr/bin/chgrp")
   GROUP="admin"
+  # Use admin only for members, so group changes can work without sudo.
+  if [[ " $(id -Gn) " != *" admin "* ]]
+  then
+    GROUP="$(id -gn)"
```
p-linnane, line 186: Could we drop group write permissions when falling back to `staff`, since other standard macOS users belong to that group and can otherwise replace `bin/brew`?

p-linnane (approved):

